PR #147, opened 2023-08-07
(MAINT) Sync reference docs

michaeltlombardi:
# PR Summary

This change syncs the reference documentation in the repository with the reference documentation published to the official docs site.

## PR Context

This enables us to keep the reference documentation in the repository as the source of truth for the schemas and CLI. When there is a release or a fix to the documentation, the docs team will sync the docs as-needed.

This should keep the friction lower for the engineering team, so they can adjust (or ping the docs team to adjust) reference documentation in the repository instead of tracking documentation across multiple repos.

Primarily, the updates in this sync include:

- Adding the required metadata
- Fixing broken links
- Fixing incorrect syntax

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,9 @@ With DSCv3, you can:
 
 ### Differences from PowerShell DSC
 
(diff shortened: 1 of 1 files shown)


SteveL-MSFT (approved):
Best that I can tell, this looks fine

Issue #96, opened 2023-07-01
Reccommend replacing `jsonschema` crate with `boon` for improved JSON Schema support


michaeltlombardi:
### Summary of the new feature / enhancement

The DSC libraries and tools implemented in rust currently use the [`jsonschema` v0.17.0 crate][01] for their JSON Schema handling. However, this implementation only fully supports JSON Schema drafts 4-7. Support for 2019-09 and 2020-12 is only partial and requires enabling a feature. Further, according to the [official test suite][02] maintained and published in collaboration with the JSON Schema  team and community, `jsonschema` currently has six errors and 17 invalid [results from the test cases][03].

By contrast, the [`boon` v0.5.0 crate][04] has full support for JSON Schema drafts 4-7, as well as 2019-09 and 2020-12. That version successfully passes every test case for the supported dialects.

Migrating from the `jsonschema` crate to `boon` would make the JSON Schema validation and integration more reliable and enable us to fully use draft 2020-12 JSON Schemas, which have several authoring improvements over prior drafts and are much easier to extend. Further, `boon` has support for custom assertions of [formats][05], [content media types][06], and [content encoding][07], all of which would be useful for DSC:

https://github.com/santhosh-tekuri/boon/blob/d466730e5e5c7c663bd6739e74e39d1e2f7baae4/tests/examples.rs#L64-L92

That being said, it looks like the [`schemars` v0.8.12 crate][08] doesn't support 2020-12 yet, though it [does support 2019-09][09]. The largest [differences between JSON Schema draft 2019-09 and 2020-12][10] are:

- 2020-12 replaces `items`/`additionalItems` with `prefixItems`/`items` - this applies mostly to how you represent tuples - when `items` is an object instead of an array, the behavior is consistent for both dialects.
- 2020-12 replaces `$recursiveRef`/`$recursiveAnchor` with `$dynamicRef`/`$dynamicAnchor` - neither of which we seem to use in our current schemas, but which has a large impact on defining custom vocabularies, which we _may_ find desirable in the future.
- 2020-12 updates the interactive behavior for the `contains` and `unevaluatedItems` keywords, but we don't use `contains`.

For our purposes, that means that exporting our schemas in the 2019-09 dialect will have no noticeable impact vs 2020-12.

[01]: https://docs.rs/jsonschema/latest/jsonschema/
[02]: https://github.com/bowtie-json-schema/bowtie/
[03]: https://bowtie-json-schema.github.io/bowtie/#/draft7
[...]

PR #154, opened 2023-08-10
(GH-96) Replace `jsonschema` crate with `boon`

michaeltlombardi:
# PR Summary

Prior to this change, the DSC library, CLI, and the Registry resource all relied on the `jsonschema` crate for schema compilation and validation.

As raised in #96, the `jsonschema` crate does not support newer versions of JSON Schema like drafts 2019-09 and 2020-12. Further, the implementation fails several of the tests designed by the JSON Schema organization to validate implementations for the newest dialect  it _does_ support (Draft 7).

This change prototypes the replacement of the `jsonschema` crate with the `boon` crate, which has full support for every released draft of JSON Schema and currently passes every test for implementation validation.

## PR Context

In this initial prototype change, the code includes duplicated behaviors that can be cleaned up or abstracted into helper functions.

Another benefit and possible avenue going forward is to turn the decomposed schemas in the repository into the source of truth for validating the incoming data. With Boon, DSC could pre-load the schemas and only have to compile schemas for resources on-demand thereafter.

In that model, the JSON schemas in the repository could be embedded in the binary or shipped alongside it.

This change also updates the use of Schemars to output Draft 2019-09 schemas for DSC and the Registry resource, to make them closer to the definitions for the schemas in the repository.

This change adds documentation to the structs in the Registry resource, showing how they can be used to provide improved documentation for the schema at author-time.

### DIFF
--- a/dsc/Cargo.toml
+++ b/dsc/Cargo.toml
@@ -22,3 +22,4 @@ serde_json = { version = "1.0", features = ["preserve_order"] }
 serde_yaml = { version = "0.9" }
 syntect = { version = "5.0", features = ["default-fancy"], default-features = false }
 thiserror = "1.0"
+boon = "0.5.1"
--- a/dsc/src/main.rs
+++ b/dsc/src/main.rs
@@ -13,8 +13,8 @@ use dsc_lib::{
     dscresources::invoke_result::{GetResult, SetResult, TestResult},
     dscresources::resource_manifest::ResourceManifest,
     dscerror::DscError};
-use jsonschema::{JSONSchema, ValidationError};
-use schemars::{schema_for, schema::RootSchema};
+use boon::{Compiler, Schemas, ValidationError};
+use schemars::{schema::RootSchema, gen::SchemaSettings};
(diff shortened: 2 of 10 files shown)


SteveL-MSFT:
It looks like `jsonschema` does (partially) support newer versions of jsonschema, but needs some features enabled in the `cargo.toml` file (see https://github.com/Stranger6667/jsonschema-rs).  The concern I have moving to `boon` is that it appears to be a less active repo than `jsonschema`.

SteveL-MSFT (changes requested):
Test if enabling the feature flags for `jsonschema` crate is sufficient

SteveL-MSFT:
Closing for now, we can revisit later

Stranger6667:
Folks, FYI, in `jsonschema==0.21.0` only `unevaluatedItems` & `vocabulary` are missing (other non-optional tests from the JSON Schema test suite depend on `unevaluatedItems`). I am going to implement `unevaluatedItems` in the next few releases. Thank you for your patience.

Issue #166, opened 2023-08-23
ResourceManifest::Get should be optional


anmenaga:
### Prerequisites

- [X] Write a descriptive title.
- [X] Make sure you are able to repro it on the latest version
- [X] Search the existing issues.

### Steps to reproduce

With addition of `Export` functionality, that may be the only functionality that a resource implements;
however currently specifying 'Get' in the manifest is mandatory, which produces an error if resource does not support it:

```
PS C:\DSCv3> dsc resource export -r Microsoft/ProcessList
Error: Manifest: C:\DSCv3\bin\debug\process.dsc.resource.json
JSON: missing field `get` at line 9 column 1
```

### Expected behavior

```console
NA
```

### Actual behavior

```console
NA
```

### Environment data

```PowerShell
NA
```

### Version

v1

michaeltlombardi:
Should `get` be optional, or should the manifest have a property like `export` that validates against a subschema like:

```yaml
title:       Resource Exporting
description: |
  Defines whether the resource can be used to export instances. When this value is `true`,
  the resource is processed normally with the `dsc resource export` and `dsc config export`
  commands. When this value is `false`, the resource is ignored. When this value is `only`,
  the resource is invalid for other operations, like `get` and `set`.
type:    [boolean, string]
enum:    [true, false, only]
default: false
```

And then an export-only manifest would define as either:

1. Top-level keyword:
   
   ```jsonc
   {
     "export": "only",
     "get": {
       // get definition
     }
   }
   ```
   
1. Keyword in `get`:

   ```jsonc
   {
     "get": {
       "export": "only",
       // get definition
     }
   }
   ```

Are there times where we would expect the `export` to be something different from `get`, except for whether instance properties are passed to the resource? Do I have to define `export` and `get` fully to support both, or can I do something like this?

```jsonc
{
  "get": {
    // get definition
  },
  "export": true
}
```

Instead of:

```jsonc
{
  "get": {
    // get definition
  },
  "export": {
    // same as get definition
  }
}
```

SteveL-MSFT:
It's also not clear to me a real example where a resource would only support `export`.  It means it can't be used for auditing and if it implements `export`, it shouldn't be much more to support `get`.

EmperorArthur:
> It's also not clear to me a real example where a resource would only support `export`. It means it can't be used for auditing and if it implements `export`, it shouldn't be much more to support `get`.

[...]

PR #171, opened 2023-08-25
Export scenario support

### LINKED ISSUE #73: Need a way to support `export` scenario

Users want to manually configure an application/workload and get all the relevant configuration out and apply it to many other instances.  Currently, `get` uses the supplied config as a filter.  Also some resources like `registry` and `file` wouldn't make sense to enumerate the whole registry or filesystem.

Propose an optional `export` method in addition to `get`, `set`, and `test`.  `dsc config export` would ensure all the resources in the config support this method otherwise it fails.

anmenaga:
# PR Summary

Fix #73 , #174

This adds support for enumeration of all resource instances.
Export support is declared in a resource manifest similar to Get/Set/Test methods:
```
    "export": {
      "executable": "process",
      "args": [
        "list"
      ]
    },
```
In response to this call a resource is expected to return JSONLINES - one per resource instance.

Not implemented yet:
1) Export for provider-based resources #183 
2) Export for classic PS DSC resources (this will need #183 plus new version of v2 `PSDesiredStateConfiguration` module)

## PR Context

Following scenarios supported:

1) Export of a single resource type.
    This returns a configuration document that can be piped to `dsc config set`
    For example:
```
dsc resource export -r Microsoft/Process | dsc config set
```

2) Export of all types in a configuration.
    This returns a configuration document that can be piped to `dsc config set`
    For example:
```
Get-Content -raw C:\DSCv3\process\ExportTest.dsc.yaml | dsc config export | dsc config set
```

3. "resource get --all" operation that returns `dsc config get` result for every instance of the specified resource type.
     For example:
```
dsc resource get --all -r Microsoft/Process
```

### DIFF
--- a/dsc/src/resource_command.rs
+++ b/dsc/src/resource_command.rs
@@ -3,6 +3,9 @@
 
 use crate::args::OutputFormat;
 use crate::util::{EXIT_DSC_ERROR, EXIT_INVALID_ARGS, EXIT_JSON_ERROR, add_type_name_to_json, write_output};
+use dsc_lib::configure::config_doc::Resource;
+use dsc_lib::configure::config_doc::Configuration;
+use std::collections::HashMap;
 
 use dsc_lib::{
     dscresources::dscresource::{Invoke, DscResource},
@@ -130,27 +133,37 @@ pub fn schema(dsc: &mut DscManager, resource: &str, format: &Option<OutputFormat
 }
 
 pub fn export(dsc: &mut DscManager, resource: &str, format: &Option<OutputFormat>) {
-    let resource = get_resource(dsc, resource);
+    let dsc_resource = get_resource(dsc, resource);
 
-    match resource.export() {
-        Ok(result) => {
-            for instance in result.actual_state
-            {
-                let json = match serde_json::to_string(&instance) {
-                    Ok(json) => json,
-                    Err(err) => {
-                        eprintln!("JSON Error: {err}");
-                        exit(EXIT_JSON_ERROR);
-                    }
-                };
-                write_output(&json, format);
-            }
-        }
+    let mut conf = Configuration::new();
+
+    let export_result = match dsc_resource.export() {
+        Ok(export) => { export }
         Err(err) => {
             eprintln!("Error: {err}");
             exit(EXIT_DSC_ERROR);
         }
+    };
+
+    for (i, instance) in export_result.actual_state.iter().enumerate()
+    {
+        let mut r = Resource::new();
+        r.resource_type = dsc_resource.type_name.clone();
+        r.name = format!("{}-{i}", r.resource_type);
+        let props: HashMap<String, serde_json::Value> = serde_json::from_value(instance.clone()).unwrap();
+        r.properties = Some(props);
+
+        conf.resources.push(r);
     }
+
+    let json = match serde_json::to_string(&conf) {
+        Ok(json) => json,
+        Err(err) => {
+            eprintln!("JSON Error: {err}");
+            exit(EXIT_JSON_ERROR);
+        }
+    };
+    write_output(&json, format);
 }
 
 pub fn get_resource(dsc: &mut DscManager, resource: &str) -> DscResource {
--- a/dsc/src/resource_command.rs
+++ b/dsc/src/resource_command.rs
@@ -6,6 +6,7 @@ use crate::util::{EXIT_DSC_ERROR, EXIT_INVALID_ARGS, EXIT_JSON_ERROR, add_type_n
 use dsc_lib::configure::config_doc::Resource;
 use dsc_lib::configure::config_doc::Configuration;
 use std::collections::HashMap;
+use dsc_lib::dscresources::invoke_result::GetResult;
 
 use dsc_lib::{
     dscresources::dscresource::{Invoke, DscResource},
@@ -42,6 +43,34 @@ pub fn get(dsc: &mut DscManager, resource: &str, input: &Option<String>, stdin: 
             eprintln!("Error: {err}");
             exit(EXIT_DSC_ERROR);
         }
+    }
+}
+
+pub fn get_all(dsc: &mut DscManager, resource: &str, _input: &Option<String>, _stdin: &Option<String>, format: &Option<OutputFormat>) {
+    let resource = get_resource(dsc, resource);
+
+    let export_result = match resource.export() {
+        Ok(export) => { export }
+        Err(err) => {
+            eprintln!("Error: {err}");
+            exit(EXIT_DSC_ERROR);
+        }
+    };
+
+    for instance in export_result.actual_state
+    {
+        let get_result = GetResult {
+            actual_state: instance.clone(),
+        };
+
+        let json = match serde_json::to_string(&get_result) {
+            Ok(json) => json,
+            Err(err) => {
+                eprintln!("JSON Error: {err}");
+                exit(EXIT_JSON_ERROR);
+            }
+        };
+        write_output(&json, format);
     }
 }
 
--- a/dsc/src/resource_command.rs
+++ b/dsc/src/resource_command.rs
@@ -3,9 +3,8 @@
 
 use crate::args::OutputFormat;
 use crate::util::{EXIT_DSC_ERROR, EXIT_INVALID_ARGS, EXIT_JSON_ERROR, add_type_name_to_json, write_output};
-use dsc_lib::configure::config_doc::Resource;
 use dsc_lib::configure::config_doc::Configuration;
-use std::collections::HashMap;
+use dsc_lib::configure::add_resource_export_results_to_configuration;
 use dsc_lib::dscresources::invoke_result::GetResult;
 
 use dsc_lib::{
@@ -166,24 +165,7 @@ pub fn export(dsc: &mut DscManager, resource: &str, format: &Option<OutputFormat
 
     let mut conf = Configuration::new();
 
-    let export_result = match dsc_resource.export() {
-        Ok(export) => { export }
-        Err(err) => {
-            eprintln!("Error: {err}");
-            exit(EXIT_DSC_ERROR);
-        }
-    };
-
-    for (i, instance) in export_result.actual_state.iter().enumerate()
-    {
-        let mut r = Resource::new();
-        r.resource_type = dsc_resource.type_name.clone();
-        r.name = format!("{}-{i}", r.resource_type);
-        let props: HashMap<String, serde_json::Value> = serde_json::from_value(instance.clone()).unwrap();
-        r.properties = Some(props);
-
-        conf.resources.push(r);
-    }
+    add_resource_export_results_to_configuration(&dsc_resource, &mut conf);
 
     let json = match serde_json::to_string(&conf) {
         Ok(json) => json,
--- a/dsc/src/resource_command.rs
+++ b/dsc/src/resource_command.rs
@@ -129,6 +129,30 @@ pub fn schema(dsc: &mut DscManager, resource: &str, format: &Option<OutputFormat
     }
 }
 
+pub fn export(dsc: &mut DscManager, resource: &str, format: &Option<OutputFormat>) {
+    let resource = get_resource(dsc, resource);
+
+    match resource.export() {
+        Ok(result) => {
+            for instance in result.actual_state
+            {
+                let json = match serde_json::to_string(&instance) {
+                    Ok(json) => json,
+                    Err(err) => {
+                        eprintln!("JSON Error: {err}");
+                        exit(EXIT_JSON_ERROR);
+                    }
+                };
+                write_output(&json, format);
+            }
+        }
+        Err(err) => {
+            eprintln!("Error: {err}");
+            exit(EXIT_DSC_ERROR);
+        }
+    }
+}
+
 pub fn get_resource(dsc: &mut DscManager, resource: &str) -> DscResource {
     // check if resource is JSON or just a name
     match serde_json::from_str(resource) {
--- a/dsc/tests/dsc_export.tests.ps1
+++ b/dsc/tests/dsc_export.tests.ps1
@@ -51,4 +51,23 @@ Describe 'resource export tests' {
         $set_results.results.count | Should -BeGreaterThan 1
         $set_results.results[0].result.afterState.result | Should -BeExactly "Ok"
     }
+
+    It 'Duplicate resource types in Configuration Export should result in error' {
+
+        $yaml = @'
+            $schema: https://schemas.microsoft.com/dsc/2023/03/configuration.schema.json
+            resources:
+            - name: Processes
+              type: Microsoft/Process
+              properties:
+                pid: 0
+            - name: Processes
+              type: Microsoft/Process
+              properties:
+                pid: 0
+'@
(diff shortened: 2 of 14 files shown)


dsc/tests/dsc_export.tests.ps1
```diff
@@ -0,0 +1,45 @@
...
+    It 'Export can be called on individual resource' {
+
+        $processes = dsc resource export -r Microsoft/Process
+        $LASTEXITCODE | Should -Be 0
+        $processes.count | Should -BeGreaterThan 1
+    }
+
+    It 'Export can be called on a configuration' {
+
+        $json = @'
```
SteveL-MSFT, line 15: To avoid confusion, this should be `$yaml`

dsc/tests/dsc_export.tests.ps1
```diff
@@ -0,0 +1,45 @@
...
+            $schema: https://schemas.microsoft.com/dsc/2023/03/configuration.schema.json
+            resources:
+            - name: Processes
```
SteveL-MSFT, line 18: It seems that there isn't a good case where you would export the same resource twice, we should add some validation in `dsc` to emit an error if the same resource is specified more than once for export

dsc/tests/dsc_export.tests.ps1
```diff
@@ -0,0 +1,45 @@
...
+              type: Microsoft/Process
+              properties:
+                pid: 0
+'@
+        $out = $json | dsc config export
+        $LASTEXITCODE | Should -Be 0
+        $config_with_process_list = $out | ConvertFrom-Json
+        $config_with_process_list.resources.count | Should -BeGreaterThan 1
```
SteveL-MSFT, line 26: Not a blocker for this PR, but we should probably have an env var test resource instead of process since it's something we can more easily control and validate

dsc/tests/dsc_export.tests.ps1
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    It 'Configuration Export can be piped to configuration Set' {
+
+        $json = @'
```
SteveL-MSFT, line 31: $yaml

dsc_lib/src/configure/mod.rs
```diff
@@ -150,6 +152,40 @@ impl Configurator {
         Ok(result)
     }
 
+    pub fn invoke_export(&self, _error_action: ErrorAction, _progress_callback: impl Fn() + 'static) -> Result<ConfigurationExportResult, DscError> {
+        let (config, messages, had_errors) = self.validate_config()?;
+        let mut result = ConfigurationExportResult::new();
```
SteveL-MSFT, line 157: You don't need to use the default constructor here, instead, use this syntax:

```rust
let mut result = ConfigurationExportResult {
  messages,
  had_errors
};
```

SteveL-MSFT (changes requested):
Create issues for the two items you identified for providers and PS not supported and link to this PR

michaeltlombardi:
To clarify my understanding of the behavior here:

1. The `dsc resource export` command is for returning the full list of instances for a resource without any special handling from `dsc` - just returning the JSONLINES the resource emits. No formatting or result object.
2. The `dsc config export` command is for returning an immediately usable configuration document, where the `resources` key is a flat array of the instances returned by every defined instance in the input configuration.

Assuming that's correct, a few thoughts:

#### Regarding the semantics/output of `dsc resource export`

The `dsc resource export` output breaks from the way all other `dsc` commands behave, where they have a result schema that handles some behaviors/collection for you. I would expect a result that mirrors the output from [`dsc resource get`][01] to some degree. I think the _resource_ should definitely return the list of instances as JSONLINES, but I'm not sure that's the right contract for the `dsc resource export` command. It would make this output unique for `dsc` in that it just passes the resource output back through to the caller.

I think the semantic UX of the `dsc resource export` command is confusing. If I was a naive user, I would expect to be able to call `dsc resource export -r foo/bar` to get a configuration document that _only_ includes instances of that resource. Functionally, a shorter way for me to get the functionality of `dsc config export` when I only care about a single resource.
  
As discussed in #73 ([comment link][02]), the current functionality of `dsc resource export` more closely mirrors the proposal for `dsc resource get --all`, which is a definitively distinct operation from `dsc resource export`. There's valid reasons both to say "I want the current state of all instances of this resource" (`dsc resource get --all`) and "I want a configuration based on every defined instance of this resource" (`dsc resource export`). The _underlying code_ for what the resource does to return the values is the same, but how DSC presents the operation option in UX and processes the output are different.

#### Regarding the schema/data model

In #166, I [mentioned][03] that we might want to have a better DevX for defining the manifest entry to indicate whether a resource supports the `dsc resource get --all`/`dsc resource export`/`dsc config export` commands. I suspect the most _common_ model for implementing support for this will be one of the following:

1. When you call the command defined for `get` without passing it any input JSON or argument flags, it returns every instance of the resouce.
1. When you call the command defined for `get` with a single boolean flag, like `--all`, it ignores other filtering options and input JSON, and returns every instance of the resource.

I think a fully separate command structure is less likely, but should be enabled. However, it would be good DevX to allow manifest authors to opt-in for these operations without having to redefine the setting if the case is `1`. It's also potentially less intuitive to define `get` and `export` separately than to define an `all` property in `get` to indicate alternate arguments. What the author is defining is whether their resource supports returning every instance of itself.

```jsonc
// 'process get --all' expects no input, returns all instances
{
    "get": {
        "executable": "process",
        "args": ["get"],
        "all": {
          // executable is optional
          "args": ["get", "--all"]
        }
    }
}
```

```jsonc
// 'process get' returns all instances when called without input
{
    "get": {
        "executable": "process",
        "args": ["get"],
        "all": true // indicates that calling without input returns all
    }
}
```

```jsonc
// 'process get' returns null/error when called without input
{
    "get": {
        "executable": "process",
        "args": ["get"],
        "all": false // indicates that the resource doesn't support this op
    }
}
```

```jsonc
// 'osinfo' always and only returns one instance
{
    // this doesn't exist yet, but worth noting we can't tell the difference now
    "isSingleInstance": true,
    "get": {
        "executable": "osinfo",
        "args": [],
        "all": true
    }
}
```
[01]: https://github.com/PowerShell/DSC/blob/main/schemas/2023/08/bundled/outputs/resource/get.json
[02]: https://github.com/PowerShell/DSC/issues/73#issuecomment-1603487668
[03]: https://github.com/PowerShell/DSC/issues/166#issuecomment-1691973740

anmenaga:
> 1. The `dsc resource export` command is for returning the full list of instances for a resource without any special handling from `dsc` - just returning the JSONLINES the resource emits. No formatting or result object.

"No formatting" part is not correct. Internally there is deserialization from json, emitted by resource, into internal representation and then serialization into output format. This means that dsc will catch errors like invalid json, issues by resource, and can report back json as well as yaml based on `dsc --format` flag.

> I would expect to be able to call `dsc resource export -r foo/bar` to get a configuration document that _only_ includes instances of that resource. Functionally, a shorter way for me to get the functionality of `dsc config export` when I only care about a single resource.

I tend to agree with this. This would make it consistent so that `export` command always (whether `resource export` or `config export`) returns a configuration document that can be used in `config set`.

dsc_lib/src/configure/mod.rs
anmenaga: Updated.

dsc/tests/dsc_export.tests.ps1
anmenaga: Updated.

dsc/tests/dsc_export.tests.ps1
anmenaga: Updated.

dsc/tests/dsc_export.tests.ps1
anmenaga: I agree;  env var test resource would be better here.

dsc/tests/dsc_export.tests.ps1
```diff
@@ -0,0 +1,72 @@
...
+        $config_with_process_list = $out | ConvertFrom-Json
+        $config_with_process_list.resources.count | Should -BeGreaterThan 1
+    }
+
+    It 'get --all can be called on individual resource' {
+
+        $out = dsc resource get --all -r Microsoft/Process
+        $LASTEXITCODE | Should -Be 0
+        $process_list = $out | ConvertFrom-Json
+        $process_list.resources.count | Should -BeGreaterThan 1
```
SteveL-MSFT, line 19: should do some validation that each result is a getresult

dsc/tests/dsc_export.tests.ps1
```diff
@@ -0,0 +1,72 @@
...
+# Licensed under the MIT License.
+
+Describe 'resource export tests' {
+    
+    It 'Export can be called on individual resource' {
+
+        $out = dsc resource export -r Microsoft/Process
+        $LASTEXITCODE | Should -Be 0
+        $config_with_process_list = $out | ConvertFrom-Json
+        $config_with_process_list.resources.count | Should -BeGreaterThan 1
```
SteveL-MSFT, line 11: do some validation that this is a config doc

dsc/tests/dsc_export.tests.ps1
```diff
@@ -0,0 +1,72 @@
...
+            resources:
+            - name: Processes
+              type: Microsoft/Process
+              properties:
+                pid: 0
+'@
+        $out = $yaml | dsc config export
+        $LASTEXITCODE | Should -Be 0
+        $config_with_process_list = $out | ConvertFrom-Json
+        $config_with_process_list.resources.count | Should -BeGreaterThan 1
```
SteveL-MSFT, line 35: some validation that the output is a config doc

dsc_lib/src/configure/mod.rs
```diff
@@ -7,8 +7,10 @@ use crate::dscerror::DscError;
 use crate::dscresources::dscresource::Invoke;
 use crate::discovery::Discovery;
 use self::config_doc::Configuration;
-use self::config_result::{ConfigurationGetResult, ConfigurationSetResult, ConfigurationTestResult, ResourceMessage, MessageLevel};
 use self::depends_on::get_resource_invocation_order;
+use self::config_result::{ConfigurationGetResult, ConfigurationSetResult, ConfigurationTestResult, ConfigurationExportResult, ResourceMessage, MessageLevel};
+use std::collections::HashMap;
+use std::collections::HashSet;
```
SteveL-MSFT, line 13:
```suggestion
use std::collections::{HashMap, HashSet};
```

dsc_lib/src/configure/mod.rs
```diff
@@ -152,6 +154,72 @@ impl Configurator {
...
+        result.into_iter().collect()
+    }
+
+    pub fn invoke_export(&self, _error_action: ErrorAction, _progress_callback: impl Fn() + 'static) -> Result<ConfigurationExportResult, DscError> {
+        let (config, messages, had_errors) = self.validate_config()?;
+
+        let duplicates = Self::find_duplicate_resource_types(&config);
+        if !duplicates.is_empty()
+        {
+            let duplicate = &duplicates[0];
```
SteveL-MSFT, line 184: There can be multiple dupes here.  I think you can format this as an array:

`Resource(s) Name, Name2 specified multiple times`

There's probably a join function you can use to create the new string with commas.

dsc_lib/src/configure/mod.rs
```diff
@@ -152,6 +154,72 @@ impl Configurator {
...
+            for (i, instance) in export_result.actual_state.iter().enumerate()
+            {
+                let mut r = config_doc::Resource::new();
+                r.resource_type = dsc_resource.type_name.clone();
+                r.name = format!("{}-{i}", r.resource_type);
+                let props: HashMap<String, serde_json::Value> = serde_json::from_value(instance.clone()).unwrap();
+                r.properties = Some(props);
+
+                conf.resources.push(r);
+            }
```
SteveL-MSFT, line 214: This code seems very similar or same as above.  Maybe have a helper that creates a new config based on export_result?

SteveL-MSFT (changes requested):

dsc/src/resource_command.rs
```diff
@@ -3,6 +3,10 @@
 
 use crate::args::OutputFormat;
 use crate::util::{EXIT_DSC_ERROR, EXIT_INVALID_ARGS, EXIT_JSON_ERROR, add_type_name_to_json, write_output};
+use dsc_lib::configure::config_doc::Resource;
+use dsc_lib::configure::config_doc::Configuration;
```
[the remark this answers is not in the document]
anmenaga: Updated.

dsc/tests/dsc_export.tests.ps1
anmenaga: Updated.

dsc/tests/dsc_export.tests.ps1
anmenaga: Updated.

dsc/tests/dsc_export.tests.ps1
anmenaga: Updated.

dsc_lib/src/configure/mod.rs
anmenaga: Updated.

dsc_lib/src/configure/mod.rs
anmenaga: Updated.

dsc_lib/src/configure/mod.rs
anmenaga: Updated.

dsc/tests/dsc_export.tests.ps1
```diff
@@ -0,0 +1,77 @@
...
+# Licensed under the MIT License.
+
+Describe 'resource export tests' {
+    
+    It 'Export can be called on individual resource' {
+
+        $out = dsc resource export -r Microsoft/Process
+        $LASTEXITCODE | Should -Be 0
+        $config_with_process_list = $out | ConvertFrom-Json
+        $config_with_process_list.'$schema' | Should -Not -BeNullOrEmpty
```
SteveL-MSFT, line 11: Can we validate the `$schema` value?

dsc/tests/dsc_export.tests.ps1
```diff
@@ -0,0 +1,77 @@
...
+            resources:
+            - name: Processes
+              type: Microsoft/Process
+              properties:
+                pid: 0
+'@
+        $out = $yaml | dsc config export
+        $LASTEXITCODE | Should -Be 0
+        $config_with_process_list = $out | ConvertFrom-Json
+        $config_with_process_list.'$schema' | Should -Not -BeNullOrEmpty
```
SteveL-MSFT, line 38: Same for $schema

SteveL-MSFT (changes requested):

dsc/tests/dsc_export.tests.ps1
anmenaga: Updated.

dsc/tests/dsc_export.tests.ps1
anmenaga: Updated.

SteveL-MSFT (approved):

Issue #45, opened 2023-03-29
Implement `dependsOn` processing for config


SteveL-MSFT:
### Summary of the new feature / enhancement

Build execution graph for a config

SteveL-MSFT:
Using ARM syntax, this appears to be: `[resourceId('<type>','<name>')]` where `<name>` could be generated using parameters/variables at runtime.  For now, (since parameters and variables isn't implemented), it'll just be a string, but we'll follow the syntax.

SteveL-MSFT:
https://github.com/PowerShell/DSC/pull/175

PR #175, opened 2023-08-29
Dependson

SteveL-MSFT:
# PR Summary

Implement `dependsOn` support for configurations:

- refactor config mod.rs to use new function to get resource invocation order
- order is based on dependencies first then the resource
- detect invalid dependencies and have strict validation of syntax
- detect circular dependencies
- added unit tests

## PR Context

#45

### DIFF
--- a/dsc_lib/src/configure/config_doc.rs
+++ b/dsc_lib/src/configure/config_doc.rs
@@ -71,6 +71,7 @@ pub struct Resource {
     /// A friendly name for the resource instance
     pub name: String, // friendly unique instance name
     #[serde(rename = "dependsOn", skip_serializing_if = "Option::is_none")]
+    #[schemars(regex(pattern = r"^\[resourceId\('[a-zA-Z0-9\.]+/[a-zA-Z0-9]+','[a-zA-Z0-9 ]+'\)]$"))]
     pub depends_on: Option<Vec<String>>,
     // `identity` can be used for run-as
     #[serde(skip_serializing_if = "Option::is_none")]
--- a/dsc_lib/src/configure/config_doc.rs
+++ b/dsc_lib/src/configure/config_doc.rs
@@ -71,7 +71,7 @@ pub struct Resource {
     /// A friendly name for the resource instance
     pub name: String, // friendly unique instance name
     #[serde(rename = "dependsOn", skip_serializing_if = "Option::is_none")]
-    #[schemars(regex(pattern = r"^\[resourceId\('[a-zA-Z0-9\.]+/[a-zA-Z0-9]+','[a-zA-Z0-9 ]+'\)]$"))]
+    #[schemars(regex(pattern = r"^\[resourceId\(\s*'[a-zA-Z0-9\.]+/[a-zA-Z0-9]+'\s*,\s*'[a-zA-Z0-9 ]+'\s*\)]$"))]
     pub depends_on: Option<Vec<String>>,
     // `identity` can be used for run-as
     #[serde(skip_serializing_if = "Option::is_none")]
--- a/dsc_lib/src/configure/depends_on.rs
+++ b/dsc_lib/src/configure/depends_on.rs
@@ -1,0 +1,254 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+use regex::Regex;
+
+use crate::configure::config_doc::Resource;
+use crate::configure::Configuration;
+use crate::DscError;
+
+pub fn get_resource_invocation_order(config: &Configuration) -> Result<Vec<Resource>, DscError> {
+    let mut order: Vec<Resource> = Vec::new();
+    let depends_on_regex = Regex::new(r"^\[resourceId\('(?<type>[a-zA-Z0-9\.]+/[a-zA-Z0-9]+)','(?<name>[a-zA-Z0-9 ]+)'\)]$")?;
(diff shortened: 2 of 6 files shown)


dsc_lib/src/configure/config_doc.rs
```diff
@@ -71,6 +71,7 @@ pub struct Resource {
     /// A friendly name for the resource instance
     pub name: String, // friendly unique instance name
     #[serde(rename = "dependsOn", skip_serializing_if = "Option::is_none")]
+    #[schemars(regex(pattern = r"^\[resourceId\('[a-zA-Z0-9\.]+/[a-zA-Z0-9]+','[a-zA-Z0-9 ]+'\)]$"))]
```
michaeltlombardi, line 74: Can this allow for an optional space between the type name and the instance name? Does it _have_ to use single quotes?

Thinking about examples like:

```yaml
resources:
- name: current user registry
  type: Microsoft.Windows/Registry
  properties:
    keyPath: HKCU\example
    _ensure: Present
  dependsOn:
  # Current implementation
  - "[resourceId('DSC/AssertionGroup','my assertions')]"
  # With outer quotes as single quotes
  - '[resourceId("DSC/AssertionGroup","my assertions")]'
  # With a space between the names
  - "[resourceId('DSC/AssertionGroup', 'my assertions')]"
```

The regex for the type name also doesn't match our current regex - I'm okay with adjusting either of them:

https://github.com/PowerShell/DSC/blob/5a1f8b582edfbe8a040195baa7ff4bdb7231addd/schemas/src/definitions/resourceType.yaml#L3-L17

We also don't have a regex currently applied to instance names, we allow arbitrary strings. Should that be more limited?

dsc_lib/src/configure/depends_on.rs
```diff
@@ -0,0 +1,254 @@
...
+
+use regex::Regex;
+
+use crate::configure::config_doc::Resource;
+use crate::configure::Configuration;
+use crate::DscError;
+
+pub fn get_resource_invocation_order(config: &Configuration) -> Result<Vec<Resource>, DscError> {
+    let mut order: Vec<Resource> = Vec::new();
+    let depends_on_regex = Regex::new(r"^\[resourceId\('(?<type>[a-zA-Z0-9\.]+/[a-zA-Z0-9]+)','(?<name>[a-zA-Z0-9 ]+)'\)]$")?;
```
michaeltlombardi, line 12: Wondering if it might be useful, organizationally, to define common regexes in one place and re-use them. Like defining the regex for:

- Resource type names
- Instance names
- The resourceID syntax
- Other custom syntaxes that we add later, like for looking up parameters or variables or outputs.

dsc_lib/src/configure/config_doc.rs
SteveL-MSFT: Best I can tell from ARM, whitespace is allowed.  I'll adjust the regex.

dsc_lib/src/configure/depends_on.rs
SteveL-MSFT: Yes, it would make sense to have them in one place.  One limitation with Rust I hit is that I can't define a const regex (or string) and use it as an attribute value which only takes a literal.

dsc_lib/src/configure/depends_on.rs
```diff
@@ -0,0 +1,254 @@
...
+    for resource in &config.resources {
```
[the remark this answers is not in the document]
SteveL-MSFT: You would have to depend on the provider resource instance which should still work.  You can't depend on a resource within a provider or group resource directly.  This is by-design.

dsc_lib/src/configure/depends_on.rs
SteveL-MSFT: I'll leave it here for now unless I can figure out a way to use the same regex in the attribute and in code.  It doesn't seem like Rust macros solve this.  This regex is currently the only one used in code.

dsc_lib/src/configure/depends_on.rs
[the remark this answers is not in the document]
anmenaga: Discussed offline. For now, this will be sufficient because dependencies can depend on the parent provider resource.

anmenaga (approved):

dsc_lib/src/configure/config_doc.rs
michaeltlombardi: With the updated whitespace, dangling questions are:

1. Can the regex also support using double quotes instead of single quotes?
2. Can the instance names be arbitrary strings or are they limited to alphanumeric characters and spaces? If they're limited, I can adjust the docs and schemas instead of allowing arbitrary strings. Right now, this regex makes the following instance names invalid:
  ```yaml
  - name: Security.PublicAccess.Disable
  - name: Security/PublicAccess/Disable
  - name: Security - Disable public access
  - name: 'Security: Disable public access'
  ```

SteveL-MSFT: @michaeltlombardi best I can tell looking at the ARM docs it's single quote only because it's in JSON which would get wrapped by double quotes.  Don't see any specific rules on characters, but the examples are all alphanumeric with spaces allowed, so I think that's ok unless you can find something more specific to what ARM allows

SteveL-MSFT: We can update this as a separate PR as needed

Issue #133, opened 2023-08-03
Provide ability to define dsc-v3 resource path (Support sxs installation)


BerheAbrha:
### Summary of the new feature / enhancement

dscv3 is relying on system variable ($env:path) for resource discovery.  Provide an ability to specify the directory for all dscv3 resources.  
          - This will enable sxs installation possible.  (GC-Agent supports this)
          - DSC is slow locating resources since it is checking a big list of paths in $env:path.

### Proposed technical implementation details (optional)
Pass optional DSC_Resouce_Path to DSC .
_No response_

BerheAbrha:
Overwriting the system $env:path in the process may impact locating executable or other resource dependencies that are already under $env:path

Bpoe:
I agree with BerheAbrha. Resources called and the processes spawned calling those may depend on PATH being intact. We will need to support either specifying the search path(s) as an argument or support another ENV var that would be used if specified instead of PATH. (Think something similar to PSModulePath)

SteveL-MSFT:
We can have a `DSC_RESOURCE_PATH` env var that, if defined, overrides any search of `PATH`.

PR #176, opened 2023-08-29
add support for DSC_RESOURCE_PATH

SteveL-MSFT:
# PR Summary

If `DSC_RESOURCE_PATH` env var exists, we use that instead of `PATH` for resource discovery.

## PR Context

Fix https://github.com/PowerShell/DSC/issues/133

### DIFF
--- a/dsc/tests/dsc_discovery.tests.ps1
+++ b/dsc/tests/dsc_discovery.tests.ps1
@@ -1,0 +1,28 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+Describe 'tests for resource discovery' {
+    It 'Use DSC_RESOURCE_PATH instead of PATH when defined' {
+        $resourceJson = @'
+        {
+            "manifestVersion": "1.0",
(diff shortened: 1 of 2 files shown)


michaeltlombardi:
@SteveL-MSFT - so, to clarify, the _only_ way to set this is to do either:

```sh
DSC_RESOURCE_PATH=/usr/mikey/resources dsc resource list
```
or
```powershell
$env:DSC_RESOURCE_PATH = 'C:\Users\Mikey\resources'
dsc resource list
```
but not:
```sh
dsc --resource_path /usr/mikey/resources resource list
```
or
```powershell
$ResourcePaths = @(
  'C:\Users\Mikey\resources',
  'C:\dsc_resources'
) -join ';'
dsc resource list --resource_path $ResourcePaths
```

This is an environment variable _only_ override, not an optional argument?

anmenaga (approved):
LGTM

SteveL-MSFT:
> This is an environment variable _only_ override, not an optional argument?

@michaeltlombardi Correct, this is by env var only

Issue #150, opened 2023-08-09
Support sending control signals to DSC


BerheAbrha:
### Summary of the new feature / enhancement

-  When other agents are invoking DSC and if issued operation need to stop.  It will need a way to send control signal to the child process. 

### Proposed technical implementation details (optional)
Send event to parent process so it can close the process handle. 

WINDOWS
```
API DOC for sending events: - https://learn.microsoft.com/en-us/windows/win32/api/synchapi/nf-synchapi-createeventa
```
LINUX 
```
https://www.tutorialspoint.com/unix_system_calls/sigaction.htm
```

Bpoe:
Would responding to SIGINT as a request to stop meet the needs?
(https://stackoverflow.com/questions/813086/can-i-send-a-ctrl-c-sigint-to-an-application-on-windows)

michaeltlombardi:
The Rust CLI book has a chapter on [signal handling][01], which points to the [signal_hook][02] crate as an option. 

It looks like the crate currently supports signals fully for not-Windows, and supports a limited subset of signals on windows, notably including `SIGABRT`, `SIGBREAK`, `SIGINT`, and `SIGTERM`.

[01]: https://rust-cli.github.io/book/in-depth/signals.html
[02]: https://docs.rs/signal-hook/latest/signal_hook/#portability

jodi-boone:
Please re-open as we need to ensure child processes are killed as well as parent.

PR #177, opened 2023-08-30
exit program if ctrl+c is received

SteveL-MSFT:
# PR Summary

Program exits if `ctrl+c` (sigint) is received.  At this point, it's probably safer to let child processes continue to run and finish than to kill them mid-process.

## PR Context

Fix https://github.com/PowerShell/DSC/issues/150

### DIFF
--- a/dsc/src/util.rs
+++ b/dsc/src/util.rs
@@ -32,6 +32,7 @@ pub const EXIT_DSC_ERROR: i32 = 2;
 pub const EXIT_JSON_ERROR: i32 = 3;
 pub const EXIT_INVALID_INPUT: i32 = 4;
 pub const EXIT_VALIDATION_FAILED: i32 = 5;
+pub const EXIT_CTRL_C: i32 = 6;
 
 pub fn serde_json_value_to_string(json: &serde_json::Value) -> String
 {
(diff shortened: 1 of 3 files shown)


dsc/src/util.rs
```diff
@@ -32,6 +32,7 @@ pub const EXIT_DSC_ERROR: i32 = 2;
 pub const EXIT_JSON_ERROR: i32 = 3;
 pub const EXIT_INVALID_INPUT: i32 = 4;
 pub const EXIT_VALIDATION_FAILED: i32 = 5;
+pub const EXIT_CTRL_C: i32 = 6;
```
michaeltlombardi, line 35: After this PR lands, I'm going to update the CLI docs to include exit codes for DSC itself.

michaeltlombardi: Actually, would it be possible to get short docs strings for these constants? That would make documenting them in CLI docs easier. Right now I'm trying to track down how they're used throughout the code.

anmenaga:
> At this point, it's probably safer to let child processes continue to run and finish than to kill them mid-process.

This really depends on the business logic of the resource/scenario that is running. I can see the other option valuable as well.
In short term, this is fine (assuming GC is happy with this).
In the long run, this behavior should probably be configurable by dsc config file or something of that sort. 

PS. In the long run, we'll also need a timeout/cancelation protection from hanging resource processes.

anmenaga (approved):

dsc/src/util.rs
SteveL-MSFT: We can add them in a separate PR.  Can probably put them in the `.dsc.resource.json` files.  Created https://github.com/PowerShell/DSC/issues/181

PR #185, opened 2023-09-01
Document `dependsOn` update

michaeltlombardi:
# PR Summary

This change:

- Updates the schema and documentation for instance names to clarify that they must be non-empty strings containing only letters, numbers, and spaces.
- Updates the schema and documentation for the `dependsOn` property to use the new `resourceId()` function syntax as added in #175.
- Regenerates the schemas with the updated definitions for `2023/08`
- Updates the YAML metadata blocks in the reference documentation to match the updated blocks on the learn platform.

## PR Context

Prior to this change, the schema for the `dependsOn` keyword used the PSDSC syntax and the schema for instance names allowed arbitrary strings.

For now, this change does not define new reference documentation for the configuration functions. When more configuration functions are added, we can define a separate set of reference documentation for those functions.

### DIFF
--- a/docs/reference/schemas/config/resource.md
+++ b/docs/reference/schemas/config/resource.md
@@ -14,9 +14,9 @@ Defines a DSC Resource instance in a configuration document.
 ## Metadata
 
 ```yaml
-Schema Dialect : https://json-schema.org/draft/2020-12/schema
-Schema ID      : https://raw.githubusercontent.com/PowerShell/DSC/main/schemas/2023/08/config/document.resource.json
-Type           : object
+SchemaDialect: https://json-schema.org/draft/2020-12/schema
+SchemaID:      https://raw.githubusercontent.com/PowerShell/DSC/main/schemas/2023/08/config/document.resource.json
+Type:          object
 ```
 
 ## Description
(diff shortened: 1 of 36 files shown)


docs/reference/schemas/config/resource.md
```diff
@@ -80,30 +84,57 @@ Required: true
...
+This property defines a list of DSC Resource instances that DSC must successfully process before
+processing this instance. Each value for this property must be the `resourceID()` lookup for
+another instance in the configuration. Multiple instances can depend on the same instance, but
+every dependency for an instance must be unique in that instance's `dependsOn` property.
+
+The `resourceID()` function uses this syntax:
 
 ```yaml
-"[<resource-type-name>]<instance-name>"
+"[resourceId('<resource-type-name>', '']<instance-name>"
```
SteveL-MSFT, line 95: Shouldn't this be:
```suggestion
"[resourceId('<resource-type-name>', '<instance-name>']"
```

docs/reference/schemas/config/resource.md
```diff
@@ -80,30 +84,57 @@ Required: true
...
+```yaml
+- name: Update Tailspin Automatically
+  type: Microsoft.Windows/Registry
+  properties:
+    keyPath:   HKCU\tailspin\updates
+    valueName: automatic
+    valueData:
+      String: enable
+  dependsOn:
+    - "[resourceId('Microsoft.Windows/Registry', 'Tailspin Key')]"
```
SteveL-MSFT, line 114: Would the example be better to have the resource named `Tailspin Key` in it?

docs/reference/schemas/config/resource.md
```diff
@@ -80,30 +84,57 @@ Required: true
...
-Multiple instances can depend on the same instance, but every dependency for an instance must be
-unique in that instance's `dependsOn` property.
+> [!NOTE]
+> When defining dependencies for [nested resource instances][02], instances can only reference
+> dependencies in the same resource provider or group instance. They can't use the `resourceId()`
+> function to lookup instances at the top-level of the configuration document or inside another
+> provider or group instance.
+>
+> If a top-level instance depends on a nested instance, use the `resourceId()` function to lookup
+> the instance of the provider or group containing the dependency instance instead.
```
SteveL-MSFT, line 124: Does it make sense to say that if there is a need to depend on something deeply nested, you can use grouping to achieve the same effect.  An example would probably be best.

SteveL-MSFT (changes requested):
minor feedback

docs/reference/schemas/config/resource.md
michaeltlombardi: Can you share a pseudo-example of what you mean? I can convert this into a full example afterward, though that may better belong in the eventual conceptual article for managing dependencies in configuration documents.

SteveL-MSFT (approved):

PR #198, opened 2023-09-19
add support to pass JSON as env var

### LINKED ISSUE #98: Implement InputKind Env

InputKind = `Args` is intended for simple cases like `bash script resource` where a flat JSON schema would result as name/value pairs sent as parameters and parameter values to the script:

```json
{
  "property1": 1,
  "property2": "hello"
}
```

Would result in:

```console
yourresource -property1 1 -property2 "hello"
```

_No response_

SteveL-MSFT:
# PR Summary

Enable passing JSON as env vars to command resource.  Only strings, numbers, array of strings, array of numbers, or booleans are supported.  Boolean array is excluded as it doesn't seem useful.

Note there is a change in behavior in that previously if `get` in the resource manifest didn't specify `input`, it was assumed `stdin` and still passed through.  However, if `input` is not defined, it's now assuming no input even if it was sent via `stdin`.

## PR Context

Fix #98

### DIFF
--- a/dsc/tests/dsc_resource_input.tests.ps1
+++ b/dsc/tests/dsc_resource_input.tests.ps1
@@ -1,0 +1,120 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+Describe 'tests for resource input' {
+    BeforeAll {
+        $manifest = @'
+    {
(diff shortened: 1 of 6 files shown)


anmenaga (approved):
LGTM

Issue #127, opened 2023-08-03
Update or replace `manifestVersion`


michaeltlombardi:
### Summary of the new feature / enhancement

In prior conversations and in the decomposed schema, I have the `manifestVersion` key in the resouce manifest documented as defining the version of the resource manifest schema that the manifest adheres to, requiring it to be semver and limited to the versions of the schema we support, which is currently `1.0`.

A few things:

- `1.0` isn't valid semver by the [regex recommended on semver.org][01] - it would have to be `1.0.0` (and should maybe more accurately be `0.1.0`)
- I thought I remembered there being a check on this version in the code, but a quick check didn't turn one up.
- If this is actually referencing the canonical schema for the manifest, should this maybe just be the `$schema` key and point to the resource manifest schema's `$id`?

[01]: https://semver.org/#is-there-a-suggested-regular-expression-regex-to-check-a-semver-string

### Proposed technical implementation details (optional)

Either:

1. Update `manifestVersion` to use a valid semantic version string, like `0.1.0`
2. Replace `manifestVersion` with the `$schema` keyword and require the specified schema to be one of the list of supported schemas.

SteveL-MSFT:
@michaeltlombardi I think using `$schema` is the better option, can you suggest the URL?

michaeltlombardi:
I would recommend the bundled schema for the manifest, which is currently:

```yaml
https://raw.githubusercontent.com/PowerShell/DSC/main/schemas/2023/08/bundled/resource/manifest.json
```

This will also automatically associate the JSON file with the schema for IntelliSense and hover help (though not the extended authoring schema, which is something we can investigate - same URL, but `manifest.vscode.json`).

I would recommend making the value an enum so we can be forward-thinking about managing schema versions. Also, in the next release, we could mark the `manifestVersion` key as deprecated and indicate they should migrate to using `$schema`, then remove `manifestVersion` in the next release or two.

PR #199, opened 2023-09-19
change ManifestVersion to $schema and validate

SteveL-MSFT:
# PR Summary

Change `manifestVersion` in the resource manifest to `$schema` using a URI instead of a semver.  Added code to validate URI is a supported version by abstracting new `import_manifest` function.

## PR Context

Fix https://github.com/PowerShell/DSC/issues/127

### DIFF
--- a/dsc/assertion.dsc.resource.json
+++ b/dsc/assertion.dsc.resource.json
@@ -1,5 +1,5 @@
 {
-  "manifestVersion": "1.0",
+  "$schema": "https://raw.githubusercontent.com/PowerShell/DSC/main/schemas/2023/08/bundled/resource/manifest.json",
   "type": "DSC/AssertionGroup",
   "version": "0.1.0",
   "description": "`test` will be invoked for all resources in the supplied configuration.",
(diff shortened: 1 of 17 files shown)


SteveL-MSFT:
@michaeltlombardi I'm leaving it to you to update the schemas

dsc/assertion.dsc.resource.json
```diff
@@ -1,5 +1,5 @@
 {
-  "manifestVersion": "1.0",
+  "$schema": "https://raw.githubusercontent.com/PowerShell/DSC/main/schemas/2023/08/bundled/resource/manifest.json",
```
miroman9364, line 2: I don't think it matters at this point, but keep in mind that GitHub is going to server this as "text/plain" and at some point we will probably want/need to have it served as "application/schema+json". There's a short discussion about this in the JSON Schemas repo: https://github.com/orgs/json-schema-org/discussions/29

SteveL-MSFT: Long term, we need to decide if we want to publish the schemas to https://github.com/Microsoft/json-schemas or keep it in the repo.  There's pros/cons for both.  Not a blocker for this PR.

miroman9364 (approved):

PR #203, opened 2023-09-24
(MAINT) Schema improvements

michaeltlombardi:
# PR Summary

This change adds a set of enhancement to the schemas, adding online documentation to the contextual help, new snippets, and reformatting the help.

This change set also includes the following updates to the schemas to reflect the changes for `alpha.3`:

1. Replaced the `manifestVersion` keyword with `$schema` pointing to the canonical bundled schema URI, per #199.
2. Replaced `set.preTest` with `set.implementsPretest`, per #197.
3. Replaces `args` with `env` for input kind, per #198.
4. Fixes erroneous munging of always-array keywords, like `enum` and `required`, to scalar values when the array only contains a single item.

## PR Context

This change helps improve the experience for configuration and manifest authors, improving their contextual help and giving them a faster way to add sections and entries to those files.

### DIFF
--- a/schemas/2023/08/bundled/outputs/resource/list.json
+++ b/schemas/2023/08/bundled/outputs/resource/list.json
@@ -113,18 +113,19 @@
       "description": "Defines the information DSC and integrating require to process and call a command-based DSC Resource.",
       "type": "object",
       "required": [
-        "manifestVersion",
+        "$schema",
         "type",
         "version",
         "get"
       ],
       "properties": {
-        "manifestVersion": {
-          "title": "Manifest Version",
-          "description": "The semver of the DSC Resource manifest schema to validate this manifest with.",
-          "$ref": "/PowerShell/DSC/main/schemas/2023/08/definitions/semver.json",
+        "$schema": {
+          "title": "Manifest Schema",
+          "description": "This property must be the canonical URL of the Command-based DSC Resource Manifest schema that the manifest is implemented for.",
+          "type": "string",
+          "format": "uri",
           "enum": [
-            "1.0"
+            "https://raw.githubusercontent.com/PowerShell/DSC/main/schemas/2023/08/bundled/resource/manifest.json"
           ]
         },
         "type": {
--- a/schemas/2023/08/bundled/outputs/resource/list.json
+++ b/schemas/2023/08/bundled/outputs/resource/list.json
@@ -632,13 +632,12 @@
       "$schema": "https://json-schema.org/draft/2020-12/schema",
       "$id": "https://raw.githubusercontent.com/PowerShell/DSC/main/schemas/2023/08/definitions/inputKind.json",
       "title": "Executable Command Input Type",
-      "description": "Defines how DSC should pass input to the command, either as arguments or JSON over stdin.",
+      "description": "Defines how DSC should pass input to the command, either as environment variables or JSON over stdin. When this value isn't defined, DSC doesn't send the resource any input.",
       "type": "string",
       "enum": [
-        "args",
+        "env",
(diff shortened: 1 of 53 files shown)


mgreenegit:
Looks like we need to regen again. Frequent changes. Want to put them in the same PR?
https://github.com/michaeltlombardi/DSC/blob/5e10fbbb995d1a7c8ce3f41cf248786603409810/schemas/2023/08/bundled/resource/manifest.json#L14

schemas/2023/08/bundled/outputs/resource/list.json
```diff
@@ -259,7 +260,7 @@
         "input": {
           "$ref": "/PowerShell/DSC/main/schemas/2023/08/definitions/inputKind.json"
         },
-        "implementsPretest": {
+        "preTest": {
```
SteveL-MSFT, line 263: Isn't this renaming reversed?

SteveL-MSFT (changes requested):

schemas/2023/08/bundled/outputs/resource/list.json
michaeltlombardi: @SteveL-MSFT - I went looking, and it seems like the output schemas were directly updated in #197, instead of the source schemas. When we regenerated them from source, those manual changes were lost. That change happened in this latest release, which I'll address in the next PR - or I can just add all of the schema changes for `alpha.3` into this PR, whichever works better.

michaeltlombardi: Went ahead and updated the schemas to reflect the alpha.3 changes.

michaeltlombardi:
@SteveL-MSFT, @mgreenegit - ended up pulling all the changes for the `alpha.3` schema updates into this PR, _should_ be good to go now - also fixed a bunch of errors in the generation process and confirmed the (bundled, at least) schemas are valid now.

schemas/src/definitions/inputKind.yaml
```diff
@@ -4,22 +4,35 @@ $id:     <HOST>/<PREFIX>/<VERSION>/definitions/inputKind.yaml
...
-      Indicates that the resource expects the properties of an instance to be specified with
-      command line arguments. This option isn't implemented yet.
-  # stdin
-  - |
-      Indicates that the resource expects a JSON blob representing an instance from `stdin`.
+  - | # env
+      _Environment variables_
+
+      > Indicates that the resource expects the properties of an instance to be specified as
+      > environment variables. This option only supports the following data types for instance
```
SteveL-MSFT, line 21: Should we be more explicit that JSON property names become the name of the env var (case preserving) and the value is as-is.  Arrays become comma delimited strings.

SteveL-MSFT (changes requested):

schemas/src/config/document.parameter.yaml
```diff
@@ -66,23 +74,49 @@ properties:
...
+            key: value
+          ```
         body:
-          '${1:metadataName}': { '${2:key}': '${3:value}' }
-      - label:       New metadata property (array)
-        description: Defines a new key-value pair for the metadata where the value is an array.
+          ${1:metadataName}:
+            ${2:key}: ${3:value}
+
+      - label: ' New metadata property (array)'
```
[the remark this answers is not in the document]
michaeltlombardi: @SteveL-MSFT - the extra whitespace is intentional, it moves them to the top of the IntelliSense completion list - otherwise they often end up hidden at the bottom, depending on the extensions you have installed.

SteveL-MSFT (approved):

Issue #202, opened 2023-09-23
Rename `_ensure` to `_exist`


SteveL-MSFT:
### Summary of the new feature / enhancement

`_ensure` is really a directive and not a property.  This makes it inconsistent when a resource receives `_ensure` and also needs to return it.  Resources should only return properties that pertain to the object.  When a resource returns:

```yaml
foo: bar
_ensure: present
```

Does it mean that the resource requested is present or is it just passing through the initial request?

To remove this ambiguity, I would propose since `_ensure` only has `absent` and `present` as valid values to rename to `_absent` which is `true` or `false`.  In this case, if the request wants to ensure the instance is absent, then it would set `_absent = true` and the resource (if it supports `_absent`) always returns an `_absent` property setting it to `true` or `false`.

In the case that `_absent` is missing from the request or the object, it defaults to `false`.

ThomasNieto:
I believe there were DSC resources that had non-standard values and maybe some in the Microsoft resource or examples.

As for the naming, would it be better to have `present` as the name with true or false instead of a double negative?

michaeltlombardi:
> _ensure is really a directive and not a property. 

This depends on perspective - it's a property of the instance-as-a-whole, not a setting of the instance, if you follow me. Most configuration management tools that use this idiom use it as a configurable property. The expectation is that the resource implementation returns `Absent` and null-values (or in the case of JSON, none of those values are in the blob) when the instance doesn't exist.

This is often used by tools for reporting messages about whether the operation was to create, update, or remove an instance - Puppet and Chef both do this iirc.

In that regard, I think the current implementation of `_ensure` as a manageable property is correct and reuses a common idiom. 

I'm not _against_ replacing it with a boolean property instead, but I would recommend we call it something like `exist` to be even more explicit and move further away from the idiom.

Something like:

```yaml
$id:         <HOST>/<PREFIX>/<VERSION>/resource/properties/exist.json
title:       Instance Existance
description: Indicates whether the DSC Resource instance should exist.
type:        boolean
default:     true
```

SteveL-MSFT:
Based on internal discussion, we decided to change this well-known property to `_exist` as a boolean.  Default value is `true` if not specified.

ThomasNieto:
How does this impact existing resources using ensure?

michaeltlombardi:
> How does this impact existing resources using ensure?
> ~ @ThomasNieto

[...]

Issue #162, opened 2023-08-22
Make `Microsoft.Windows/Registry` idiomatic


michaeltlombardi:
### Summary of the new feature / enhancement

> As a user, I want the `Microsoft.Windows/Registry` resource to behave idiomatically, returning a representation of the current state that matches the instance schema.

The resource currently returns the following JSON representations from `registry config get`:

- For a key that exists:

  ```json
  {
    "$id": "https://developer.microsoft.com/json-schemas/windows/registry/20230303/Microsoft.Windows.Registry.schema.json",
    "keyPath": "HKCU\\example"
  }
  ```

- For a key that doesn't exist:

  ```json
  {
    "$id": "https://developer.microsoft.com/json-schemas/windows/registry/20230303/Microsoft.Windows.Registry.schema.json",
    "keyPath": ""
  }
  ```

- For a value that exists:

  ```json
  {
    "$id": "https://developer.microsoft.com/json-schemas/windows/registry/20230303/Microsoft.Windows.Registry.schema.json",
    "keyPath": "HKCU\\example\\subkey",
    "valueName": "NewKeyAndValue",
    "valueData": {
      "String": "No errors"
    }
  }
  ```

- For a value that doesn't exist, `registry config get` errors with exit code `3`:

  ```console
  Error getting config: The registry key or value name does not exist.: Failed to query registry value
  ```

### Proposed technical implementation details (optional)

The idiomatic representation of the resource instance should always include the `keyPath`, even when the key doesn't exist. The instance should use the value of the `_ensure` property to indicate whether the instance is present or absent.

The idiomatic representations for an instance are:

- For a key that exists:

  ```json
  {
    "$id": "https://developer.microsoft.com/json-schemas/windows/registry/20230303/Microsoft.Windows.Registry.schema.json",
    "_ensure": "Present",
    "keyPath": "HKCU\\example"
  }
  ```

- For a key that doesn't exist:

  ```json
  {
    "$id": "https://developer.microsoft.com/json-schemas/windows/registry/20230303/Microsoft.Windows.Registry.schema.json",
    "_ensure": "Absent",
    "keyPath": "HKCU\\example"
  }
  ```

- For a value that exists:

  ```json
  {
    "$id": "https://developer.microsoft.com/json-schemas/windows/registry/20230303/Microsoft.Windows.Registry.schema.json",
    "_ensure": "Present",
    "keyPath": "HKCU\\example\\subkey",
    "valueName": "NewKeyAndValue",
[...]

PR #206, opened 2023-09-26
Change `_ensure` to `_exist`

SteveL-MSFT:
# PR Summary

Based on discussion, renaming `_ensure` which is a directive to `_exist` which is a property.  This ensures it can be returned properly in a response and used as a request.   Registry resource updated to match the semantics.

## PR Context

Fix https://github.com/PowerShell/DSC/issues/202
Fix https://github.com/PowerShell/DSC/issues/162

### DIFF
--- a/dsc_lib/src/dscresources/dscresource.rs
+++ b/dsc_lib/src/dscresources/dscresource.rs
@@ -262,7 +262,7 @@ pub fn get_diff(expected: &Value, actual: &Value) -> Vec<String> {
                 map.insert(key.clone(), value.clone());
             }
 
-            if actual[&key].is_null() {
+            if actual.is_object() && actual[&key].is_null() {
                 actual[key.clone()] = value.clone();
             }
         }
--- a/dsc_lib/src/dscresources/dscresource.rs
+++ b/dsc_lib/src/dscresources/dscresource.rs
@@ -267,7 +267,7 @@ pub fn get_diff(expected: &Value, actual: &Value) -> Vec<String> {
             }
         }
 
-        for (key, value) in map.iter() {
+        for (key, value) in &*map {
             if value.is_object() {
                 let sub_diff = get_diff(value, &actual[key]);
                 if !sub_diff.is_empty() {
--- a/dsc_lib/src/dscresources/dscresource.rs
+++ b/dsc_lib/src/dscresources/dscresource.rs
@@ -245,12 +245,8 @@ pub fn get_diff(expected: &Value, actual: &Value) -> Vec<String> {
     }
 
     if let Some(map) = expected.as_object() {
+        let mut handled_exist = false;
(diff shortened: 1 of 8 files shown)


dsc_lib/src/dscresources/dscresource.rs
```diff
@@ -275,6 +284,18 @@ pub fn get_diff(expected: &Value, actual: &Value) -> Vec<String> {
                 }
             }
         }
+
+        // handle the case where the actual object has `_exist` but wasn't specified in the expected object
+        if !handled_exist {
+            if let Some(actual_object) = actual.as_object() {
+                if actual_object.contains_key("_exist") {
+                    // if expected didn't have `_exist`, it is assumed to be `true` so we only handle `false` case
+                    if actual["_exist"].as_bool() != Some(true) {
```
SteveL-MSFT, line 293:
```suggestion
                    if actual["_exist"].as_bool() == Some(false) {
```

michaeltlombardi:
@SteveL-MSFT - I'll test tomorrow, but to clarify for the registry resource, my understanding is:

- If the desired state instance only defines the `keyPath`, not `valueName`, `_exist` indicates whether the key exists.
- If the desired state instance defines both `keyPath` and `valueName`, `_exist` indicates whether the value exists on that key. If the key exists and the value doesn't, `_exist` is `false`. If the key doesn't exist, neither does the value, so `_exist` is `false`. Only when both exist does the resource return `true`.

If so, rad, and I'll get those changes updated after this PR lands.

SteveL-MSFT:
@michaeltlombardi yes, that should be correct

anmenaga (approved):

dsc_lib/src/dscresources/dscresource.rs
```diff
@@ -245,12 +245,8 @@ pub fn get_diff(expected: &Value, actual: &Value) -> Vec<String> {
     }
```
SteveL-MSFT, line 245: Instead of the code needed to explicitly check `_exist`, simply add it to `expected` and `actual` if it doesn't already contain it with default value `true` and the of the validation works as-is

registry/src/config.rs
```diff
@@ -74,7 +66,7 @@ impl Default for Registry {
             key_path: String::new(),
             value_name: None,
             value_data: None,
-            ensure: None,
+            exist: None,
```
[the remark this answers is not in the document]
miroman9364: Talked with @SteveL-MSFT, he's going to add the defaults, but at a lower level so it doesn't force anything. For example, my suggestion would mean all responses have `_exist`, but the preferred implementation is only respond with `_exist` when it was explicitly part of the request, or when the request (expected) differs from the response (actual).

miroman9364 (approved):

PR #213, opened 2023-09-30
Terminate process tree on ctrl+c

SteveL-MSFT:
# PR Summary

- move `test_group_resource` under `tools` where all test specific code should go
- created `dsctest` exe to be used with `Test/Sleep` resource for manually validating `ctrl+c`.  Future test resources should be subcommands of this exe.
- fn to discover process tree under the current `dsc` process and terminate each one
- fix `build.ps1` to correctly replace `debug` and `release` paths on non-Windows
- add `InputKind = Arg` to allow passing the input JSON as an argument.  See `Test/Sleep` resource on example usage.
- added a few info traces to help see what was happening during validation

## PR Context

Terminate subprocesses upon `ctrl+c`


dsc_lib/src/dscresources/resource_manifest.rs
```diff
@@ -50,6 +50,9 @@ pub struct ResourceManifest {
 
 #[derive(Debug, Clone, PartialEq, Deserialize, Serialize, JsonSchema)]
 pub enum InputKind {
+    /// The input replaces arguments with this token in the command.
+    #[serde(rename = "arg")]
+    Arg(String),
```
michaeltlombardi, line 55: I'm not sure about this change/implementation from the user-facing perspective and the schema - it moves the schema from a set of enum values (in JSON Schema syntax) to a `oneOf` that is more complicated and harder for users to adjust/grok.

I would propose an enum value like `ReplaceArg` (`replaceArg` in the JSON) that _always_ replaces a static string argument in the list, instead of allowing an arbitrary string.

We could have that argument be something like `{{instance_json}}` to be very unlikely to conflict with any actual arguments.

Then the manifest would look like:

```json
"set": {
  "executable": "foo",
  "args": ["config", "set", "--input", "{{instance_json}}"],
  "input": "replaceArg"
}
```

From a schema perspective, we can then validate that when `input` is `replaceArg` that `args` includes the static string and not when `input` is any other value.

Are there cases where users will need to specify a custom value for the replacement string?

```yaml
allOf:
  - if: { properties: { input: { const: replaceArg } } }
    then:
      properties:
        args:
          contains:  { const: '{{instance_json}}' }
          minContains: 1
          maxContains: 1
    else:
      properties:
        args:
          not: { contains:  { const: '{{instance_json}}' } }
```

dsc_lib/src/dscresources/command_resource.rs
```diff
@@ -381,6 +409,7 @@ pub fn invoke_export(resource: &ResourceManifest, cwd: &str) -> Result<ExportRes
 /// Error is returned if the command fails to execute or stdin/stdout/stderr cannot be opened.
 #[allow(clippy::implicit_hasher)]
 pub fn invoke_command(executable: &str, args: Option<Vec<String>>, input: Option<&str>, cwd: Option<&str>, env: Option<HashMap<String, String>>) -> Result<(i32, String, String), DscError> {
+    debug!("Invoking command {} with args {:?}", executable, args);
```
anmenaga, line 412: what is your logic when to use `debug!` or `info!` for diagnostics?

tools/dsctest/src/sleep.rs
```diff
@@ -0,0 +1,13 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+use schemars::JsonSchema;
+use serde::{Deserialize, Serialize};
+
+/// Returns information about the operating system.
```
anmenaga, line 7: wrong comment

dsc/src/main.rs
```diff
@@ -72,10 +75,37 @@ fn main() {
...
 
+fn terminate_subprocesses(sys: &System, process: &Process) {
+    info!("Terminating subprocesses of process {} {}", process.name(), process.pid());
+    for subprocess in sys.processes().values().filter(|p| p.parent().map_or(false, |parent| parent == process.pid())) {
+        terminate_subprocesses(sys, subprocess);
+    }
+
+    info!("Terminating process {} {}", process.name(), process.pid());
+    if !process.kill() {
+        info!("Failed to terminate process {} {}", process.name(), process.pid());
```
anmenaga, line 105: This probably should be an `error!`.

miroman9364 (approved):
Code-wise, this looks good to me. If the arg processing needs to change, it can be done in an update.

dsc_lib/src/dscresources/command_resource.rs
SteveL-MSFT: `info` for me is going to be used for `verbose` which can be useful to end-users while `debug` is really only useful to developers of `dsc` itself

tools/dsctest/src/sleep.rs
SteveL-MSFT: Good catch, will fix

dsc/src/main.rs
SteveL-MSFT: Yes

dsc_lib/src/dscresources/resource_manifest.rs
SteveL-MSFT: @michaeltlombardi please open a new issue to discuss

dsc_lib/src/dscresources/command_resource.rs
anmenaga: ok; let's stick with this approach.

PR #214, opened 2023-10-02
Add `dsc new` sub-command to generate example configuration or resource manifest

SteveL-MSFT:
# PR Summary

As I'm writing test configurations and test resources often, it would be easier to generate a simple one I can modify rather than finding an existing file to modify.

Open to suggestions on the content.  cc @michaeltlombardi 

`dsc new -t configuration` returns:

```yaml
resources:
- type: Company/Dependency
  name: Dependency Resource
  properties:
    Property1: Value1
    Property2: Value2
- type: Company/Resource
  name: My Resource Instance
  dependsOn:
  - '[resourceId("Company/Dependency","Dependency Resource")]'
  properties:
    Property2: Value2
    Property1: Value1
```

`dsc new -t resource-manifest` returns:

```json
{
  "$schema": "https://raw.githubusercontent.com/PowerShell/DSC/main/schemas/2023/08/bundled/resource/manifest.json",
  "type": "Company/Resource",
  "version": "1.0.0",
  "description": "My Resource",
  "tags": [
    "Windows",
    "Linux"
  ],
  "get": {
    "executable": "myexe",
    "args": [
      "get"
    ],
    "input": "stdin"
  },
  "set": {
    "executable": "myexe",
    "args": [
      "set"
    ],
    "input": "stdin",
    "implementsPretest": false,
    "return": "state"
  },
  "test": {
    "executable": "myexe",
    "args": [
      "test"
    ],
    "input": "stdin",
    "return": "state"
  },
  "exitCodes": {
    "0": "Success",
    "1": "Failure"
  },
  "schema": {
    "embedded": {
      "$schema": "http://json-schema.org/draft-07/schema#"
    }
  }
}
```

### DIFF
--- a/dsc/Cargo.toml
+++ b/dsc/Cargo.toml
@@ -13,11 +13,12 @@ lto = true
 [dependencies]
 atty = { version = "0.2" }
 clap = { version = "4.1", features = ["derive"] }
-crossterm = { version = "0.26.1" }
+crossterm = { version = "0.27" }
 ctrlc = { version = "3.4.0" }
 dsc_lib = { path = "../dsc_lib" }
-jsonschema = "0.17"
-schemars = { version = "0.8.12" }
+jsonschema = { version = "0.17", features = ["draft202012"]}
+schemars = { version = "0.8", features = ["semver"]}
+semver = { version = "1.0" }
 serde = { version = "1.0", features = ["derive"] }
 serde_json = { version = "1.0", features = ["preserve_order"] }
 serde_yaml = { version = "0.9" }
--- a/dsc/src/args.rs
+++ b/dsc/src/args.rs
@@ -10,6 +10,12 @@ pub enum OutputFormat {
     Yaml,
 }
 
+#[derive(Debug, Clone, PartialEq, Eq, ValueEnum)]
+pub enum NewType {
+    Configuration,
+    ResourceManifest,
+}
+
 #[derive(Debug, Parser)]
 #[clap(name = "dsc", version = "3.0.0-alpha.4", about = "Apply configuration or invoke specific DSC resources", long_about = None)]
 pub struct Args {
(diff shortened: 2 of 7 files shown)


anmenaga:
I'm not sure about generating these templates from code.
Maybe an alternative is to have them as text files next to `dsc.exe` that `dsc new -t` command can read and display.
This would make updating them much easier. plus "finding an existing file to use as example" still works in this case.
A test case can make sure these file-templates satisfy the latest schema, to ensure that they up to date.

SteveL-MSFT:
This is intended to be sort of like `New-ModuleManifest` (except without all the parameters to fill in the template) as well as `dotnet new`.  The intent is not for learning, but to have a simple valid file that a configuration author or resource author can use as a starting point without having to grab the `$schema` URL from docs.

miroman9364:
For me, I don't think it makes sense to do this in code. I would prefer to just ship the examples? Long-term, I would prefer that we have a repo for the examples that people can get from the repo if they want, Clone and templatize as they please. We could then maybe have code to make it easy to get it from the repo without having to install git. (I'm pretty sure this is actually what `dotnet new` is doing, it's getting a template from a repo and then search replacing a few strings.)

Because this is code that produces a big string, it has a lot of disadvantages for maintenance. It is hard to read the code and imagine the output. Also, when the example needs to change, it's harder to change because you have to write code to change a string instead of just updating the string itself.

miroman9364 (approved):
I don't like this as a final solution. That said, I can live with it until a better solution becomes obvious.

michaeltlombardi:
> Long-term, I would prefer that we have a repo for the examples that people can get from the repo if they want, Clone and templatize as they please.

When I was at puppet, we worked on tooling just like what you're describing here for users—we provided some default templates, but folks could define their own templates/repositories.

We also used `go:embed` directives to have a backup template available for offline use—maybe we could use [rust-embed] for similar purpose here? That would make it easier to maintain the templates + enable us to ship them with the binary, not as a separate file. A pattern I used before was to lay the files down on first usage, so from then on they could be copied.

I think there's a broader discussion we could have about the DevX for a command like this, but the implementation path here is okay for now.

I do think it would be better if the embedded schema defaulted to 2020-12 and defined the `type` and `properties` keywords, so the manifest will pass the full schema by default.

[rust-embed]: https://github.com/pyrossh/rust-embed

Issue #130, opened 2023-08-03
Support a `--file` option for `dsc config *` commands


michaeltlombardi:
### Summary of the new feature / enhancement

As a user working with DSC Configuration documents, I want to be able to specify the path to a configuration document saved as a file so that I can point DSC to saved files instead of always piping/redirecting the contents of a document to DSC.

For example:

```sh
dsc config test --file ./app.dsc.config.yaml
dsc config test -f ../security/baseline.dsc.config.yaml
```

SteveL-MSFT:
I'll add both `-input` and `-inputFile` parameters.  These will be global, so it'll be before the subcommand.

PR #217, opened 2023-10-03
Add `--input` and `--input-file` to `dsc`

SteveL-MSFT:
# PR Summary

Add `-i | --input` and `-p | --input-file` as ways to pass input to `dsc` instead of just via stdin.  The two parameters are mutually exclusive.  Currently, stdin is ignored if either of these parameters are used.

## PR Context

Fix https://github.com/PowerShell/DSC/issues/130

### DIFF
--- a/dsc/Cargo.toml
+++ b/dsc/Cargo.toml
@@ -12,7 +12,7 @@ lto = true
 
 [dependencies]
 atty = { version = "0.2" }
-clap = { version = "4.1", features = ["derive"] }
+clap = { version = "4.4", features = ["derive"] }
 clap_complete = { version = "4.4" }
 crossterm = { version = "0.27" }
 ctrlc = { version = "3.4.0" }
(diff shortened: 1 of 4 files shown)


michaeltlombardi:
Non-blocking, but we may want to have a think about the UX for global options—currently, if I want to use both the format and input file options, I have to run a command like:

```powershell
dsc -f pretty-json -i $desired resource test -r Example.Go/TSToy
```

This isn't very ergonomic, esp since these options are provided before two layers of subcommands. From my perspective, this would be the more natural command:

```powershell
dsc resource test -f pretty-json -r Example.Go/TSToy -i $desired
```

Is it possible to define option flags on the subcommands that override/define the globals for convenience? I haven't delved too deeply into clap, but in cobra you can define persistent options that work this way (define once, available on all child commands).

SteveL-MSFT:
@michaeltlombardi let's cover that topic as part of the cmdline review meeting

miroman9364 (approved):
I think we should review some of the argument names we are using and compare how they match up against pre-existing tools. However, I think we need this functionality and we can always change the names later.

PR #222, opened 2023-10-06
Tree-sitter grammar for expressions

SteveL-MSFT:
# PR Summary

ARM expressions are used as values of JSON properties:

```json
{
  "foo": "[function(arg1, arg2)]"
}
```

They are identified by the square brackets.  To have a literal square bracket as a string instead of an expression, you need to double it: `[[this is just a string]` which would be used as `[this is just a string]`.  Anything else would be treated as a string literal.

Function args can take other functions.  Strings passed to functions need to be surrounded by single-quotes.

Functions can return JSON objects which can be accessed by dot-notation: `[functionReturnObject().prop1.prop2]`.

Tree-sitter grammar for expressions in DSC (as defined by ARM)
Includes the generated source code as it only needs to be generated if the grammar changes.  Since this is generated code, having it as `unclean` as clippy complains.

The only files worth reviewing (as they aren't generated) is `grammar.js` and the two test `.txt` files

### DIFF
--- a/tree-sitter-dscexpression/binding.gyp
+++ b/tree-sitter-dscexpression/binding.gyp
@@ -1,0 +1,19 @@
+{
+  "targets": [
+    {
+      "target_name": "tree_sitter_DSCExpression_binding",
+      "include_dirs": [
+        "<!(node -e \"require('nan')\")",
+        "src"
+      ],
+      "sources": [
+        "bindings/node/binding.cc",
+        "src/parser.c",
+        # If your language uses an external scanner, add it here.
+      ],
+      "cflags_c": [
+        "-std=c99",
+      ]
+    }
+  ]
+}
--- a/tree-sitter-dscexpression/grammar.js
+++ b/tree-sitter-dscexpression/grammar.js
@@ -1,0 +1,25 @@
+module.exports = grammar({
+  name: 'dscexpression',
+
+  rules: {
+    statement: $ => choice(
+      $.escapedStringLiteral,
+      $.expression,
+      $.stringLiteral,
+    ),
+    escapedStringLiteral: $ => token(seq('[[', /.*/)),
+    expression: $ => seq('[', $.function, optional($._members), ']'),
+    stringLiteral: $ => token(prec(-11, /.*/)),
+    function: $ => seq($.functionName, '(', optional($._arguments), ')'),
+    functionName: $ => /[a-zA-Z]+/,
+    _arguments: $ => seq($._argument, repeat(seq(',', $._argument))),
+    _argument: $ => choice($.function, $.string, $.number, $.boolean),
+    string: $ => seq("'", /[^']*/, "'"),
+    number: $ => /\d+/,
+    boolean: $ => choice('true', 'false'),
+    _members: $ => repeat1($._member),
+    _member: $ => seq('.', $.memberName),
+    memberName: $ => /[a-zA-Z0-9_-]+/,
+  }
+
+});
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,7 @@ Cargo.lock
 target
 bin/
 .DS_Store
+
+# Node.js generated files for tree-sitter
+build/
+node_modules/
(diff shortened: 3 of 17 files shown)


SteveL-MSFT:
Going to merge this so I can move forward with using it.  If there's any feedback on these changes, we can address them in a separate PR.

tree-sitter-dscexpression/grammar.js
```diff
@@ -0,0 +1,25 @@
...
+    stringLiteral: $ => token(prec(-11, /.*/)),
+    function: $ => seq($.functionName, '(', optional($._arguments), ')'),
+    functionName: $ => /[a-zA-Z]+/,
+    _arguments: $ => seq($._argument, repeat(seq(',', $._argument))),
+    _argument: $ => choice($.function, $.string, $.number, $.boolean),
+    string: $ => seq("'", /[^']*/, "'"),
+    number: $ => /\d+/,
+    boolean: $ => choice('true', 'false'),
+    _members: $ => repeat1($._member),
+    _member: $ => seq('.', $.memberName),
```
daxian-dbw, line 21: A member should be in the format `<target>.<member-name>`

tree-sitter-dscexpression/grammar.js
```diff
@@ -0,0 +1,25 @@
...
+      $.expression,
+      $.stringLiteral,
+    ),
+    escapedStringLiteral: $ => token(seq('[[', /.*/)),
+    expression: $ => seq('[', $.function, optional($._members), ']'),
+    stringLiteral: $ => token(prec(-11, /.*/)),
+    function: $ => seq($.functionName, '(', optional($._arguments), ')'),
+    functionName: $ => /[a-zA-Z]+/,
+    _arguments: $ => seq($._argument, repeat(seq(',', $._argument))),
+    _argument: $ => choice($.function, $.string, $.number, $.boolean),
```
daxian-dbw, line 16: Can argument be a member expression? Like `[funcA(funcB().member)]`.

tree-sitter-dscexpression/grammar.js
```diff
@@ -0,0 +1,25 @@
...
+    string: $ => seq("'", /[^']*/, "'"),
+    number: $ => /\d+/,
```
daxian-dbw, line 18: Maybe need to take into account decimal too.

daxian-dbw (commented):
Sorry for the late review today. Left a few comments, but looks good overall.

tree-sitter-dscexpression/binding.gyp
```diff
@@ -0,0 +1,19 @@
+{
+  "targets": [
+    {
+      "target_name": "tree_sitter_DSCExpression_binding",
+      "include_dirs": [
+        "<!(node -e \"require('nan')\")",
+        "src"
+      ],
+      "sources": [
+        "bindings/node/binding.cc",
```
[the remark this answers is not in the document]
SteveL-MSFT: Will remove this

tree-sitter-dscexpression/grammar.js
```diff
@@ -0,0 +1,25 @@
+module.exports = grammar({
+  name: 'dscexpression',
+
+  rules: {
+    statement: $ => choice(
+      $.escapedStringLiteral,
+      $.expression,
+      $.stringLiteral,
+    ),
+    escapedStringLiteral: $ => token(seq('[[', /.*/)),
```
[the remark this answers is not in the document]
SteveL-MSFT: As far as I can tell from ARM docs, it's only that start to indicate it's not an expression.  There isn't a corresponding closing `]]`

tree-sitter-dscexpression/grammar.js
SteveL-MSFT: `<target>` is only ever the return value of a function

tree-sitter-dscexpression/grammar.js
SteveL-MSFT: yes, good catch

tree-sitter-dscexpression/grammar.js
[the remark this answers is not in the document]
SteveL-MSFT: However, it looks like I did miss a test case `[test] string` should just be a string

tree-sitter-dscexpression/grammar.js
SteveL-MSFT: Best I can tell, decimals are not supported

PR #223, opened 2023-10-07
Update tree-sitter grammar and tests addressing Dongbo's feedback

SteveL-MSFT:
# PR Summary

- Remove `node` bindings as they aren't used
- Remove `pal` project from rust-analyzer as it causes it to crash on macOS
- Fix casing of generated binding
- Update grammar to support nested dot-notation: `[functionOne(functionTwo().member)]` which produces an `expression` node, so tests updated to account for this

Have not figured out yet how to have `[Test] string` just pass-through as a string literal.  cc @daxian-dbw


miroman9364 (approved):
LGTM

.gitignore
```diff
@@ -6,3 +6,4 @@ bin/
 # Node.js generated files for tree-sitter
 build/
 node_modules/
+node/
```
daxian-dbw, line 9: Maybe the `node` binding is needed to run `tree-sitter generate` and `tree-sitter test`? I cannot run this command locally (after clone your branch):
```
PS:45> npx tree-sitter generate
npm ERR! could not determine executable to run

npm ERR! A complete log of this run can be found in:
npm ERR!     C:\Users\dongbow\AppData\Local\npm-cache\_logs\2023-10-09T21_07_05_532Z-debug-0.log
```

tree-sitter-dscexpression/grammar.js
```diff
@@ -4,16 +4,17 @@ module.exports = grammar({
   rules: {
     statement: $ => choice(
       $.escapedStringLiteral,
-      $.expression,
+      $._expressionString,
       $.stringLiteral,
     ),
-    escapedStringLiteral: $ => token(seq('[[', /.*/)),
-    expression: $ => seq('[', $.function, optional($._members), ']'),
+    escapedStringLiteral: $ => token(prec(1, seq('[[', /.*?/, ']'))),
```
daxian-dbw, line 10: Regarding the `[test] string` input, ideally, this rule should be `seq(/^\[\[/, /.*?/, /]$/)`, however, tree-sitter doesn't support the `^` and `$` assertions today. According to https://github.com/tree-sitter/tree-sitter-cli/issues/53, we need to use a custom scanner in order to check for "starting with" or "ending with" something.

See the newer issue https://github.com/tree-sitter/tree-sitter/issues/1970 for additional details.

daxian-dbw (approved):
LGTM. Left a few comments.

tree-sitter-dscexpression/corpus/invalid_expressions.txt
```diff
@@ -73,3 +73,14 @@ This is a string
 
     (statement
       (stringLiteral))
+
+=====
+String starting with bracket
+=====
+
+[Test] string
```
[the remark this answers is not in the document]
SteveL-MSFT: Should be incomplete since if you wanted literal `[string` it has to be `[[string` I think.

tree-sitter-dscexpression/grammar.js
SteveL-MSFT: I saw that, but I don't think this will be a problem since the entire statement is a single line.

PR #225, opened 2023-10-10
Add initial library for `file` resource

miroman9364:
## PR Summary

- Add code to compute checksums for SHA-1, SHA-256 and SHA-512 hash algorithms.
- Add code to generalize checking the environment and prompting the user to attach a debugger.

## PR Context
The `file_lib` contains the implementation for the `file` resource.

# PR Summary


file_lib/src/checksum.rs
```diff
@@ -0,0 +1,98 @@
...
+pub enum Algorithm {
+    /// The SHA-1 algorithm. (Considered insecure.)
+    Sha1,
+    /// The SHA-256 algorithm.
+    Sha256,
+    /// The SHA-512 algorithm.
+    Sha512,
+}
+
+/// Compute the checksum for a vector of bytes using a given algorithm.
```
michaeltlombardi, line 19: 💜❗

These reference docs are great, doubly love the inclusion of examples-as-tests! This is a pretty straightforward function but this information and reference documentation will make it much easier to write the higher level docs when we get to the resource implementation. 😊

file_lib/src/checksum.rs
```diff
@@ -0,0 +1,98 @@
...
+        },
+        Algorithm::Sha512 => {
+            let mut hasher = Sha512::new();
+            hasher.update(bytes);
+            let result = hasher.finalize();
+
+            HexString::from_bytes(&result.to_vec()).as_string().to_lowercase()
+        }
+    }
+}
```
SteveL-MSFT, line 87: The duplicate code bothered me.  You can do something like this since each algorithm implements the `Digest` trait.

```suggestion
pub fn compute_checksum(bytes: &[u8], algorithm: Algorithm) -> String {
    match algorithm {
        Algorithm::Sha1 => {
            get_checksum(bytes, Sha1::default())
        },
        Algorithm::Sha256 => {
            get_checksum(bytes, Sha256::default())
        },
        Algorithm::Sha512 => {
            get_checksum(bytes, Sha512::default())
        }
    }
}

fn get_checksum<D>(bytes: &[u8], mut hasher: D) -> String
where D: digest::Digest {
    hasher.update(bytes);
    let result = hasher.finalize();
    HexString::from_bytes(&result.to_vec()).as_string().to_lowercase()
}
```

Since each one also implements the Update trait, the `where` clause helps the compiler to know that we're only interested in the `Digest` trait and use those functions.

file_lib/src/debug.rs
```diff
@@ -0,0 +1,43 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+use crossterm::event;
+use std::env;
+
```
SteveL-MSFT, line 6: nit, but get rid of these extra newlines

SteveL-MSFT (changes requested):
add this project to `build.ps1` so that it gets run by clippy pedantic

file_lib/src/debug.rs
```diff
@@ -0,0 +1,43 @@
...
+            loop {
+                let event = event::read().unwrap();
+                if let event::Event::Key(_key) = event {
+                    break;
+                }
+                eprintln!("Unexpected event: {event:?}");
+            }
+        }
+    }
+}
```
SteveL-MSFT, line 43: We should probably pull this into it's own crate to be used by different projects within this repo, but that can be done as a separate PR

miroman9364: Yeah, I wasn't sure where to put it. I didn't know if it belonged in `dsc_lib` or a standalone create, or something new like `dsc_utils`, and figured we could move it later. I have a similar refactor to break out the common code for reading STDIN into a function.

miroman9364:
Build file is updated, build and test clean.

file_lib/src/debug.rs
SteveL-MSFT: @miroman9364 can you open an issue so we refactor that common code, probably a `dsc_utils` crate makes sense.

SteveL-MSFT (approved):
LGTM, please open issue for the refactoring discussion

There's some clippy pedantic issues you need to fix

PR #231, opened 2023-10-12
Add configuration objects and JSON serialization

miroman9364:
# PR Summary

This change adds the configuration data for the **File** resource. This includes all the initial properties, and code to serialize/deserialize the configuration and the schema.

## PR Context

The `File` configuration will provide the user interface for configuring the **File** resource behavior.


file_lib/src/configuration.rs
```diff
@@ -0,0 +1,171 @@
...
+///
+/// * `path` - The path to the file.
+/// * `hash` - The hash of the file to either compare or compute.
+/// * `content` - The content to use when asserting or setting the desired state.
+/// * `exist` - The well-known flag indicating whether or not the file exists or should exist.
+#[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize, Deserialize, JsonSchema)]
+#[schemars(title = "DSC.FileConfiguration", description = "File resource configuration.")]
+pub struct File {
+    pub path: String,
+    #[serde(rename = "hash", skip_serializing_if = "Option::is_none")]
```
SteveL-MSFT, line 19: Looks like `rename` is not needed here

```suggestion
    #[serde(skip_serializing_if = "Option::is_none")]
```

file_lib/src/configuration.rs
```diff
@@ -0,0 +1,171 @@
...
+    /// assert!(json.contains(EXPECTED_CHECKSUM));
+    /// assert!(json.contains(EXPECTED_CONTENT));
+    /// ```
+    #[must_use]
+    pub fn to_json(&self) -> String {
+        match serde_json::to_string(self) {
+            Ok(json) => json,
+            Err(e) => {
+                eprintln!("Failed to serialize to JSON: {e}");
+                String::new()
```
SteveL-MSFT, line 64: Does it make sense to return an empty string if it fails to serialize?  Seems better to return a `Result<>`.

file_lib/src/configuration.rs
```diff
@@ -0,0 +1,171 @@
...
+    /// assert_eq!(hash.algorithm, EXPECTED_ALGORITHM);
+    /// assert_eq!(hash.checksum.unwrap(), EXPECTED_CHECKSUM);
+    /// ```
+    #[must_use]
+    pub fn from_json(json: &str) -> Option<File> {
+        match serde_json::from_str(json) {
+            Ok(file) => Some(file),
+            Err(e) => {
+                eprintln!("Failed to deserialize from JSON: {e}");
+                None
```
SteveL-MSFT, line 105: Same here, return a `Result<>`

SteveL-MSFT (changes requested):

Issue #236, opened 2023-10-18
Script based resources do not enforce schema validation


mgreenegit:
### Prerequisites

- [X] Write a descriptive title.
- [X] Make sure you are able to repro it on the latest version
- [X] Search the existing issues.

### Steps to reproduce

'{"Name": "bits","State": "Running"}' | dsc resource get -r PsDscResources/MSFT_ServiceResource

### Expected behavior

```console
actualState:
  DesktopInteract: false
  Name: bits
  BuiltInAccount: LocalSystem
  State: Stopped
  Ensure: Present
  Description: Transfers files in the background using idle network bandwidth. If the service is disabled, then any applications that depend on BITS, such as Windows Update or MSN Explorer, will be unable to automatically download programs and other information.
  StartupType: Manual
  Dependencies:
  - RpcSs
  Path: C:\WINDOWS\System32\svchost.exe -k netsvcs -p
  DisplayName: Background Intelligent Transfer Service
```

### Actual behavior

```console
Error: Command: Resource 'DSC/PowerShellGroup' [Exit code 1] Get-TargetResource: A parameter cannot be found that matches parameter name 'State'.
```

### Error details

```console
This happens because script based resources do not enforce parameter validation, and in many cases "Get" does not implement all parameters.
Example:
https://github.com/PowerShell/PSDscResources/blob/7064eda52d939a4a3ce40e1f38756cfe6a09acfd/DscResources/MSFT_ServiceResource/MSFT_ServiceResource.psm1#L37

We could solve this issue in PSDesiredStateConfiguration, but then DSCv3 would require that version or later. When we have created dependencies like that in the past, it made adoption more difficult.
```

### Environment data

```PowerShell
7.3.8
```

### Version

0.2.0

PR #237, opened 2023-10-18
Workaround script resource parameter validation

mgreenegit:
Resolves https://github.com/PowerShell/DSC/issues/236


mgreenegit:
Calling Get-DscResource is a performance hit. We could optimize and populate more information to cache.

powershellgroup/powershellgroup.resource.ps1
```diff
@@ -55,45 +55,45 @@ Import-Module $DscModule
...
+    RefreshCache
     #TODO: following should be added to debug stream of every operation
     #$m = gmo PSDesiredStateConfiguration
     #$r += @{"DebugInfo"=@{"ModuleVersion"=$m.Version.ToString();"ModulePath"=$m.Path;"PSVersion"=$PSVersionTable.PSVersion.ToString();"PSPath"=$PSHome}}
     #$r[0] | ConvertTo-Json -Compress -Depth 3
-    foreach ($r in $DscResources)
+    foreach ($r in $script:ResourceCache.keys)
     {
-        if ($r.ImplementedAs -eq "Binary")
+        if ($script:ResourceCache[$r].ImplementedAs -eq "Binary")
```
anmenaga, line 66: It would be better perf and will look slightly easier if `$script:ResourceCache[$r]` would be assigned to a var in the top of the loop and used in many places below.

Another idea: it is probably easier to iterate over cache values which are objects that we need/use below:
`foreach ($r in $script:ResourceCache.Values)`

powershellgroup/powershellgroup.resource.ps1
```diff
@@ -152,6 +163,17 @@ elseif ($Operation -eq 'Get')
         {
             $inputht = @{}
             $ResourceTypeName = ($inputobj_pscustomobj.type -split "/")[1]
+
+            if ($r.ImplementationDetail -eq 'ScriptBased') {
```
anmenaga, line 167: `$r` is not defined here.  
`$cachedResourceInfo` is probably what is meant here.

anmenaga (changes requested):
Minor update needed.

PR #238, opened 2023-10-18
(GH-224) Convert schema constants to enums

### LINKED ISSUE #224: Update the `$schema` URIs for manifests to `2023/10`

In the most recent set of unreleased changes, we have backwards-incompatible updates to the schemas. We have so far been updating the schemas in place during early development, but we will need to update the schemas in a more graceful way eventually.

We should practice doing so now while the impact is minimal to figure out a workflow and practice that suits our needs so that when many users and integrating developers are depending on the schemas, they can do so with confidence.

We should update the schema constants for configuration documents and resource manifests from string constants to enum values:

https://github.com/PowerShell/DSC/blob/94c0f632f61aa68e3fc9914ce43e650573ac2622/dsc_lib/src/configure/config_doc.rs#L81

https://github.com/PowerShell/DSC/blob/94c0f632f61aa68e3fc9914ce43e650573ac2622/dsc_lib/src/dscresources/resource_manifest.rs#L191

This is pseudocode, since I'm not quite familiar with Rust yet, but something _like_:

```rust
pub enum SchemaUri {
    #[serde(rename = "https://raw.githubusercontent.com/PowerShell/DSC/main/schemas/2023/10/config/document.json")]
    Version2023_10 = "https://raw.githubusercontent.com/PowerShell/DSC/main/schemas/2023/10/config/document.json",
    #[serde(rename = "https://raw.githubusercontent.com/PowerShell/DSC/main/schemas/2023/08/config/document.json")]
    Version2023_08 = "https://raw.githubusercontent.com/PowerShell/DSC/main/schemas/2023/08/config/document.json",
}
```

To allow users to specify a valid schema. We should begin considering whether/how to address schema changes between versions, but I don't think we need any _specific_ handling at this time.

michaeltlombardi:
# PR Summary

This change converts the string constants for the canonical schema URIs for both the configuration document and resource manifest schemas to enums, defining the valid URIs as variants.

For each supported version of a schema, there are three variants:

1. `Version<YYYY>_<MM>` - The canonical URI to the non-bundled schema. Using this schema requires retrieving every referenced schema, but it is also the least-munged.
1. `Bundled<YYYY>_<MM>` - The canonical URI to the bundled schema. When using this schema, only the bundled schema needs to be retrieved.
1. `VSCode<YYYY>_<MM>` - The canonical URI to the enhanced authoring schema. This schema is specifically implemented to support contextual help and DevX when authoring in VS Code. It is much larger than the other schemas.

Which gives six variants:

- `Version2023_08`, `Bundled2023_08`, and `VSCode2023_08` for 2023/08
- `Version2023_10`, `Bundled2023_10`, and `VSCode2023_10` for 2023/10

Currently, the specified schema is not actually used by DSC to validate the document or manifest. In the future, we could use the version to identify how to validate and process the data.

## PR Context

With the breaking changes made in recent PRs, and keeping in mind that we will eventually need to handle different versions of the schema, this PR:

- Lays the groundwork for handling multiple valid versions of the schema, even though both of these will probably be dropped when we move to beta
- Resolves #224

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -16,11 +16,11 @@
     "json.schemas": [
         {
             "fileMatch": ["**/*.dsc.resource.json"],
-            "url": "/schemas/2023/08/bundled/resource/manifest.vscode.json"
+            "url": "/schemas/2023/10/bundled/resource/manifest.vscode.json"
         }
     ],
     "yaml.schemas": {
-        "schemas/2023/08/bundled/config/document.vscode.json": "**.dsc.{yaml,yml,config.yaml,config.yml}",
-        "schemas/2023/08/bundled/resource/manifest.vscode.json": "**.dsc.resource.{yaml,yml}"
+        "schemas/2023/10/bundled/config/document.vscode.json": "**.dsc.{yaml,yml,config.yaml,config.yml}",
+        "schemas/2023/10/bundled/resource/manifest.vscode.json": "**.dsc.resource.{yaml,yml}"
     }
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,7 +28,15 @@ changes since the last release, see the [diff on GitHub][unreleased].
 
 ### Changed
 
-- Replaced the `_ensure` well-known property with the boolean [_exist][21] property. This improves
+- Updated the canonical version of the schema URIs from `2023/08` to `2023/10`, as this release
+  includes breaking changes for the schemas.
+
+  As part of this change, the `$schema` keyword for both [configuration documents][21] and
+  [resource manifests][22] accepts any valid URI for the schemas, instead of only one. Now, you
+  can set the value for the keyword to the unbundled schema, the bundled schema, or the enhanced
+  authoring schema for any supported version.
+
+- Replaced the `_ensure` well-known property with the boolean [_exist][23] property. This improves
   the semantics for users and simplifies implementation for resources, replacing the string enum
   values `Present` and `Absent` with `true` and `false` respectively.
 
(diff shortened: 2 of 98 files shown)


michaeltlombardi:
Working through this PR, I think it would be worth considering a shortened links, like:

```diff
# Unbundled configuration document schema
-https://raw.githubusercontent.com/PowerShell/DSC/main/schemas/2023/10/config/document.json
+https://aka.ms/dsc/schemas/2023/10/config/document.json

# Bundled configuration document schema
-https://raw.githubusercontent.com/PowerShell/DSC/main/schemas/2023/10/bundled/config/document.json
+https://aka.ms/dsc/schemas/2023/10/config/document.bundled.json

# Enhanced authoring configuration document schema
-https://raw.githubusercontent.com/PowerShell/DSC/main/schemas/2023/10/bundled/config/document.vscode.json
+https://aka.ms/dsc/schemas/2023/10/config/document.vscode.json

# Unbundled configuration document schema
-https://raw.githubusercontent.com/PowerShell/DSC/main/schemas/2023/10/resource/manifest.json
+https://aka.ms/dsc/schemas/2023/10/resource/manifest.json

# Bundled configuration document schema
-https://raw.githubusercontent.com/PowerShell/DSC/main/schemas/2023/10/bundled/resource/manifest.json
+https://aka.ms/dsc/schemas/2023/10/resource/manifest.bundled.json

# Enhanced authoring configuration document schema
-https://raw.githubusercontent.com/PowerShell/DSC/main/schemas/2023/10/bundled/resource/manifest.vscode.json
+https://aka.ms/dsc/schemas/2023/10/resource/manifest.vscode.json
```

SteveL-MSFT (approved):

PR #240, opened 2023-10-19
Resource discovery optimizations

### LINKED ISSUE #121: Make resource discovery not terminating when hitting an error depending on use case

Currently, during discovery if there is any error, the entire operation fails.  Instead we need to consider the usage:

- if part of a configuration, if the error is not related to the resources actually being used, we can ignore any errors and only emit a message to stderr if `--verbose` is used
- if part of `dsc resource list`, we can emit a warning type message to stderr but continue discovery listing the results
- if direct resource invocation, if that resource is being used, then fail, otherwise same as config above

_No response_

### LINKED ISSUE #173: Discovery optimizations

Current design always performs a complete discovery, this will need to be changed with the following proposed optimizations:

1. When using `dsc resource get -r <resourceName>` if the <resourceName> is found, we don't continue with discovery.
2. If a config is provided, if a resource provider isn't used, don't call it to do discovery

_No response_

anmenaga:
# PR Summary

Fix #121 
Fix #173 

## PR Context

This PR changes resource discovery behavior as per #121 , 
There is also some perf optimizations added such as:
* early termination of discovery once all required resources for the operation are found;
* discovery of provider-based resources is done after discovery of native resources;
* no longer use RegEx for type name matching when not needed (previously was used for all operations).


dsc/src/main.rs
```diff
@@ -26,7 +26,7 @@ fn main() {
     check_debug();
 
     // create subscriber that writes all events to stderr
-    let subscriber = tracing_subscriber::fmt().pretty().with_writer(std::io::stderr).finish();
+    let subscriber = tracing_subscriber::fmt().pretty()/*.with_max_level(Level::DEBUG)*/.with_writer(std::io::stderr).finish();
```
SteveL-MSFT, line 29: We should remove this and just have the separate PR to enable setting logging level to `dsc`

dsc/src/resource_command.rs
```diff
@@ -14,14 +15,25 @@ use dsc_lib::{
 };
 use std::process::exit;
 
-pub fn get(dsc: &mut DscManager, resource: &str, input: &Option<String>, stdin: &Option<String>, format: &Option<OutputFormat>) {
+/// Get operation.
+///
+/// # Panics
+///
+/// Will panic if provider-based resource is not found.
```
SteveL-MSFT, line 22: Should never panic as we should exit gracefully

dsc/src/resource_command.rs
```diff
@@ -14,14 +15,25 @@ use dsc_lib::{
...
+        return
+    };
+
     debug!("resource.type_name - {} implemented_as - {:?}", resource.type_name, resource.implemented_as);
-    if let Some(requires) = resource.requires {
-        input = add_type_name_to_json(input, resource.type_name);
-        resource = get_resource(dsc, &requires);
+    if let Some(requires) = &resource.requires {
+        input = add_type_name_to_json(input, resource.type_name.clone());
+        resource = get_resource(dsc, requires).unwrap();
```
SteveL-MSFT, line 36: this is the line that can panic, you should do something like:

```rust
if let Some(resource) = get_resource(dsc, requires) else {
  error...
}
```

dsc/src/resource_command.rs
```diff
@@ -71,20 +86,29 @@ pub fn get_all(dsc: &mut DscManager, resource: &str, _input: &Option<String>, _s
...
+    };
 
     debug!("resource.type_name - {} implemented_as - {:?}", resource.type_name, resource.implemented_as);
 
-    if let Some(requires) = resource.requires {
-        input = add_type_name_to_json(input, resource.type_name);
-        resource = get_resource(dsc, &requires);
+    if let Some(requires) = &resource.requires {
+        input = add_type_name_to_json(input, resource.type_name.clone());
+        resource = get_resource(dsc, requires).unwrap();
```
SteveL-MSFT, line 111: same here, there generally should not be any `unwrap()` calls outside of unittests

dsc/src/resource_command.rs
```diff
@@ -106,15 +130,24 @@ pub fn set(dsc: &mut DscManager, resource: &str, input: &Option<String>, stdin:
...
```
SteveL-MSFT, line 150: same here

dsc/src/resource_command.rs
```diff
@@ -177,34 +216,25 @@ pub fn export(dsc: &mut DscManager, resource: &str, format: &Option<OutputFormat
...
 }
 
-pub fn get_resource(dsc: &mut DscManager, resource: &str) -> DscResource {
+#[must_use]
+pub fn get_resource<'a>(dsc: &'a DscManager, resource: &str) -> Option<&'a DscResource> {
+    //TODO: add dinamically generated resource to dsc
     // check if resource is JSON or just a name
-    match serde_json::from_str(resource) {
-        Ok(resource) => resource,
+    /*match serde_json::from_str(resource) {
```
SteveL-MSFT, line 223: remove commented out code

dsc/src/resource_command.rs
```diff
@@ -177,34 +216,25 @@ pub fn export(dsc: &mut DscManager, resource: &str, format: &Option<OutputFormat
     write_output(&json, format);
 }
 
-pub fn get_resource(dsc: &mut DscManager, resource: &str) -> DscResource {
+#[must_use]
+pub fn get_resource<'a>(dsc: &'a DscManager, resource: &str) -> Option<&'a DscResource> {
```
SteveL-MSFT, line 220: not clear to me why lifetime annotation is needed here

dsc/src/subcommand.rs
```diff
@@ -154,23 +154,29 @@ pub fn config(subcommand: &ConfigSubCommand, format: &Option<OutputFormat>, stdi
...
+            config_export(&mut configurator, format);
         }
     }
 }
 
+/// Validate configuration.
+///
+/// # Panics
+///
+/// Will panic if resource is not found.
```
SteveL-MSFT, line 178: same as before, handle unwraps()

dsc/src/subcommand.rs
```diff
@@ -272,15 +278,18 @@ pub fn validate_config(config: &str) {
...
-                    let validation = compiled_schema.validate(&properties);
-                    if let Err(err) = validation {
-                        let mut error = String::new();
-                        for e in err {
-                            error.push_str(&format!("{e} "));
-                        }
-                        error!("Error: Resource {type_name} failed validation: {error}");
-                        exit(EXIT_VALIDATION_FAILED);
-                    }
+                    let _result: Result<(), ValidationError> = match compiled_schema.validate(&properties) {
```
SteveL-MSFT, line 281: Why was this syntax changed from the prior one which is more idiomatic and concise?

dsc_lib/src/discovery/command_discovery.rs
```diff
@@ -71,83 +51,136 @@ impl ResourceDiscovery for CommandDiscovery {
...
                 Err(e) => {
-                    self.discovery_messages.push(StreamMessage::new_error(
-                        format!("Could not start {}: {}", list_command.executable, e),
-                        Some(provider_type_name.clone()),
-                        Some(provider_path.clone())));
-
+                    // In case of "resource list" operation - print failure from provider as warning
+                    // In case of other resource/config operations:
+                    // print failure from provider as error because this provider was specifically requested by current resource/config operation
+                    if return_all_resources { warn!("Could not start {}: {}", list_command.executable, e); } else { error!("Could not start {}: {}", list_command.executable, e); }
```
SteveL-MSFT, line 117: follow the style convention and put on multiple lines

dsc_lib/src/discovery/command_discovery.rs
```diff
@@ -71,83 +51,136 @@ impl ResourceDiscovery for CommandDiscovery {
...
                     continue;
                 },
             };
 
             if exit_code != 0 {
-                self.discovery_messages.push(StreamMessage::new_error(
-                    format!("Provider failed to list resources with exit code {exit_code}: {stderr}"),
-                    Some(provider_type_name.clone()),
-                    Some(provider_path.clone())));
+                    // In case of "resource list" operation - print failure from provider as warning
```
SteveL-MSFT, line 123: looks too far indented?

dsc_lib/src/discovery/command_discovery.rs
```diff
@@ -71,83 +51,136 @@ impl ResourceDiscovery for CommandDiscovery {
...
+                    // In case of other resource/config operations:
+                    // print failure from provider as error because this provider was specifically requested by current resource/config operation
+                    if return_all_resources { warn!("Provider failed to list resources with exit code {exit_code}: {stderr}"); } else { error!("Provider failed to list resources with exit code {exit_code}: {stderr}"); }
```
SteveL-MSFT, line 126: put on multiple lines

dsc_lib/src/discovery/command_discovery.rs
```diff
@@ -71,83 +51,136 @@ impl ResourceDiscovery for CommandDiscovery {
...
             for line in stdout.lines() {
                 match serde_json::from_str::<DscResource>(line){
                     Result::Ok(resource) => {
                         if resource.requires.is_none() {
-                            self.discovery_messages.push(StreamMessage::new_error(
-                                DscError::MissingRequires(provider.clone(), resource.type_name.clone()).to_string(),
-                                Some(resource.type_name.clone()),
-                                Some(resource.path.clone())));
-
+                            if return_all_resources { warn!("{}", DscError::MissingRequires(provider.clone(), resource.type_name.clone()).to_string()); } else { error!("{}", DscError::MissingRequires(provider.clone(), resource.type_name.clone()).to_string()); }
```
SteveL-MSFT, line 133: same

dsc_lib/src/discovery/command_discovery.rs
```diff
@@ -71,83 +51,136 @@ impl ResourceDiscovery for CommandDiscovery {
...
+                            }
+                        }
                     },
                     Result::Err(err) => {
-                        self.discovery_messages.push(StreamMessage::new_error(
-                            format!("Failed to parse resource: {line} -> {err}"),
-                            Some(provider_type_name.clone()),
-                            Some(provider_path.clone())));
-
+                        if return_all_resources { warn!("Failed to parse resource: {line} -> {err}"); } else { error!("Failed to parse resource: {line} -> {err}"); }
```
SteveL-MSFT, line 153: same

dsc_lib/src/discovery/mod.rs
```diff
@@ -5,82 +5,99 @@ mod command_discovery;
...
+    /// List operation.
     ///
-    /// # Errors
+    /// # Panics
     ///
-    /// This function will return an error if the underlying discovery fails.
-    pub fn initialize(&mut self) -> Result<(), DscError> {
-        // TODO: this vec is leftover when we had multiple discovery types, we should
-        //       probably just have a single command discovery type
+    /// Will panic if `RegEx` creation fails.
```
SteveL-MSFT, line 34: handle and fail gracefully instead of panic

dsc_lib/src/discovery/mod.rs
```diff
@@ -5,82 +5,99 @@ mod command_discovery;
...
         for mut discovery_type in discovery_types {
-            discovery_type.initialize()?;
-            discovery_type.print_initialization_messages(StreamMessageType::Warning, StreamMessageType::Warning)?;
-            
-            let discovered_resources = discovery_type.discover();
+
+            let discovered_resources = match discovery_type.list_available_resources() {
+                Ok(value) => value,
+                Err(err) => {
+                        error!("{err}");
```
SteveL-MSFT, line 59: indent too much here

dsc_lib/src/discovery/mod.rs
```diff
@@ -5,82 +5,99 @@ mod command_discovery;
...
-        }
+        let discovery_types: Vec<Box<dyn ResourceDiscovery>> = vec![
+            Box::new(command_discovery::CommandDiscovery::new()),
+        ];
 
-        ResourceIterator::new(resources)
+        let mut remaining_required_resource_types = required_resource_types.to_owned();
+        for mut discovery_type in discovery_types {
+
+            let discovered_resources = match discovery_type.discover_resources(remaining_required_resource_types.clone()) {
```
SteveL-MSFT, line 88: we should be able to pass by ref instead of cloning

SteveL-MSFT (changes requested):

dsc/src/resource_command.rs
anmenaga: Agreed.

dsc/src/resource_command.rs
anmenaga: Updated in other way. "let Some" syntax creates unused variable while I need this one to be overwritten here and used later.

dsc/src/resource_command.rs
anmenaga: Updated

dsc/src/resource_command.rs
anmenaga: Updated

dsc/src/resource_command.rs
anmenaga: removed.

dsc/src/resource_command.rs
anmenaga: "missing lifetime specifier
this function's return type contains a borrowed value, but the signature does not say whether it is borrowed from `dsc` or `resource`"

dsc/src/subcommand.rs
anmenaga: Updated.

dsc/src/subcommand.rs
anmenaga: not sure; maybe rebasing artifact;
Changed it to the original.

dsc_lib/src/discovery/command_discovery.rs
anmenaga: Updated.

dsc_lib/src/discovery/command_discovery.rs
anmenaga: Updated.

dsc_lib/src/discovery/command_discovery.rs
anmenaga: Updated.

dsc_lib/src/discovery/command_discovery.rs
anmenaga: Updated.

dsc_lib/src/discovery/command_discovery.rs
anmenaga: Updated.

dsc/src/main.rs
anmenaga: ok. Removed.

dsc_lib/src/discovery/mod.rs
anmenaga: updated

dsc_lib/src/discovery/mod.rs
anmenaga: updated.

dsc_lib/src/discovery/mod.rs
anmenaga: Updated.

dsc/tests/dsc_args.tests.ps1
```diff
@@ -61,7 +61,7 @@ Describe 'config argument tests' {
 '@ }
     ) {
         param($text)
-        $output = $text | dsc resource get -r *registry
```
SteveL-MSFT, line 64: We should have a specific case for wildcard searching if you're replacing these

dsc_lib/src/configure/mod.rs
```diff
@@ -240,23 +238,29 @@ impl Configurator {
...
     }
 
-    fn validate_config(&self) -> Result<(Configuration, Vec<ResourceMessage>, bool), DscError> {
+    fn validate_config(&mut self) -> Result<(Configuration, Vec<ResourceMessage>, bool), DscError> {
         let config: Configuration = serde_json::from_str(self.config.as_str())?;
         let mut messages: Vec<ResourceMessage> = Vec::new();
         let mut has_errors = false;
+        
+        // Perform discovery of resources used in config
+        let required_resources = config.resources.iter().map(|p| p.resource_type.to_lowercase()).collect::<Vec<String>>();
```
SteveL-MSFT, line 258: This doesn't prune duplicates right?  We should ensure the list is unique before discovery

SteveL-MSFT (changes requested):

dsc/src/resource_command.rs
```diff
@@ -14,14 +15,24 @@ use dsc_lib::{
 };
 use std::process::exit;
 
-pub fn get(dsc: &mut DscManager, resource: &str, input: &Option<String>, stdin: &Option<String>, format: &Option<OutputFormat>) {
+pub fn get(dsc: &DscManager, resource_str: &str, input: &Option<String>, stdin: &Option<String>, format: &Option<OutputFormat>) {
```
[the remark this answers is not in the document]
anmenaga: Updated to 'resource_type' throughout this file.

dsc/src/resource_command.rs
```diff
@@ -14,14 +15,24 @@ use dsc_lib::{
...
+        return
+    };
+
     debug!("resource.type_name - {} implemented_as - {:?}", resource.type_name, resource.implemented_as);
-    if let Some(requires) = resource.requires {
-        input = add_type_name_to_json(input, resource.type_name);
-        resource = get_resource(dsc, &requires);
+    if let Some(requires) = &resource.requires {
+        input = add_type_name_to_json(input, resource.type_name.clone());
+        if let Some(pr) = get_resource(dsc, requires) {
```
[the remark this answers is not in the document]
anmenaga: I've tried it; the var will be scope to `let Some...`, while in this case I need to overwrite the one in outer scope.

dsc/tests/dsc_args.tests.ps1
anmenaga: With this change `get/set/test` operations no longer support wildcards. only `List` does.
I had to do this to simplify current changes.
I've created issue #247 to track this if we need to get this back.

dsc_lib/src/configure/mod.rs
anmenaga: Good point;  Added removal of duplicated.

SteveL-MSFT:
@michaeltlombardi as part oft his change, individual resource execution will no longer take a wildcard.  There's a new issue opened if we need this back.

SteveL-MSFT (approved):

PR #241, opened 2023-10-20
Implement use of tree-sitter parser

SteveL-MSFT:
# PR Summary

- The primary change is to use the generated tree-sitter parser.
- Grammar was updated to have parent node for `arguments` and `member_access`
- Grammar updated so that `string` arguments don't include the quotes
- The statement parser creates `expression` which can contain `functions`
- Added `base64` and `concat` functions for testing
- Some clippy fixes as the rules appeared to have been updated

Note that this parser is not integrated yet with config.  That will come next replacing `resourceId()` with real function.

### DIFF
--- a/dsc_lib/src/functions/mod.rs
+++ b/dsc_lib/src/functions/mod.rs
@@ -1,0 +1,82 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+use std::collections::HashMap;
+use crate::DscError;
+use crate::parser::functions::{FunctionArg, FunctionResult};
+use tracing::debug;
+
+pub mod base64;
+pub mod concat;
+
+#[derive(Debug, PartialEq)]
+pub enum AcceptedArgKind {
+    String,
+    Integer,
+    Boolean,
+}
+
+pub trait Function {
+    fn min_args(&self) -> usize;
+    fn max_args(&self) -> usize;
+    fn accepted_arg_types(&self) -> Vec<AcceptedArgKind>;
+    fn invoke(&self, args: &Vec<FunctionArg>) -> Result<FunctionResult, DscError>;
+}
+
+pub struct FunctionDispatcher {
+    functions: HashMap<String, Box<dyn Function>>,
+}
+
+impl FunctionDispatcher {
+    pub fn new() -> Self {
+        let mut functions: HashMap<String, Box<dyn Function>> = HashMap::new();
+        functions.insert("base64".to_string(), Box::new(base64::Base64{}));
+        functions.insert("concat".to_string(), Box::new(concat::Concat{}));
+        Self {
+            functions,
+        }
+    }
+
+    pub fn invoke(&self, name: &str, args: &Vec<FunctionArg>) -> Result<FunctionResult, DscError> {
+        let function = self.functions.get(name);
+        match function {
+            Some(function) => {
+                // check if arg number are valid
+                if args.len() < function.min_args() {
+                    return Err(DscError::Parser(format!("Function {0} requires at least {1} arguments", name, function.min_args())));
+                }
+                if args.len() > function.max_args() {
+                    return Err(DscError::Parser(format!("Function {0} requires at most {1} arguments", name, function.max_args())));
+                }
+                // check if arg types are valid
+                for arg in args {
+                    match arg {
+                        FunctionArg::String(_) => {
+                            if !function.accepted_arg_types().contains(&AcceptedArgKind::String) {
+                                return Err(DscError::Parser(format!("Function {0} does not accept string arguments", name)));
+                            }
+                        },
+                        FunctionArg::Integer(_) => {
+                            if !function.accepted_arg_types().contains(&AcceptedArgKind::Integer) {
+                                return Err(DscError::Parser(format!("Function {0} does not accept integer arguments", name)));
+                            }
+                        },
+                        FunctionArg::Boolean(_) => {
+                            if !function.accepted_arg_types().contains(&AcceptedArgKind::Boolean) {
+                                return Err(DscError::Parser(format!("Function {0} does not accept boolean arguments", name)));
+                            }
+                        },
+                        FunctionArg::Expression(_) => {
+                            debug!("An expression was not resolved before invoking a function");
+                            return Err(DscError::Parser("Error in parsing".to_string()));
+                        }
+                    }
+                }
+
+
+                function.invoke(args)
+            },
+            None => Err(DscError::Parser(format!("Unknown function {0}", name))),
+        }
+    }
+}
--- a/dsc_lib/src/functions/mod.rs
+++ b/dsc_lib/src/functions/mod.rs
@@ -66,45 +66,54 @@ impl FunctionDispatcher {
     ///
     /// This function will return an error if the function fails to execute.
     pub fn invoke(&self, name: &str, args: &Vec<FunctionArg>) -> Result<FunctionResult, DscError> {
-        let function = self.functions.get(name);
-        match function {
-            Some(function) => {
-                // check if arg number are valid
-                if args.len() < function.min_args() {
-                    return Err(DscError::Parser(format!("Function '{name}' requires at least {0} arguments", function.min_args())));
-                }
-                if args.len() > function.max_args() {
-                    return Err(DscError::Parser(format!("Function '{name}' supports at most {0} arguments", function.max_args())));
-                }
-                // check if arg types are valid
-                for arg in args {
-                    match arg {
-                        FunctionArg::String(_) => {
-                            if !function.accepted_arg_types().contains(&AcceptedArgKind::String) {
-                                return Err(DscError::Parser(format!("Function '{name}' does not accept string arguments")));
-                            }
-                        },
-                        FunctionArg::Integer(_) => {
-                            if !function.accepted_arg_types().contains(&AcceptedArgKind::Integer) {
-                                return Err(DscError::Parser(format!("Function '{name}' does not accept integer arguments")));
-                            }
-                        },
-                        FunctionArg::Boolean(_) => {
-                            if !function.accepted_arg_types().contains(&AcceptedArgKind::Boolean) {
-                                return Err(DscError::Parser(format!("Function '{name}' does not accept boolean arguments")));
-                            }
-                        },
-                        FunctionArg::Expression(_) => {
-                            debug!("An expression was not resolved before invoking a function");
-                            return Err(DscError::Parser("Error in parsing".to_string()));
-                        }
-                    }
-                }
+        let Some(function) = self.functions.get(name) else {
+            return Err(DscError::Parser(format!("Unknown function '{name}'")));
+        };
 
-                function.invoke(args)
-            },
-            None => Err(DscError::Parser(format!("Unknown function '{name}'"))),
+        // check if arg number are valid
+        let min_args = function.min_args();
+        let max_args = function.max_args();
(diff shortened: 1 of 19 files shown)


dsc_lib/src/parser/functions.rs
```diff
@@ -0,0 +1,115 @@
...
+                        resolved_args.push(FunctionArg::String(value));
+                    },
+                    _ => {
+                        resolved_args.push(arg.clone());
+                    }
+                }
+            }
+        }
+
+        self.function_dispatcher.invoke(&self.name, &resolved_args)
```
daxian-dbw, line 80: same here, the `invoke` method should take a `& FunctionDispatcher` parameter, and call `invoke` on it here.

dsc_lib/src/parser/mod.rs
```diff
@@ -0,0 +1,139 @@
...
+            "escapedStringLiteral" => {
+                // need to remove the first character: [[ => [
+                let Ok(value) = child_node.utf8_text(statement.as_bytes()) else {
+                    return Err(DscError::Parser("Error parsing escaped string literal".to_string()));
+                };
+                Ok(value[1..].to_string())
+            },
+            "expression" => {
+                let expression = Expression::new(&self.function_dispatcher, statement, &child_node)?;
+                Ok(expression.invoke()?)
```
daxian-dbw, line 78: Just pass the `& FunctionDispatcher` in here.

dsc_lib/src/parser/mod.rs
```diff
@@ -0,0 +1,139 @@
...
+        let Some(child_node) = root_node.named_child(0) else {
+            return Err(DscError::Parser("Child node not found".to_string()));
+        };
+        if child_node.is_error() {
+            return Err(DscError::Parser("Error parsing statement child".to_string()));
+        }
+        let kind = child_node.kind();
+        match kind {
+            "stringLiteral" | "bracketInStringLiteral" => {
+                let Ok(value) = child_node.utf8_text(statement.as_bytes()) else {
```
daxian-dbw, line 64: `statement.as_bytes()` is called many times when traverse the tree, which is unneeded overhead. I think you should call it once in this method, and use the returned value for the rest places.

dsc_lib/src/parser/mod.rs
```diff
@@ -0,0 +1,139 @@
...
+            },
+            "escapedStringLiteral" => {
+                // need to remove the first character: [[ => [
+                let Ok(value) = child_node.utf8_text(statement.as_bytes()) else {
+                    return Err(DscError::Parser("Error parsing escaped string literal".to_string()));
+                };
+                Ok(value[1..].to_string())
+            },
+            "expression" => {
+                let expression = Expression::new(&self.function_dispatcher, statement, &child_node)?;
```
daxian-dbw, line 77: For example, here you can pass in `source &[u8]` instead of `statement` to the constructor.

registry/src/main.rs
```diff
@@ -91,7 +91,7 @@ fn main() {
         args::SubCommand::Config { subcommand } => {
             let json: String;
             match regconfighelper::validate_config(&config) {
-                Ok(_) => {},
+                Ok(()) => {},
                 Err(err) => {
                     eprintln!("Error validating config: {err}");
                     exit(EXIT_INVALID_INPUT);
```
daxian-dbw, line 97: Same here, `if Err(err) = regconfighelper::validate_config(&config) { ... }` would get us better readability.

dsc_lib/src/functions/mod.rs
```diff
@@ -0,0 +1,115 @@
...
+    /// # Errors
+    ///
+    /// This function will return an error if the function fails to execute.
+    pub fn invoke(&self, name: &str, args: &Vec<FunctionArg>) -> Result<FunctionResult, DscError> {
+        let function = self.functions.get(name);
+        match function {
+            Some(function) => {
+                // check if arg number are valid
+                if args.len() < function.min_args() {
+                    return Err(DscError::Parser(format!("Function '{name}' requires at least {0} arguments", function.min_args())));
```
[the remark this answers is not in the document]
SteveL-MSFT: The problem with that is that `concat()` takes an arbitrary number of args, so showing maxInt as a number isn't very appealing.

dsc_lib/src/parser/expressions.rs
```diff
@@ -0,0 +1,88 @@
...
+            member_access,
+        })
+    }
+
+    /// Invoke the expression.
+    ///
+    /// # Errors
+    ///
+    /// This function will return an error if the expression fails to execute.
+    pub fn invoke(&self) -> Result<String, DscError> {
```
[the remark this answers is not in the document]
SteveL-MSFT: Seems a bit cleaner to me to use the lifespan annotation

dsc_lib/src/parser/mod.rs
SteveL-MSFT: `as_bytes()` returns a slice, so seems ok here vs allocating a variable that might not get used.

tree-sitter-dscexpression/grammar.js
```diff
@@ -18,19 +18,20 @@ module.exports = grammar({
...
 
-    string: $ => seq("'", /[^']*/, "'"),
-    number: $ => /\d+/,
+    // ARM strings do not allow to contain single-quote characters
+    string: $ => /[^']*/,
+    number: $ => /-?\d+/,
     boolean: $ => choice('true', 'false'),
 
-    _members: $ => repeat1($._member),
+    memberAccess: $ => repeat1($._member),
```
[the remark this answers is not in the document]
SteveL-MSFT: Let's discuss this at the review

dsc_lib/src/parser/expressions.rs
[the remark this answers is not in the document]
daxian-dbw: lifespan annotation generally adds cognitive burden to someone reading the code.

dsc_lib/src/parser/mod.rs
daxian-dbw: It should always be used when reaching the line `match kind`.

`as_bytes` converts a string slice into a byte slice through `mem::transmute<Src, Dst>(src: Src) -> Dst`, whose documentation indicates a copy operation. Also, in general, avoiding calling a method multiple times to get the same value is a good practice to follow.

> `transmute` is semantically equivalent to a bitwise move of one type
> into another. It copies the bits from the source value into the
> destination value, then forgets the original. Note that source and destination
> are passed by-value, which means if `Src` or `Dst` contain padding, that padding
> is *not* guaranteed to be preserved by `transmute`.

dsc_lib/src/parser/expressions.rs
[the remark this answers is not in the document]
daxian-dbw: lifespan annotation generally adds cognitive burden to someone who is reading the code.

dsc_lib/src/parser/mod.rs
SteveL-MSFT: Didn't realize it did a copy.  Will change.

tree-sitter-dscexpression/grammar.js
```diff
@@ -18,19 +18,20 @@ module.exports = grammar({
...
     stringLiteral: $ => token(prec(PREC.STRINGLITERAL, /[^\[].*?/)),
 
-    function: $ => seq($.functionName, '(', optional($._arguments), ')'),
-    functionName: $ => /[a-zA-Z]+/,
-    _arguments: $ => seq($._argument, repeat(seq(',', $._argument))),
-    _argument: $ => choice($.expression, $.string, $.number, $.boolean),
+    function: $ => seq(field('name', $.functionName), '(', field('args', optional($.arguments)), ')'),
+    functionName: $ => /[a-z][a-zA-Z0-9]*/,
+    arguments: $ => seq($._argument, repeat(seq(',', $._argument))),
+    _argument: $ => choice($.expression, seq('\'', $.string, '\''), $.number, $.boolean),
```
SteveL-MSFT, line 27: _quotedString

dsc_lib/src/parser/functions.rs
```diff
@@ -0,0 +1,115 @@
...
+use crate::parser::{
+    expressions::Expression,
+    FunctionDispatcher,
+};
+
+#[derive(Clone)]
+pub struct Function<'a> {
+    name: String,
+    args: Option<Vec<FunctionArg<'a>>>,
+    function_dispatcher: &'a FunctionDispatcher,
```
SteveL-MSFT, line 17: Consider having the trait implementation directly of the dispatcher

tree-sitter-dscexpression/grammar.js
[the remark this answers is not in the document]
SteveL-MSFT: I still want the `memberAccess` parent node over the actual `memberName` nodes, so changing the grammar so `memberAccess` is not optional and the definition of `memberAccess` is to `repeat(member)` doesn't work as tree-sitter complains that this can match an empty string.

tree-sitter-dscexpression/grammar.js
SteveL-MSFT: fixed

dsc_lib/src/parser/functions.rs
SteveL-MSFT: This won't work as the dispatcher also validates inputs

dsc_lib/src/parser/functions.rs
```diff
@@ -0,0 +1,113 @@
...
+    /// # Arguments
+    ///
+    /// * `function_dispatcher` - The function dispatcher to use.
+    /// * `statement` - The statement that the function is part of.
+    /// * `function` - The function node.
+    ///
+    /// # Errors
+    ///
+    /// This function will return an error if the function node is not valid.
+    pub fn new(function_dispatcher: &FunctionDispatcher, statement_bytes: &[u8], function: &Node) -> Result<Self, DscError> {
```
daxian-dbw, line 45: `function_dispatcher` can be removed now for `new`. It's not really used, but just passed around.

dsc_lib/src/parser/functions.rs
```diff
@@ -0,0 +1,113 @@
...
+                    }
+                }
+            }
+        }
+
+        function_dispatcher.invoke(&self.name, &resolved_args)
+    }
+}
+
+fn convert_args_node(function_dispatcher: &FunctionDispatcher, statement_bytes: &[u8], args: &Option<Node>) -> Result<Option<Vec<FunctionArg>>, DscError> {
```
daxian-dbw, line 82: Same here, `function_dispatcher` can be removed.

daxian-dbw (changes requested):

dsc_lib/src/parser/expressions.rs
```diff
@@ -0,0 +1,88 @@
...
+    ///
+    /// # Errors
+    ///
+    /// This function will return an error if the expression node is not valid.
+    pub fn new(function_dispatcher: &FunctionDispatcher, statement_bytes: &[u8], expression: &Node) -> Result<Self, DscError> {
+        let Some(function) = expression.child_by_field_name("function") else {
+            return Err(DscError::Parser("Function node not found".to_string()));
+        };
+        let function = Function::new(function_dispatcher, statement_bytes, &function)?;
+        let member_access = if let Some(member_access) = expression.child_by_field_name("members") {
```
daxian-dbw, line 33: For better readability, please use different variable names here.
```suggestion
        let member_access = if let Some(members) = expression.child_by_field_name("members") {
```

daxian-dbw (approved):

PR #242, opened 2023-10-23
Add example `brew` resource

SteveL-MSFT:
# PR Summary

Demo resource for `brew` to manage packages on macOS.

Also updated `build.ps1` with a new `copy_files.txt` for resources that don't need a build step where the file indicates the files to be copied to the target folder.

A seperate PR will move the other resources under the `resources` folder.

## PR Context

This demos a bash script resource using env var arguments

### DIFF
--- a/resources/brew/brew.dsc.resource.sh
+++ b/resources/brew/brew.dsc.resource.sh
@@ -1,0 +1,47 @@
+#!/bin/sh
+
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+export exist=true
+
+check_args() {
+    if [[ -z $packageName ]]; then
+        echo "packageName not set"
+        exit 1
+    fi
+}
+
+to_json() {
+    while read line; do
+        echo $line | awk '{print "{ \"packageName\": \""$1"\", \"version\": \""$2"\", \"_exist\": "ENVIRON["exist"]" }"}'
+    done
+}
+
+if [ $# -eq 0 ]; then
+    echo "Command not provided, valid commands: get, export"
+    exit 1
+elif [[ $1 == "get" ]]; then
+    check_args
+    output="$(brew list ${packageName} --versions)"
+    if [[ $? -ne 0 ]]; then
+        export exist=false
+        output="${packageName}"
+    fi
+    echo $output | to_json
+elif [[ $1 == "set" ]]; then
+    check_args
+    if [[ -z $_exist ]]; then
+        _exist=true
+    fi
+    if [[ $_exist = true ]]; then
+        brew install "${packageName}"
+    else
+        brew uninstall "${packageName}"
+    fi
+elif [[ $1 == "export" ]]; then
+    brew list --versions | to_json
+else
+    echo "Invalid command, valid commands: get, export"
+    exit 1
+fi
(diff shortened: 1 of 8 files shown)


resources/brew/brew.dsc.resource.sh
```diff
@@ -0,0 +1,48 @@
+#!/bin/sh
+
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+export exist=true
+export NONINTERACTIVE=1
+
```
JamesWTruher, line 8: you might add a comment about `$packageName` which is expected to be exported in the environment

resources/brew/brew.dsc.resource.sh
```diff
@@ -0,0 +1,48 @@
...
+    fi
+    if [[ $_exist = true ]]; then
+        brew install "${packageName}"
+    else
+        brew uninstall "${packageName}"
+    fi
+elif [[ $1 == "export" ]]; then
+    brew list --versions | to_json
+else
+    echo "Invalid command, valid commands: get, export"
```
JamesWTruher, line 46:
```suggestion
    echo "Invalid command, valid commands: get, set, export"
```

resources/brew/brew.dsc.resource.sh
```diff
@@ -0,0 +1,48 @@
...
+}
+
+to_json() {
+    while read line; do
+        echo $line | awk '{print "{ \"packageName\": \""$1"\", \"version\": \""$2"\", \"_exist\": "ENVIRON["exist"]" }"}'
+    done
+}
+
+if [ $# -eq 0 ]; then
+    echo "Command not provided, valid commands: get, export"
```
JamesWTruher, line 23:
```suggestion
    echo "Command not provided, valid commands: get, set, export"
```

resources/brew/brew.dsc.resource.sh
```diff
@@ -0,0 +1,48 @@
...
+    exit 1
+elif [[ $1 == "get" ]]; then
+    check_args
+    output="$(brew list ${packageName} --versions)"
+    if [[ $? -ne 0 ]]; then
+        export exist=false
```
JamesWTruher, line 29: since `exist` is exported above in line 6, the `export` should not be needed here.

resources/brew/brew.dsc.resource.sh
```diff
@@ -0,0 +1,48 @@
...
+        output="${packageName}"
+    fi
+    echo $output | to_json
+elif [[ $1 == "set" ]]; then
+    check_args
+    if [[ -z $_exist ]]; then
+        _exist=true
```
JamesWTruher, line 36: how would `$_exist` have otherwise been set?

JamesWTruher (approved):
just a few questions and comments

resources/brew/brew.dsc.resource.sh
SteveL-MSFT: `dsc.exe` will set the environment to include `_exist` as an env var before this script is run.  That's how it passes the "JSON" payload as env var name-value pairs so there's no JSON parsing within the script.  Line 35-36 is so that `_exist` default to `true` if it's not specified in the original JSON payload.

resources/brew/brew.dsc.resource.sh
SteveL-MSFT: good point, adding

PR #250, opened 2023-11-01
Added debug trace with DSC version

anmenaga:
# PR Summary

2 updates with this PR:
1) Changed Clap::Args::version (used in `dsc -V`) to use value from `dsc\Cargo.toml`. This makes one less place to update when dsc version is incremented after a release.
2) Added debug statement with the current dsc version; e.g.:

```
PS C:\DSCv3> dsc -l trace resource get -r Microsoft/OSInfo
  2023-11-01T22:59:24.235180Z DEBUG dsc: Running dsc 3.0.0-alpha.4
    at src\main.rs:48
```

### DIFF
--- a/dsc/src/args.rs
+++ b/dsc/src/args.rs
@@ -13,7 +13,7 @@ pub enum OutputFormat {
 }
 
 #[derive(Debug, Parser)]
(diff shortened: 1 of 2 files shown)


SteveL-MSFT (approved):
For debug it may make sense to include basic OS info: windows, linux, mac version and architecture, but that can be a different PR.  This way when people open bugs, we can just have them copy the output to the issue.

Issue #57, opened 2023-04-05
Implement expression functions


SteveL-MSFT:
### Summary of the new feature / enhancement

ARM has a specific set of functions that can be used within a configuration: https://learn.microsoft.com/en-us/azure/azure-resource-manager/templates/template-functions

#### Array

- [x] array
- [x] concat
- [x] contains
- [x] createArray
- [x] empty
- [x] first
- [x] indexOf
- [x] intersection
- [x] last
- [x] lastIndexOf
- [x] length
- [x] min
- [x] max
- [x] range
- [x] skip
- [x] take
- [x] union

#### CIDR

These aren't relevant to DSC

#### Comparison

- [x] coalesce
- [x] equals
- [x] less
- [x] lessOrEquals
- [x] greater
- [x] greaterOrEquals

#### Date

- [ ] dateTimeAdd
- [ ] dateTimeFromEpoch
- [ ] dateTimeToEpoch
- [x] utcNow

#### Deployment

- [ ] deployment?
- [ ] environment*
- [x] parameters*
- [x] variables*

#### Logical

- [x] and
- [x] bool
- [x] false
- [x] if
- [x] not
- [x] or
- [x] true

#### Numeric

- [x] add
- [x] copyIndex
- [x] div
- [ ] float
- [x] int
- [x] min
- [x] max
- [x] mod
- [x] mul
- [x] sub

#### Object

- [x] contains
- [x] createObject
- [x] empty
- [x] intersection
- [x] items
- [x] json
- [x] length
- [x] null
- [x] union

#### Resource

Subset needed for DSC

- [x] reference*
- [ ] references
- [x] resourceId

#### Scope

Not applicable to DSC

#### String

- [x] base64
- [ ] base64ToJson
- [x] base64ToString
- [x] concat
- [x] contains
- [ ] dataUri
- [ ] dataUriToString
- [x] empty
- [x] endsWith
- [x] first
- [x] format
- [ ] guid
- [x] indexOf
- [x] join
- [x] last
- [x] lastIndexOf
- [x] length
- [ ] newGuid
- [ ] padLeft
- [ ] replace
- [x] skip
- [ ] split
- [x] startsWith
- [x] string
- [x] substring
- [x] take
- [x] toLower
- [x] toUpper
- [x] trim
- [x] uniqueString
- [x] uri
- [x] uriComponent
- [x] uriComponentToString

anmenaga:
[...]

PR #252, opened 2023-11-02
Integrate expressions/functions with DSC

SteveL-MSFT:
# PR Summary

- Removed `resourceId()` regex and replaced with a function
- Add execution of expressions to get/set/test resource properties
- This integration allows use of `concat()` and `base64()` functions within a configuration
- Change from HashMap to serde Map for resource properties so don't need to transform from one type to the other due to nested properties
- Expressions are evaluated only for resource properties at invocation time
- NOTE: when converting Value to string, you have to use `as_str()` instead of `to_string()` as the latter includes outer quotes

## PR Context

Partially implements https://github.com/PowerShell/DSC/issues/57


dsc_lib/src/functions/mod.rs
```diff
@@ -50,6 +51,7 @@ impl FunctionDispatcher {
         let mut functions: HashMap<String, Box<dyn Function>> = HashMap::new();
```
anmenaga, line 51: Can we have one end-to-end Pester test for each of these functions?
That would be also a good example demonstrating their use.

anmenaga (approved):

dsc_lib/src/configure/config_doc.rs
```diff
@@ -66,16 +66,14 @@ pub struct Resource {
...
     #[serde(rename = "type")]
     pub resource_type: String,
-    // TODO: `apiVersion` is required by ARM but doesn't make sense here
-
     /// A friendly name for the resource instance
     pub name: String, // friendly unique instance name
     #[serde(rename = "dependsOn", skip_serializing_if = "Option::is_none")]
     #[schemars(regex(pattern = r"^\[resourceId\(\s*'[a-zA-Z0-9\.]+/[a-zA-Z0-9]+'\s*,\s*'[a-zA-Z0-9 ]+'\s*\)]$"))]
     pub depends_on: Option<Vec<String>>,
     // `identity` can be used for run-as
```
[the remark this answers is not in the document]
SteveL-MSFT: Comment is correct, but we've made the statement to be a subset of ARM, so it no longer is needed

dsc_lib/src/functions/mod.rs
SteveL-MSFT: We need what-if support as a way to execute the functions and return the results.  I could stick in `concat()` somewhere right now although it would be artificial.

SteveL-MSFT:
The test failure is due to not handling nested types in expression invocation so fixing that.

SteveL-MSFT:
The remaining failure is due to properties that get returned may look like an expression so if you pipe that back to dsc, it fails.  I think the right thing to do here is if it's retrieved via config, then it needs to escape it if it starts and ends with brackets.

PR #248, opened 2023-11-01
Added --logging-level arg

anmenaga:
# PR Summary

This PR adds `--logging-level` / `-l` arg that specifies the level of tracing messages printed to stderr.
```
   Error,
   Warning,
   Info,
   Debug,
   Trace
```

A value from this list will also enable all above levels; for example "--logging-level Debug" will enable `Error`, `Warning`, `Info` and `Debug`.

Help Example:
```
PS C:\DSCv3> dsc --help                                   
...
Options:
  ...
  -l, --logging-level <LOGGING_LEVEL>  Log level to display [default: Info] [possible values: error, warning, info, debug, trace]
...
```

Usage example:
```
PS C:\DSCv3> dsc -l debug resource get -r Microsoft/OSInfo
  2023-11-01T17:18:53.142033Z DEBUG dsc_lib::discovery::command_discovery: Found Microsoft/OSInfo in C:\DSCv3\bin\debug\osinfo.dsc.resource.json
    at C:\DSCv3\dsc_lib\src\discovery\command_discovery.rs:86

  2023-11-01T17:18:53.142987Z DEBUG dsc::resource_command: resource.type_name - Microsoft/OSInfo implemented_as - Command
    at src\resource_command.rs:27

  2023-11-01T17:18:53.143721Z  INFO dsc_lib::dscresources::command_resource: Invoking get Microsoft/OSInfo using osinfo
    at C:\DSCv3\dsc_lib\src\dscresources\command_resource.rs:50

  2023-11-01T17:18:53.144443Z DEBUG dsc_lib::dscresources::command_resource: Invoking command osinfo with args None
    at C:\DSCv3\dsc_lib\src\dscresources\command_resource.rs:412

actualState:
  $id: https://developer.microsoft.com/json-schemas/dsc/os_info/20230303/Microsoft.Dsc.OS_Info.schema.json
  family: Windows
  version: 10.0.22631
  edition: Windows 11 Enterprise
  bitness: '64'
```

PR #262, opened 2023-11-08
(DOCS) Document `--logging-level` in CLI reference

michaeltlombardi:
# PR Summary

This change documents the new `-l` / `--logging-level` option added to the `dsc` command in #248 and updates the changelog.

## PR Context

N/A

### DIFF
--- a/docs/reference/cli/dsc.md
+++ b/docs/reference/cli/dsc.md
@@ -108,6 +108,31 @@ Type:      String
 Mandatory: false
 ```
 
+### -l, --logging-level
+
+Defines the minimum log level DSC should emit during an operation. Messages in DSC are categorized
+by their log level.
+
+The following list shows the valid log levels from highest to lowest level. When this option is
+set to any value in the list, DSC emits messages at that level and above.
+
+- `Error`
+- `Warning`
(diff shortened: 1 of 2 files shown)


docs/reference/cli/dsc.md
```diff
@@ -108,6 +108,31 @@ Type:      String
...
+
+Defines the minimum log level DSC should emit during an operation. Messages in DSC are categorized
+by their log level.
+
+The following list shows the valid log levels from highest to lowest level. When this option is
+set to any value in the list, DSC emits messages at that level and above.
+
+- `Error`
+- `Warning`
+- `Info`
```
SteveL-MSFT, line 121: Current default, if not specified, is `Info`

SteveL-MSFT (changes requested):

docs/reference/cli/dsc.md
michaeltlombardi:
```suggestion
- `Info` (default)
```

We indicate that in the metadata block below, but good to put it here, too.

SteveL-MSFT (approved):

PR #269, opened 2023-11-10
Static linking the MSVC runtime

### LINKED ISSUE #256: DSC executable doesn't run on Windows Server unless C++ redistributables are installed
### Prerequisites

### Steps to reproduce

When running the latest Alpha release on windows server 2019 and windows server 2022 the command fails with error: The code execution cannot proceed because vcruntime140.dll was not found.  Reinstalling the program may fix this problem.  Installing the latest C++ redistributable corrects this issue.  (https://learn.microsoft.com/en-us/cpp/windows/latest-supported-vc-redist?view=msvc-170)

### Expected behavior

```console
dsc resource list returns resources and not an error.
```

### Actual behavior

```console
error: The code execution cannot proceed because vcruntime140.dll was not found.  Reinstalling the program may fix this problem
```

### Error details

```console
error: The code execution cannot proceed because vcruntime140.dll was not found.  Reinstalling the program may fix this problem
```

### Environment data

```PowerShell
PSVersion                      7.3.9
GitCommitId                    7.3.9
OS                             Microsoft Windows 10.0.17763
```

### Version

7.3.9

### Visuals

image

anmenaga:
# PR Summary

Fix #256 

This PR forces static linking of MSVC runtime to avoid errors like `vcruntime140.dll was not found` on clean Windows machines.
[This uses `rustflags/+crt-static`](https://rust-lang.github.io/rfcs/1721-crt-static.html).

Side note: beware of [a bug in the area of static linking](https://github.com/rust-lang/rust/issues/71651) (looks like DSC project is not impacted by this at this point).

Fix was verified on a clean Windows Server 2022.

### DIFF
--- a/dsc/.cargo/config.toml
+++ b/dsc/.cargo/config.toml
@@ -1,0 +1,2 @@
+[target.x86_64-pc-windows-msvc]
+rustflags = ["-C", "target-feature=+crt-static"]
--- a/dsc/.cargo/config.toml
+++ b/dsc/.cargo/config.toml
@@ -1,2 +1,4 @@
 [target.x86_64-pc-windows-msvc]
 rustflags = ["-C", "target-feature=+crt-static"]
+[target.aarch64-pc-windows-msvc]
+rustflags = ["-C", "target-feature=+crt-static"]
--- a/dsc_lib/.cargo/config.toml
+++ b/dsc_lib/.cargo/config.toml
@@ -1,0 +1,2 @@
+[target.x86_64-pc-windows-msvc]
+rustflags = ["-C", "target-feature=+crt-static"]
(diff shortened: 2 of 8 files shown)


dsc/.cargo/config.toml
```diff
@@ -0,0 +1,2 @@
+[target.x86_64-pc-windows-msvc]
```
SteveL-MSFT, line 1: need to include arm64

SteveL-MSFT (changes requested):

dsc/.cargo/config.toml
anmenaga: Updated.

SteveL-MSFT (approved):

PR #277, opened 2023-11-15
(DOCS) Add reference for config functions

michaeltlombardi:
# PR Summary

This change adds the reference documentation for the newly implemented configuration document functions and updates the existing documentation as needed. It also updates the changelog to note the new functionality.

## PR Context

N/A

### DIFF
--- a/docs/reference/schemas/config/functions/concat.md
+++ b/docs/reference/schemas/config/functions/concat.md
@@ -1,0 +1,82 @@
+---
+description: Reference for the 'concat' DSC configuration document function
+ms.date:     11/15/2023
+ms.topic:    reference
+title:       concat
+---
+
+# concat
+
+## Synopsis
+
+Returns a string of combined values.
+
+## Syntax
+
+```Syntax
+concat(<inputValue>, <inputValue>[, <inputValue>...])
+```
+
+## Description
+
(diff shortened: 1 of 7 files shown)


docs/reference/schemas/config/functions/concat.md
```diff
@@ -0,0 +1,82 @@
...
+
+```Syntax
+concat(<inputValue>, <inputValue>[, <inputValue>...])
+```
+
+## Description
+
+The `concat()` function combines multiple values and returns the concatenated values as a single
+string. Separate each value with a comma. The `concat()` function is variadic. You must pass at
+least two values to the function. The function can accept up to `18446744073709551615` arguments.
```
SteveL-MSFT, line 24: Does it make sense to show such a large number?  Seems like it could be expressed differently?  See https://learn.microsoft.com/en-us/azure/azure-resource-manager/templates/template-functions-array#concat

SteveL-MSFT (changes requested):

docs/reference/schemas/config/functions/concat.md
michaeltlombardi, line 24:
```suggestion
least two values to the function. The function can accept any number of arguments.
```

SteveL-MSFT (approved):

Issue #183, opened 2023-09-01
Export support for provider-based resources 


anmenaga:
### Prerequisites

- [X] Write a descriptive title.
- [X] Make sure you are able to repro it on the latest version
- [X] Search the existing issues.

### Steps to reproduce

Need a way to support Export operation for provider-based resources.

### Expected behavior

```console
NA
```

### Actual behavior

```console
NA
```

### Environment data

```PowerShell
NA
```

### Version

v1

SteveL-MSFT:
Currently, a group resource can indicate it supports `export` within its resource manifest, but we haven't defined what the output would look like so it can be passed back to `set`.

Issue #274, opened 2023-11-14
Need a way to selectively disable individual resources/providers


anmenaga:
### Prerequisites

- [X] Write a descriptive title.
- [X] Make sure you are able to repro it on the latest version
- [X] Search the existing issues.

### Steps to reproduce

An example that I've hit with WMIGroup resource provider:
On Win11 it returns ~1300 resource types, making "dsc resource list" output useless if user is interested in other resource types.

Need a way to disable/enable specific resource types/providers (this will help with performance too).
Maybe have a config file, that can be edited by the user, that lists disabled resources.

### Expected behavior

```console
N/A
```

### Actual behavior

```console
N/A
```

### Environment data

```PowerShell
N/A
```

### Version

3.0.0-alpha.4

michaeltlombardi:
This raises a question about being able to use a configuration file to control how `dsc` behaves - continually defining options at the prompt will get more and more user-unfriendly as the amount of options increases.

SteveL-MSFT:
My immediate thought is to have a `--recurse` type switch to enumerate provider resources (although we'd call it something more user friendly like `--include-provider-resources`).  We might even consider if used interactive to prompt before dumping out 1300 WMI resources.

michaeltlombardi:
I'm not _opposed_ to a flag for listing resources from providers as an opt-in flag, but it should probably be an array option, so I could specify _which_ provider resources I want enumerated.

If we consider a hypothetical where I have PSDSC, WMI, and Python providers, I might only want to list the resources for PSDSC or Python.

If we had a `--provider` option with a builtin `all` value (enumerate _everything_), that might be more wieldy.

SteveL-MSFT:
Let's have the config discussion in https://github.com/PowerShell/DSC/issues/282 and keep this the interactive cmdline discussion.

SteveL-MSFT:
WG discussed:

- first, we need https://github.com/PowerShell/DSC/issues/139 implemented to expose `kind`
- show `kind` as a new column after `version` and before `methods`
- by default, dsc does not perform a `list` against adapters, but shows the adapters
- a new `--adapter` parameter taking the name or wildcard and for `*` it shows everything including non-adapters or adapted resources otherwise just resources that match the adapter name
- `dsc resource list *Network* --adapter *` would search for any resources or adapted resource where the name includes `Network`, for example

Issue #243, opened 2023-10-24
Group and Provider naming conventions


michaeltlombardi:
### Summary of the new feature / enhancement

As a user, I want to be able to tell at a glance whether an instance of a resource in a configuration manages a single instance, a group of instances, or is a provider for other instances.

Right now, there is an informal convention of appending `Group` to the type name, but this is used for both resource groups and resource providers.

We should have some standard guidance and follow it for the first-party resources.

> [!NOTE]
> This may not be as necessary if #139 is implemented, as we could surface this information more easily in the editor extension and in the CLI output. I think the naming convention would still have some value, but less.

### Proposed technical implementation details (optional)

The current documentation for [fully qualified type names][01] says:

> The name segment of the type name is mandatory. It identifies the component that the resource manages. This segment should be a singular noun unless the resource always manages a list of components in a single resource instance. In that case, the resource name should be the plural form of the noun it manages or the singular form of the noun followed by the word List, like `JeaRoleCapabilities` or `JeaRoleCapabilityList`.

We could add additional guidance here for resource groups and providers, like:

> If the resource is a resource group, add the `Group` suffix to the name. If it's a resource provider, add the `Provider` suffix to the name.

[01]: https://learn.microsoft.com/powershell/dsc/reference/schemas/definitions/resourcetype?view=dsc-3.0

michaeltlombardi:
Another alternative naming convention would be to have the last segment of the type name before the `/` be `Group` for group resources and `Provider` for provider resources.

Then the naming syntax would be:

```Syntax
<owner>[.<group>][.<area>][.(Group|Provider)]/<name>
```

Either convention could be extended to help indicate other resource types, like assertion resources.

SteveL-MSFT:
We may want to consider different naming conventions for DSC groups resources that are part of dsc.exe (basically built-in) vs DSC resources that may ship with DSC (PowerShellGroup, WMIGroup, for example).  Seems like built-in should use `DSC/<resourceName`, but the others we may want as `Microsoft/WindowsPowerShellGroup`, `Microsoft/PowerShell7Group`, and `Microsoft.Windows/WMIGroup`.

SteveL-MSFT:
WG discussed and closing this in favor of Kind and new naming conventions discussed.

SteveL-MSFT:
This was fixed as part of https://github.com/PowerShell/DSC/pull/341/files

PR #279, opened 2023-11-16
WMI resource provider

### LINKED ISSUE #263: WMIGroupResource

Build a WMIGroupResource that allows using WMI classes as DSC resources.  This will automatically add lots of Windows inbox resources.  Probably focus on audit scenarios first.

Some technical challenges:

- For audit (get), maybe only support WQL queries
- For set may need a way to execute a WMI method and need to consider both instance and static methods.  Also, CIM has an explicit Delete operation, so just map that from `_exists = false`?

anmenaga:
# PR Summary

Fix #263 

Prototype of a resource provider to read data from WMI.

## PR Context

Supported operations:  `dsc resource list`, `dsc resource get`, `dsc config get`.   `Export` will be easy to add once #183 is closed.
This resource provider uses `Windows PowerShell` and `CimCmdlets` PS module. Initially tried using [WMI-rs](https://docs.rs/wmi/latest/wmi/#wmi-rs) crate, but it didn't work out because looks like it does not return some required class metadata.
 
Example:
```
PS C:\DSCv3> dsc resource list *BIOS
Type                         Version  Requires      Description
---------------------------------------------------------------
root\cimv2/Win32_BIOS                 DSC/WMIGroup
root\cimv2/Win32_SystemBIOS           DSC/WMIGroup

PS C:\DSCv3> dsc resource get -r root\cimv2/Win32_BIOS
actualState:
  EmbeddedControllerMinorVersion: 255
  BuildNumber: null
  Description: 1.22.0
  ReleaseDate: /Date(1678924800000)/
  Manufacturer: Dell Inc.
  SoftwareElementID: 1.22.0
...
```


wmigroup/copy_files.txt
```diff
@@ -0,0 +1 @@
+wmigroup.resource.ps1
```
anmenaga, line 1: I'm not sure about auto-enabling this WMI provider in default DSCv3 deployment.
Discovery in this provider takes ~5-10 seconds and returns ~1300 resource types. This is a significant perf hit especially when a user is not interested in WMI resources. Related issue #274 .

wmigroup/Tests/test_wmi_config.dsc.yaml
```diff
@@ -0,0 +1,13 @@
...
+- name: Get info from WMI
+  type: DSC/WMIGroup
+  properties:
+    resources:
+    - name: Get OS Info
+      type: root\cimv2/Win32_OperatingSystem
+    - name: Get OS Info
+      type: root\cimv2/Win32_BIOS
+    - name: Get OS Info
+      type: root\cimv2/Win32_Processor
```
michaeltlombardi, line 13: Technically, these type names violate our permitted pattern for [fully qualified type names][01]:

```regex
^\w+(\.\w+){0,2}\/\w+$
```
```Syntax
<owner>[.<group>][.<area>]/<name>
```

I wonder if this could be `root.cimv2/Win32_OperatingSystem` or `cimv2/Win32_OperatingSystem`.

[01]: https://learn.microsoft.com/powershell/dsc/reference/schemas/definitions/resourcetype?view=dsc-3.0&preserveView=true

wmigroup/copy_files.txt
michaeltlombardi: This one, at least for now, we might want to document opting-in until #274 is handled. Otherwise it'll give the impression that DSC is _very_ slow and require users not interested in WMI during the alpha to opt out. I also expect that to impact test runs.

wmigroup/Tests/test_wmi_config.dsc.yaml
```diff
@@ -0,0 +1,13 @@
+# Example configuration for reading data from Windows WMI
+$schema: https://raw.githubusercontent.com/PowerShell/DSC/main/schemas/2023/10/config/document.json
+resources:
+- name: Get info from WMI
+  type: DSC/WMIGroup
```
michaeltlombardi, line 5: I mentioned this problem in #243, but it's probably worth considering whether we want to name this `DSC.Provider/WMI` or `DSC/WMIProvider` instead, given how differently providers and groups are from a user perspective.

wmigroup/wmigroup.resource.ps1
```diff
@@ -0,0 +1,129 @@
...
+        $propertyList = @()
+        foreach ($p in $r.CimClassProperties)
+        {
+            if ($p.Name)
+            {
+                $propertyList += $p.Name
+            }
+        }
+
+        $namespace = $r.CimSystemProperties.Namespace.ToLower().Replace('/','\')
```
michaeltlombardi, line 35: If we _do_ want to make the type name compliant with the current pattern for DSC Resource type names, I think we'd have to use `.` instead of `\`.

wmigroup/copy_files.txt
anmenaga: For now, I'll rename `wmigroup.dsc.resource.json` to `wmigroup.dsc.resource.json.optout` so that the `WMIGroup` resource is still copied to the release folder but will be skipped by default.

wmigroup/Tests/test_wmi_config.dsc.yaml
anmenaga: We'll rename (we'll need to) this after we reach conclusion on #243.

wmigroup/Tests/test_wmi_config.dsc.yaml
anmenaga: Replaced with dots.

wmigroup/wmigroup.resource.ps1
anmenaga: I agree. Updated PR to use `.` instead of `\`.

SteveL-MSFT (approved):
Looks fine for a prototype until we get more feedback

Issue #46, opened 2023-03-29
Enable resources to participate in progress reporting during config execution


SteveL-MSFT:
### Summary of the new feature / enhancement

As resources get executed, need progress returned as JSON for agents to display.  Consider how child processes can participate including resources.

anmenaga:
To put more context to this.
The main work item to enable progress reporting (as well as streaming results from a resource in real-time without waiting until it completes the operation) is actually to switch resource execution from synchronous to async.
I did initial implementation of it using tokio runtime.

ThomasNieto:
Can we have the resources output a JSON object to STDERR like we do with the other logging levels? Maybe just reuse the progress output that DSC is emitting currently if set to JSON.

```
{"id":"0776289b-2896-491a-b87e-d54fac240dbb","activity":"Searching for extensions","totalItems":1,"completedItems":0}
```

michaeltlombardi:
While not directly related, we will need to be sure that migrating to Open Telemetry includes handling for progress events and clearly define how resources (and extensions/adapters/etc) can and should participate for improved UX.

PR #287, opened 2023-12-01
Support for resource tracing info

### LINKED ISSUE #89: Enable way for resources and dsc.exe to provide verbose/debug information

Do we need both verbose and debug?  Maybe start with just verbose.
Have a well defined JSONLINE schema for resources to output verbose information.  Do we need to a way to turn on verbose for commands or just have it write it all the time and we throw it away if the user hasn't specified `dsc --verbose`?
What schema makes sense?  Perhaps:

```yaml
resourceType: <type name of the resource>
name: <friendly name if there are multiple instances of resources in config>
message: <open text>
timestamp: <filled in by dsc.exe>
```

It probably makes sense that resources only return the verbose message itself and dsc.exe fills in the rest.  So an example JSON from a resource going to stderr could look like:

```json
{
  "verbose": "this is my verbose message"
}
```

anmenaga:
# PR Summary

Fix #89 

This is a basic support for resources and resource providers to push their diagnostic info to the log of `dsc.exe`.

Contract is:
If a resource/provider wants to submit a diagnostic message to the DSC, it has to format the message as JSON string with `"Error"/"Warning"/"Info"/"Debug"/"Trace"` field and publish it to the STDERR stream.
For example:
`{"Trace":"PSPath=c:\PowerShell-7.3.1-win-x64"}`

Strings in the STDERR that are not a valid JSON are ignored (unless exit code is non-zero);

Note:
Timestamps currently show the time when `dsc` receives all output when resource process completes. Timestamps will become accurate (i.e. will show exact time when they are generated by a resource/provider) once #46 is done.

Example:
q2


dsc_lib/src/dscresources/command_resource.rs
```diff
@@ -6,10 +6,30 @@ use serde_json::Value;
 use std::{collections::HashMap, process::Command, io::{Write, Read}, process::Stdio};
 use crate::dscerror::DscError;
 use super::{dscresource::get_diff,resource_manifest::{ResourceManifest, InputKind, ReturnKind, SchemaKind}, invoke_result::{GetResult, SetResult, TestResult, ValidateResult, ExportResult}};
-use tracing::{debug, info};
+use tracing::{debug, info, trace};
 
 pub const EXIT_PROCESS_TERMINATED: i32 = 0x102;
 
+
+pub fn print_resource_traces(stderr: &String)
```
SteveL-MSFT, line 14: `print` implies direct output to console, perhaps it should be `log_resource_traces()`

dsc_lib/src/dscresources/command_resource.rs
```diff
@@ -6,10 +6,30 @@ use serde_json::Value;
...
+{
+    if !stderr.is_empty()
+    {
+        for trace_line in stderr.lines() {
+            match serde_json::from_str::<Value>(trace_line){
```
SteveL-MSFT, line 19: I think we should have structs defined for trace, debug, info and try to deserialize the output and do the matching logging when it deserializes otherwise we can debug an unknown trace with the unexpected output.

SteveL-MSFT (changes requested):

dsc_lib/src/dscresources/command_resource.rs
anmenaga: Updated.

dsc_lib/src/dscresources/command_resource.rs
SteveL-MSFT: As discussed, we can do this later and probably align with the tracing crate JSON format to make it easier for ourselves

SteveL-MSFT (approved):

dsc_lib/src/dscresources/command_resource.rs
anmenaga: Attempting several deserializations on the same string is a big perf hit.
It is better to deserialize once and then check if the first member is Error/Warning/Debug/etc... and log accordingly.

anmenaga:
Updated PR Summary to be accurate to latest changes.

Issue #49, opened 2023-03-30
Implement `parameters` support


SteveL-MSFT:
### Summary of the new feature / enhancement

Parameters will eventually become the way to have templates

PR #291, opened 2024-01-05
Add parameters support

SteveL-MSFT:
# PR Summary

Added:

- `dsc` CLI can accept parameters JSON/YAML as string or file
- parameters are validated against constraints (some TODOs in the code to fill in later)
- context added to configurator keeping necessary state for parameters but also future use for variables/references

The osinfo example YAMLs currently do not work due to schema validation happening before expression execution

## PR Context

Fix https://github.com/PowerShell/DSC/issues/49

### DIFF
--- a/dsc_lib/src/configure/context.rs
+++ b/dsc_lib/src/configure/context.rs
@@ -11,11 +11,18 @@ pub struct Context {
 }
 
 impl Context {
+    #[must_use]
     pub fn new() -> Self {
         Self {
             parameters: HashMap::new(),
             _variables: HashMap::new(),
             _outputs: HashMap::new(),
         }
+    }
+}
+
+impl Default for Context {
+    fn default() -> Self {
+        Self::new()
     }
 }
--- a/dsc_lib/src/configure/context.rs
+++ b/dsc_lib/src/configure/context.rs
@@ -1,0 +1,21 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+use serde_json::Value;
+use std::collections::HashMap;
+
+pub struct Context {
+    pub parameters: HashMap<String, Value>,
+    pub _variables: HashMap<String, Value>,
+    pub _outputs: HashMap<String, Value>, // This is eventually for References function to get output from resources
+}
+
+impl Context {
+    pub fn new() -> Self {
+        Self {
+            parameters: HashMap::new(),
+            _variables: HashMap::new(),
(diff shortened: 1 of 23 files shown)


dsc_lib/src/configure/context.rs
```diff
@@ -0,0 +1,28 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+use serde_json::Value;
+use std::collections::HashMap;
+
+pub struct Context {
```
anmenaga, line 7: If params can potentially be used in direct resource invocations (in addition to configurations) then this file probably should be moved to a more shared place.

process/Cargo.toml
```diff
@@ -4,6 +4,6 @@ version = "0.1.0"
 edition = "2021"
 
 [dependencies]
-sysinfo = "*"
+sysinfo = "0.30"
```
anmenaga, line 7: 1) Are this changes in "process" related to this PR/parameters?
2) why the "sysinfo" package version change?

dsc_lib/src/configure/contraints.rs
```diff
@@ -0,0 +1,170 @@
...
+///
+/// # Returns
+///
+/// * `Ok(())` if the value matches the constraints.
+/// * `Err(DscError::Validation)` if the value does not match the constraints.
+///
+/// # Errors
+///
+/// * `DscError::Validation` if the value does not match the constraints.
+pub fn check_number(name: &str, value: &Value, constraint: &Parameter) -> Result<(), DscError> {
```
anmenaga, line 92: I would rename this to be more specific like something like "check_number_limits" or "check_max_min" .

anmenaga (approved):

dsc_lib/src/configure/context.rs
SteveL-MSFT: There's no intent to have params supported for direct resource invocation, it's a feature of config only at this time

process/Cargo.toml
SteveL-MSFT: Yes, because the `*`, CI picked up a different older version than what I was building on my local system so some of the syntax changed between versions.  Having explicit version keeps it consistent so it builds both locally and in CI.

dsc_lib/src/configure/contraints.rs
SteveL-MSFT: agree on naming, will make this change.

PR #294, opened 2024-01-16
Fix parameters to work with property values

SteveL-MSFT:
# PR Summary

Before this PR, resources were being statically validated against their schema.  This fails because a property value may use an expression and that string may fail schema validation as the expression hasn't been invoked yet.  This change moves resource input validation later when the resource is actually getting invoked and expressions have been invoked.

There was also an issue where string parameters were being returned enclosed in quotes, this PR also fixes that issue.

Also added mock `validate` implementation for PowerShellGroup and WmiGroup resources rather than have code in DSC to skip validation.


anmenaga (approved):
LGTM

PR #298, opened 2024-01-26
fix version check in psgroup resource

SteveL-MSFT:
# PR Summary

The version check was guarding against 7.4-preview, but didn't account for 7.5+.  Fix is to change minor version from `>` to `=`

### DIFF
--- a/powershellgroup/powershellgroup.resource.ps1
+++ b/powershellgroup/powershellgroup.resource.ps1
@@ -21,7 +21,7 @@ function RefreshCache
     $script:ResourceCache = @{}
 
     $DscResources = Get-DscResource
-    
+
     foreach ($r in $DscResources)
     {
         $moduleName = "";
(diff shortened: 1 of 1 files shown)


powershellgroup/powershellgroup.resource.ps1
```diff
@@ -33,7 +33,7 @@ function RefreshCache
     }
 }
 
-if (($PSVersionTable.PSVersion.Major -ge 7) -and ($PSVersionTable.PSVersion.Minor -ge 4) `
+if (($PSVersionTable.PSVersion.Major -ge 7) -and ($PSVersionTable.PSVersion.Minor -eq 4) `
```
anmenaga, line 36: Looks like "Major -ge 7)" should also be "-eq",
otherwise this will be triggered for 8.4-previews as well... and 9.4-previews... etc...

anmenaga (changes requested):

powershellgroup/powershellgroup.resource.ps1
SteveL-MSFT: true

SteveL-MSFT, line 36:
```suggestion
if (($PSVersionTable.PSVersion.Major -eq 7) -and ($PSVersionTable.PSVersion.Minor -eq 4) `
```

anmenaga (approved):

Issue #227, opened 2023-10-10
Add `--format` to subcommands


michaeltlombardi:
### Summary of the new feature / enhancement

Currently, to define a format for the output, a user needs to add the `--format` / `-f` option after the root command and before any subcommands, like:

```powershell
dsc --format pretty-json resource test -r Microsoft/OSInfo -i '{ "bitness": "64" }'
```

We should make it possible for users to define the format option on the subcommands as well or instead, so they follow a semantic of selecting the command to use, then supplying options for it:

```powershell
dsc resource test -r Microsoft/OSInfo -i '{ "bitness": "64" }' -f pretty-json
```

> [!NOTE]
> Additionally, `--format` doesn't apply to the `dsc completer` command.

### Proposed technical implementation details (optional)

1. Define `--format` as a reusable option and add it to the final-leaf subcommands where it applies. For example, the option should be added to `dsc config set` but not `dsc config`.
1. (Optional) Remove the global option, or make `dsc completer` error when the global option is specified.

SteveL-MSFT:
Discussed in WG and agreed on this change

PR #303, opened 2024-01-31
Move format parameter to relevant subcommands

tgauth:
# PR Summary
- address #227 

## PR Context
- moves `--format` parameter from a global command to relevant subcommands

### DIFF
--- a/dsc/src/args.rs
+++ b/dsc/src/args.rs
@@ -18,9 +18,6 @@ pub struct Args {
     /// The subcommand to run
     #[clap(subcommand)]
     pub subcommand: SubCommand,
-    /// The output format to use
-    #[clap(short = 'f', long)]
-    pub format: Option<OutputFormat>,
     #[clap(short = 'i', long, help = "The input to pass to the configuration or resource", conflicts_with = "input_file")]
     pub input: Option<String>,
     #[clap(short = 'p', long, help = "The path to a file used as input to the configuration or resource")]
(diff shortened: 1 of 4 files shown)


dsc/src/args.rs
```diff
@@ -54,21 +52,35 @@ pub enum SubCommand {
...
+        format: Option<OutputFormat>,
     },
 }
 
 #[derive(Debug, PartialEq, Eq, Subcommand)]
 pub enum ConfigSubCommand {
     #[clap(name = "get", about = "Retrieve the current configuration")]
-    Get,
+    Get {
+        #[clap(short = 'f', long)]
```
SteveL-MSFT, line 64: Should have help or `///` comment so description show up in help

dsc/tests/dsc_args.tests.ps1
```diff
@@ -85,7 +85,7 @@ actualState:
     ) {
         param($format, $expected)
 
-        $out = dsc --format $format resource get -r Test/Hello | Out-String
+        $out = dsc resource get -r Test/Hello --format $format | Out-String
```
SteveL-MSFT, line 88: We should probably add some more tests since this parameter is now with every sub-command vs being global

SteveL-MSFT (changes requested):

dsc/tests/dsc_set.tests.ps1
```diff
@@ -67,6 +67,49 @@ Describe 'config set tests' {
         ($result.psobject.properties | Measure-Object).Count | Should -Be 3
     }
 
+    It 'can accept the use of --format <format> as a subcommand' -Skip:(!$IsWindows) -TestCases @(
+        @{ format = 'yaml'; expected = @'
+beforeState:
```
SteveL-MSFT, line 72: For tests, we should stop using the registry resource since it only works on Windows.  You should use the [`Test/Echo`](https://github.com/PowerShell/DSC/blob/main/tools/dsctest/dscecho.dsc.resource.json) resource instead.

SteveL-MSFT (changes requested):

dsc/tests/dsc_set.tests.ps1
tgauth: had trouble using Test/Echo with set and "text: hello" as input so ended up using the resource from the set tests that works on all platforms

SteveL-MSFT (approved):

Issue #226, opened 2023-10-10
Refactor `--input` and `--input-file` options


michaeltlombardi:
### Summary of the new feature / enhancement

The `--input` and `--input-file` options were added as global options in #217, but these options aren't actually global - they don't apply to the `dsc schema`, `dsc completer`, `dsc resource schema`, `dsc resource list`, or `dsc resource export` commands.

Instead, the CLI should be updated so that these options are local to the commands they apply to. This will also aid in the CLI ergonomics, so instead of these commands:

```powershell
dsc --input-file ./example.dsc.config.yaml config set
dsc --p ./example.dsc.config.yaml config set

dsc --input $cfg config set
dsc --i $cfg config set

dsc --input $desired resource set --resource Microsoft.Windows/Registry
dsc -i $desired resource set -r Microsoft.Windows/Registry
```

Users would enter these commands:

```powershell
dsc config set --path ./example.dsc.config.yaml
dsc config set -p ./example.dsc.config.yaml

dsc config set --document $cfg
dsc config set -d $cfg

dsc resource set --resource Microsoft.Windows/Registry --input $desired
dsc resource set -r Microsoft.Windows/Registry -i $desired
```

### Proposed technical implementation details (optional)

1. Remove the `--input` and `--input-file` global options.
1. Add `--path` and `--document` options to the `config *` subcommands.
1. Add `--path` and `--input` options to the `resource get|set|test` subcommands.

The new argument names are suggestions - could be `--path` and `--input` or `--file` and `--data`, whatever semantics the team thinks will work best.

PR #305, opened 2024-02-02
Refactor input options

tgauth:
# PR Summary
- address #226 

## PR Context
- move `--input-file`/`--input` from global commands to relevant subcommands


dsc/src/args.rs
```diff
@@ -77,15 +73,40 @@ pub enum SubCommand {
 #[derive(Debug, PartialEq, Eq, Subcommand)]
 pub enum ConfigSubCommand {
     #[clap(name = "get", about = "Retrieve the current configuration")]
-    Get,
+    Get {
+        #[clap(short = 'd', long, help = "The input to pass to the configuration or resource", conflicts_with = "path")]
```
SteveL-MSFT, line 77:
```suggestion
        #[clap(short = 'd', long, help = "The document to pass to the configuration or resource", conflicts_with = "path")]
```

Change everywhere

dsc/src/subcommand.rs
```diff
@@ -411,20 +394,20 @@ pub fn resource(subcommand: &ResourceSubCommand, format: &Option<OutputFormat>,
                 table.print();
             }
         },
-        ResourceSubCommand::Get { resource, input, all } => {
+        ResourceSubCommand::Get { resource, input, path, all } => {
```
SteveL-MSFT, line 397: Should we have a function that takes `input` (maybe needs to be renamed to `document` now?), `stdin`, and `path`, and just returns the string contents so that the `get`, `set`, etc... functions just take one `input` parameter instead of all of them?  You already have the `get_input()` below that can be used.

dsc/src/util.rs
```diff
@@ -288,3 +289,61 @@ pub fn enable_tracing(trace_level: &TraceLevel, trace_format: &TraceFormat) {
...
+            return String::new();
+        },
+        _default => {
+            /* clap should handle these cases via conflicts_with so this should not get reached */
+            error!("Error: Invalid input");
+            exit(EXIT_INVALID_ARGS);
+        }
+    };
+
+    if value.is_empty() {
```
SteveL-MSFT, line 344: Unclear to me when this value is empty?  Maybe add a comment.  Should also have test cases when any of the inputs is empty string.

SteveL-MSFT (changes requested):

dsc/src/util.rs
tgauth: I'm not completely sure either - I put it in because the existing dsc_args.tests.ps1 were failing on the CI but not locally within `get_inputs` - it seems like it was detecting `Some` for both `stdin` and either `document` or `path` but `stdin` was actually empty.

SteveL-MSFT: I think you can technically pass an empty string via stdin: "" | dsc, although not entirely sure

SteveL-MSFT (changes requested):

dsc/src/util.rs
```diff
@@ -288,3 +289,71 @@ pub fn enable_tracing(trace_level: &TraceLevel, trace_format: &TraceFormat) {
...
+}
+
+pub fn get_input(input: &Option<String>, stdin: &Option<String>, path: &Option<String>) -> String {
+    let value = match (input, stdin, path) {
+        (Some(_), Some(_), None) | (None, Some(_), Some(_)) => {
+            error!("Error: Cannot specify both stdin and --input or --path");
+            exit(EXIT_INVALID_ARGS);
+        },
+        (Some(input), None, None) => {
+            info!("Reading input from command line parameter");
```
SteveL-MSFT, line 323: `info!()` should still be end-user useful.  You should use `debug!()` here.

SteveL-MSFT (changes requested):

SteveL-MSFT (approved):

PR #307, opened 2024-02-06
Export support for PS class-based resources

### LINKED ISSUE #183: Export support for provider-based resources

anmenaga:
# PR Summary

Fix #183 

2 updates:
1) Adds `Export` support for provider-based resources.
    It is done using `export` node in `*.dsc.resource.json` file similar to `get/set/test`.
2) Adds `Export` support to class-based PS resources.
   To support this a DscResource class has to implement a static `Export` function like so:

```
using namespace System.Collections.Generic

[DscResource()]
class TestClassResource
{
# omitting properties and Get,Set,Test methods here

    static [TestClassResource[]] Export()
    {
        $resultList = [List[TestClassResource]]::new()
        1..5 | %{
            $obj = New-Object TestClassResource
            $obj.Name = "Object$_"
            $obj.Prop1 = "Property of object$_"
            $resultList.Add($obj)
        }
        
        return $resultList.ToArray()
    }
}
```

### DIFF
--- a/dsc/src/resource_command.rs
+++ b/dsc/src/resource_command.rs
@@ -55,12 +55,24 @@ pub fn get(dsc: &DscManager, resource_type: &str, input: &Option<String>, stdin:
 }
 
 pub fn get_all(dsc: &DscManager, resource_type: &str, _input: &Option<String>, _stdin: &Option<String>, format: &Option<OutputFormat>) {
-    let Some(resource) = get_resource(dsc, resource_type) else {
+    let mut input = String::new() ;
+    let Some(mut resource) = get_resource(dsc, resource_type) else {
         error!("{}", DscError::ResourceNotFound(resource_type.to_string()).to_string());
         return
     };
+
     debug!("resource.type_name - {} implemented_as - {:?}", resource.type_name, resource.implemented_as);
-    let export_result = match resource.export() {
+    if let Some(requires) = &resource.requires {
+        input = add_type_name_to_json(input, resource.type_name.clone());
+        if let Some(pr) = get_resource(dsc, requires) {
+            resource = pr;
+        } else {
+            error!("Provider {} not found", requires);
+            return;
+        };
+    }
+
+    let export_result = match resource.export(input.as_str()) {
         Ok(export) => { export }
         Err(err) => {
             error!("Error: {err}");
@@ -203,14 +215,26 @@ pub fn schema(dsc: &DscManager, resource_type: &str, format: &Option<OutputForma
 }
 
 pub fn export(dsc: &mut DscManager, resource_type: &str, format: &Option<OutputFormat>) {
+    let mut input = String::new();
     let Some(dsc_resource) = get_resource(dsc, resource_type) else {
         error!("{}", DscError::ResourceNotFound(resource_type.to_string()).to_string());
         return
     };
 
+    let mut provider_resource: Option<&DscResource> = None;
+    if let Some(requires) = &dsc_resource.requires {
+        input = add_type_name_to_json(input, dsc_resource.type_name.clone());
+        if let Some(pr) = get_resource(dsc, requires) {
+            provider_resource = Some(pr);
+        } else {
+            error!("Provider {} not found", requires);
+            return;
+        };
+    }
+
     let mut conf = Configuration::new();
 
-    if let Err(err) = add_resource_export_results_to_configuration(dsc_resource, &mut conf) {
+    if let Err(err) = add_resource_export_results_to_configuration(dsc_resource, provider_resource, &mut conf, input.as_str()) {
         error!("Error: {err}");
         exit(EXIT_DSC_ERROR);
     }
--- a/dsc/src/resource_command.rs
+++ b/dsc/src/resource_command.rs
@@ -67,12 +67,12 @@ pub fn get_all(dsc: &DscManager, resource_type: &str, _input: &Option<String>, _
         if let Some(pr) = get_resource(dsc, requires) {
             resource = pr;
         } else {
-            error!("Provider {} not found", requires);
+            error!("Provider '{}' not found", requires);
             return;
         };
     }
 
-    let export_result = match resource.export(input.as_str()) {
+    let export_result = match resource.export(&input) {
         Ok(export) => { export }
         Err(err) => {
             error!("Error: {err}");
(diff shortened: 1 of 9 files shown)


dsc/src/resource_command.rs
```diff
@@ -55,12 +55,24 @@ pub fn get(dsc: &DscManager, resource_type: &str, input: &Option<String>, stdin:
...
     };
+
     debug!("resource.type_name - {} implemented_as - {:?}", resource.type_name, resource.implemented_as);
-    let export_result = match resource.export() {
+    if let Some(requires) = &resource.requires {
+        input = add_type_name_to_json(input, resource.type_name.clone());
+        if let Some(pr) = get_resource(dsc, requires) {
+            resource = pr;
+        } else {
+            error!("Provider {} not found", requires);
```
SteveL-MSFT, line 70: We want to start putting identifiers in single quotes as it makes it easier reading the error message which resource is at fault

```suggestion
            error!("Provider '{}' not found", requires);
```

dsc/src/resource_command.rs
```diff
@@ -55,12 +55,24 @@ pub fn get(dsc: &DscManager, resource_type: &str, input: &Option<String>, stdin:
...
+            return;
+        };
+    }
+
+    let export_result = match resource.export(input.as_str()) {
```
SteveL-MSFT, line 75: Use ref syntax instead of calling the function for consistency

```suggestion
    let export_result = match resource.export(&input) {
```

dsc/src/resource_command.rs
```diff
@@ -203,14 +215,26 @@ pub fn schema(dsc: &DscManager, resource_type: &str, format: &Option<OutputForma
...
+        } else {
+            error!("Provider {} not found", requires);
+            return;
+        };
+    }
+
     let mut conf = Configuration::new();
 
-    if let Err(err) = add_resource_export_results_to_configuration(dsc_resource, &mut conf) {
+    if let Err(err) = add_resource_export_results_to_configuration(dsc_resource, provider_resource, &mut conf, input.as_str()) {
```
SteveL-MSFT, line 237:
```suggestion
    if let Err(err) = add_resource_export_results_to_configuration(dsc_resource, provider_resource, &mut conf, &input) {
```

dsc/src/resource_command.rs
```diff
@@ -203,14 +215,26 @@ pub fn schema(dsc: &DscManager, resource_type: &str, format: &Option<OutputForma
...
         return
     };
 
+    let mut provider_resource: Option<&DscResource> = None;
+    if let Some(requires) = &dsc_resource.requires {
+        input = add_type_name_to_json(input, dsc_resource.type_name.clone());
+        if let Some(pr) = get_resource(dsc, requires) {
+            provider_resource = Some(pr);
+        } else {
+            error!("Provider {} not found", requires);
```
SteveL-MSFT, line 230:
```suggestion
            error!("Provider '{}' not found", requires);
```

powershellgroup/Tests/powershellgroup.config.tests.ps1
```diff
@@ -39,4 +39,24 @@ Describe 'PowerShellGroup resource tests' {
         $res.results.result.afterState[0].RebootRequired | Should -Not -BeNull
         $res.results.result.afterState[1].RebootRequired | Should -Not -BeNull
     }
+
+    It 'Export works on config with class-based resources' -Skip:(!$IsWindows){
```
SteveL-MSFT, line 43: There shouldn't be a reason to skip on non Windows

powershellgroup/Tests/powershellgroup.config.tests.ps1
```diff
@@ -39,4 +39,24 @@ Describe 'PowerShellGroup resource tests' {
...
+                resources:
+                - name: Class-resource Info
+                  type: PSTestModule/TestClassResource
+'@
+        $out = $yaml | dsc config export
+        $LASTEXITCODE | Should -Be 0
+        $res = $out | ConvertFrom-Json
+        $res.'$schema' | Should -BeExactly 'https://raw.githubusercontent.com/PowerShell/DSC/main/schemas/2023/08/config/document.json'
+        $res.'resources' | Should -Not -BeNullOrEmpty
+        $res.resources.count | Should -Be 5
```
SteveL-MSFT, line 60: Since we control the output and know exactly what should be returned, we should validate at least the Name property of the objects.

SteveL-MSFT (changes requested):
Fix build issues.  This design will allow class based resources implementing `export` to still work with DSC v2, right?  v2 should just ignore the new `export` method?  Ideally we should have a test to ensure we aren't breaking compat.

dsc/src/resource_command.rs
anmenaga: Good point; fixed.

dsc/src/resource_command.rs
anmenaga: Fixed.

dsc/src/resource_command.rs
anmenaga: Fixed.

powershellgroup/Tests/powershellgroup.config.tests.ps1
anmenaga: This export scenario still relies on PSDesiredStateConfiguration v2 (for resource discovery), which is not stable on non-windows.

powershellgroup/Tests/powershellgroup.config.tests.ps1
anmenaga: Updated to check properties of the first object.

anmenaga:
> This design will allow class based resources implementing `export` to still work with DSC v2, right?

Yes.

> v2 should just ignore the new `export` method?

Correct.

> Ideally we should have a test to ensure we aren't breaking compat.

We already test that implicitly. [We are using v2 throughout PowerShellGroup resource tests](https://github.com/PowerShell/DSC/blob/1fb90f6f9015658daeed21587f6bc6d2e9bbf454/build.ps1#L231).

SteveL-MSFT (approved):

PR #308, opened 2024-02-06
(DOCS) Changes to top-level options

michaeltlombardi:
# PR Summary

This change documents the updated options due to tracing formatting in the root command.

## PR Context

N/A

### DIFF
--- a/docs/reference/cli/dsc.md
+++ b/docs/reference/cli/dsc.md
@@ -1,6 +1,6 @@
 ---
 description: Command line reference for the 'dsc' command
-ms.date:     01/17/2024
+ms.date:     02/05/2024
 ms.topic:    reference
 title:       dsc
 ---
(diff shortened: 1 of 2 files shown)


docs/reference/cli/dsc.md
```diff
@@ -108,29 +108,47 @@ Type:      String
...
-- `Error`
-- `Warning`
-- `Info` (default)
-- `Debug`
-- `Trace`
+- `error`
+- `warning` (default)
+- `info`
+- `debug`
+- `trace`
```
SteveL-MSFT, line 123: We should mention that `trace` level will write out the contents of JSON input/output and isn't sanitized and only intended for developer use.  As such, such traces should NOT be redirected to storage as it may contain sensitive information.

SteveL-MSFT (changes requested):

docs/reference/cli/dsc.md
michaeltlombardi: Added the following warning:

> [!WARNING]
> The `trace` level output emits all JSON input/output that DSC processes during execution. DSC doesn't sanitize the JSON before emitting it. This trace level is only intended for developer use. Never redirect `trace` level output to storage as it may contain sensitive information.

SteveL-MSFT (approved):

Issue #129, opened 2023-08-03
Support YAML and JSONC for resource manifests


michaeltlombardi:
### Summary of the new feature / enhancement

As a DSC Resource author, I want to be able to author and maintain my resource manifests in YAML without needing to convert them to JSON as a build step, because authoring, maintaining, documenting, and reading YAML files is easier for me.

I'd like to be able to author my manifests in either JSONC or YAML so that I can include comments for both local development and maintenance purposes.

### Proposed technical implementation details (optional)

Support discovering DSC Resource manifests if they're in the `PATH` with any of the following file extensions:

- `.dsc.resource.json`
- `.dsc.resource.jsonc`
- `.dsc.resource.yaml`
- `.dsc.resource.yml`

SteveL-MSFT:
WG discussed this to allow resource manifest to be in YAML, but not mix of YAML/JSON in a single file.

PR #311, opened 2024-02-07
Support yaml resource discovery

tgauth:
# PR Summary
address #129 

## PR Context
- adds support for yml/yaml resource discovery, does not add support for jsonc

### DIFF
--- a/dsc_lib/src/discovery/command_discovery.rs
+++ b/dsc_lib/src/discovery/command_discovery.rs
@@ -52,9 +52,9 @@ impl CommandDiscovery {
                     if path.is_file() {
                         let file_name = path.file_name().unwrap().to_str().unwrap();
                         let file_name_lowercase = file_name.to_lowercase();
-                        if file_name_lowercase.ends_with(".dsc.resource.json") |
-                        file_name_lowercase.ends_with(".dsc.resource.yaml") |
-                        file_name_lowercase.ends_with(".dsc.resource.yml") {
+                        if file_name_lowercase.ends_with(".dsc.resource.json") ||
+                           file_name_lowercase.ends_with(".dsc.resource.yaml") ||
+                           file_name_lowercase.ends_with(".dsc.resource.yml") {
                             let resource = match load_manifest(&path)
                             {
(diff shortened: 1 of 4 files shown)


dsc_lib/src/discovery/command_discovery.rs
```diff
@@ -50,7 +51,9 @@ impl CommandDiscovery {
                     let path = entry.path();
                     if path.is_file() {
                         let file_name = path.file_name().unwrap().to_str().unwrap();
-                        if file_name.to_lowercase().ends_with(".dsc.resource.json") {
+                        if file_name.to_lowercase().ends_with(".dsc.resource.json") |
```
SteveL-MSFT, line 54: since we use the lowercase version multiple times, we should probably do it once and store in a variable

SteveL-MSFT (changes requested):

dsc_lib/src/discovery/command_discovery.rs
```diff
@@ -50,7 +51,10 @@ impl CommandDiscovery {
                     let path = entry.path();
                     if path.is_file() {
                         let file_name = path.file_name().unwrap().to_str().unwrap();
-                        if file_name.to_lowercase().ends_with(".dsc.resource.json") {
+                        let file_name_lowercase = file_name.to_lowercase();
+                        if file_name_lowercase.ends_with(".dsc.resource.json") |
```
SteveL-MSFT, line 55: Should you use the double bar `||` logical OR operator instead of the bitwise OR operator here?

dsc_lib/src/discovery/command_discovery.rs
```diff
@@ -50,7 +51,10 @@ impl CommandDiscovery {
...
+                        file_name_lowercase.ends_with(".dsc.resource.yaml") |
```
SteveL-MSFT, line 56: These should be indented under the `if`

SteveL-MSFT (changes requested):

SteveL-MSFT (approved):

PR #313, opened 2024-02-08
Added DSCConfigRoot env var in configs

### LINKED ISSUE #75: Need a way to reference file paths relative to the config file

A `winget configure` scenario is that a git repo has a config.dsc.yaml which uses a resource that uses a file within the repo.  Ask is to have a built-in variable similar to `$PSScriptRoot` like `$DSCConfigRoot` that contains the path to the config file.  

_No response_

anmenaga:
# PR Summary

Fix #75 

For config operations, if `--path` dsc parameter is used, then:
1) `DSCConfigRoot` env var is defined and inherited by child resource processes that can be used to determine the location of the configuration file that is running;
2) `envvar` function can be used in the configuration text like so:
```
resources:
- name: Class-resource Info
  type: PSTestModule/TestClassResource
  properties:
    Name: Instance1
    ResourcePath: "[envvar('DSCConfigRoot')]"
```

Notes:
1) if configuration is piped using stdin or `-d`/`--document` parameter, then value of env var is empty string.
2) for non-config operations env var is not created.

### DIFF
--- a/dsc/src/util.rs
+++ b/dsc/src/util.rs
@@ -18,7 +18,6 @@ use dsc_lib::{
 };
 use schemars::{schema_for, schema::RootSchema};
 use serde_yaml::Value;
-use serde_json::json;
 use std::collections::HashMap;
 use std::env;
 use std::path::Path;
@@ -361,7 +360,7 @@ pub fn get_input(input: &Option<String>, stdin: &Option<String>, path: &Option<S
     parse_input_to_json(&value)
 }
 
-pub fn process_macros(json_string: &str, config_path: &str) -> String
+pub fn set_dscconfigroot(config_path: &str)
 {
     let path = Path::new(config_path);
     let config_root = match path.parent()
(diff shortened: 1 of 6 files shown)


dsc/src/util.rs
```diff
@@ -357,3 +360,28 @@ pub fn get_input(input: &Option<String>, stdin: &Option<String>, path: &Option<S
...
+    // we are replacing a substring in json, so the new substring must be properly escaped per json rules
+    let v = json!(config_root);
+    let mut json_escaped_config_root = v.to_string();
+    json_escaped_config_root.pop(); // remove last double quote
+    if !json_escaped_config_root.is_empty() {
+        json_escaped_config_root.remove(0); // remove first double quote
+    }
+
+    debug!("Escaped json 'DSCConfigRoot' is '{}'", json_escaped_config_root);
+    json_string.replace("_DSCConfigRoot_", &json_escaped_config_root)
```
SteveL-MSFT, line 386: We don't want to introduce this syntax within the config as a way to get env vars even this special one.  As noted in the issue, we eventually will want a function to retrieve any env var.

SteveL-MSFT (changes requested):

dsc/src/util.rs
anmenaga: I agree. Added `envvar` function and updated PR description.

SteveL-MSFT (approved):

Issue #309, opened 2024-02-06
Add `Methods` to table view for `dsc resource list`


SteveL-MSFT:
### Summary of the new feature / enhancement

It would be useful to config authors to see what methods `get`, `set`, `test`, `export` are supported by a resource within the table view

michaeltlombardi:
Are there any resources that can't support `get`? Any resource that supports `get` supports `test` (either directly implemented or synthetic through DSC), I think.

tgauth:
I don't think there's any that don't support `get` because it's a required field in the resource manifest but not sure about `test` as that's still an optional field in the resource manifest

michaeltlombardi:
Even when a resource doesn't implement the `test` command, DSC is able to use [synthetic testing][01] to compare the actual state to the desired state:

> When this property isn't defined, DSC uses a synthetic test method for the resource. The synthetic test method:
>
> 1. Gets the actual state of the instance using the resource's `get` method.
> 1. Compares every defined property of the instance's desired state to the actual state.
> 1. If the desired state of a property isn't equal to the actual state of that property, DSC reports that the instance isn't in the desired state.
>
> Because the synthetic test only checks for equivalency, it can't accurately test resources with properties that can't be evaluated with equivalency alone. For example, if a resource manages package versions and allows setting the version to `latest`, DSC would report an instance with a version of `3.1.0` as being out of the desired state, even if `3.1.0` is the latest version of the package.

[01]: https://learn.microsoft.com/powershell/dsc/reference/schemas/resource/manifest/test?view=dsc-3.0#description

SteveL-MSFT:
I think the intent of this right now is to simply expose what the resource advertises it implements.  So I would expect it to show `get`, `set`, `test`, and/or `export`.  Although `get` is currently mandatory, that may or may not change in the future and I think it would be confusing ot users to not see it shown.  Similarly, although dsc will synthesize `test`, I think it makes sense to show whether the resource implements it directly and it's via docs that indicate `test` is available even if the resource doesn't directly implement it as it does have limitations (direct comparison).

PR #314, opened 2024-02-08
add methods column to table view

tgauth:
# PR Summary

- address #309 
## PR Context

- add methods column to table view listing supported commands

### DIFF
--- a/dsc/src/subcommand.rs
+++ b/dsc/src/subcommand.rs
@@ -320,14 +320,15 @@ pub fn resource(subcommand: &ResourceSubCommand, stdin: &Option<String>) {
         ResourceSubCommand::List { resource_name, description, tags, format } => {
 
             let mut write_table = false;
-            let mut table = Table::new(&["Type", "Version", "Requires", "Description"]);
+            let mut methods = String::new();
+            let mut table = Table::new(&["Type", "Version", "Methods", "Requires", "Description"]);
             if format.is_none() && atty::is(Stream::Stdout) {
-                // write as table if fornat is not specified and interactive
+                // write as table if format is not specified and interactive
                 write_table = true;
             }
             for resource in dsc.list_available_resources(&resource_name.clone().unwrap_or_default()) {
                 // if description is specified, skip if resource description does not contain it
-                if description.is_some() || tags.is_some() {
+                if description.is_some() || tags.is_some() || write_table {
                     let Some(ref resource_manifest) = resource.manifest else {
                         continue;
                     };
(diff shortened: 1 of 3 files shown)


dsc/src/subcommand.rs
```diff
@@ -320,14 +320,15 @@ pub fn resource(subcommand: &ResourceSubCommand, stdin: &Option<String>) {
...
+            let mut table = Table::new(&["Type", "Version", "Methods", "Requires", "Description"]);
             if format.is_none() && atty::is(Stream::Stdout) {
-                // write as table if fornat is not specified and interactive
+                // write as table if format is not specified and interactive
                 write_table = true;
             }
             for resource in dsc.list_available_resources(&resource_name.clone().unwrap_or_default()) {
                 // if description is specified, skip if resource description does not contain it
-                if description.is_some() || tags.is_some() {
+                if description.is_some() || tags.is_some() || write_table {
```
SteveL-MSFT, line 331: This doesn't seem right.  The description on line 329 probably should be updated, but this `if` is if the user specified either the `--description` or `--tags` filters then it goes in here to perform the filtering.

dsc/src/subcommand.rs
```diff
@@ -359,16 +358,20 @@ pub fn resource(subcommand: &ResourceSubCommand, stdin: &Option<String>) {
                                 }
                             }
                         }
-                        if !found {
-                            continue;
-                        }
+                        if !found { continue; }
                     }
+
+                    methods = "get".to_string();
```
SteveL-MSFT, line 364: Should probably still keep a check here for `get` as the discussion of whether it must be mandatory might change in the future.

dsc/src/subcommand.rs
```diff
@@ -320,14 +320,15 @@ pub fn resource(subcommand: &ResourceSubCommand, stdin: &Option<String>) {
         ResourceSubCommand::List { resource_name, description, tags, format } => {
 
             let mut write_table = false;
-            let mut table = Table::new(&["Type", "Version", "Requires", "Description"]);
+            let mut methods = String::new();
```
SteveL-MSFT, line 323: I think it would be better to have this as a vec of strings which has a `.join()` function you can use

dsc_lib/src/dscresources/resource_manifest.rs
```diff
@@ -24,6 +24,7 @@ pub struct ResourceManifest {
     /// Tags for the resource.
     pub tags: Option<Vec<String>>,
     /// Details how to call the Get method of the resource.
+    // If this field becomes optional, will need to update dsc resource list table view methods column
```
SteveL-MSFT, line 27: I'd remove this comment

SteveL-MSFT (changes requested):

dsc/src/subcommand.rs
tgauth: I added the `write_table` here because it also requires the resource manifest, and I didn't want to duplicate effort to retrieve it.

I initially tried pulling the resource manifest code out of this `if` but the `continue` / `break` logic broke and would require additional refactoring to fix.

I tested `dsc resource list` with and without `--description` and I think it behaves as expected.

dsc/src/subcommand.rs
tgauth: That makes sense - what do you suggest checking? 
The get field of the struct isn't optional so the `is_some()` check doesn't apply.

SteveL-MSFT (approved):

Issue #266, opened 2023-11-09
Make output for group and provider resources in `dsc config *` more readable


michaeltlombardi:
### Summary of the new feature / enhancement

> As a user, I want to be able to use group and provider resources in my configuration documents and understand the output for `dsc config` commands without having to know how DSC works internally or investigate the data to know which keys mean what.

While inspecting the output for group and provider resources, I noticed that the output structure for those resources is very difficult to read.

1. Define a configuration that uses a group or provider resource.

   <details><summary>Configuration document</summary>

   ```yaml
   # repro.dsc.config.yaml
   $schema: https://raw.githubusercontent.com/PowerShell/DSC/main/schemas/2023/08/config/document.json
   resources:
     - name: current user registry
       type: Microsoft.Windows/Registry
       properties:
         keyPath: HKCU\example
         _exist:  true
       dependsOn:
         - "[resourceId('DSC/AssertionGroup','my assertions')]"
     - name: my assertions
       type: DSC/AssertionGroup
       properties:
         $schema: https://raw.githubusercontent.com/PowerShell/DSC/main/schemas/2023/08/config/document.json
         resources:
           - name: os
             type: Microsoft/OSInfo
             properties:
               family: Windows
           - name: system root
             type: Microsoft.Windows/Registry
             properties:
               keyPath: HKLM\Software\Microsoft\Windows NT\CurrentVersion
               valueName: SystemRoot
               valueData:
                 String: Z:\Windows
   ```

  </details>

2. Get the current state of the configuration.

   ```sh
   dsc --input-file ./assertion.dsc.config.yaml config get
   ```

   <details><summary>Actual Output</summary>

   ```yaml
   results:
   - name: my assertions
     type: DSC/AssertionGroup
     result:
       actualState:
         results:
         - name: os
           type: Microsoft/OSInfo
           result:
             desiredState:
               family: Windows
             actualState:
               $id: https://developer.microsoft.com/json-schemas/dsc/os_info/20230303/Microsoft.Dsc.OS_Info.schema.json
               family: Windows
               version: 10.0.22621
               edition: Windows 11 Enterprise
               bitness: '64'
             inDesiredState: true
             differingProperties: []
         - name: system root
           type: Microsoft.Windows/Registry
           result:
             desiredState:
               keyPath: HKLM\Software\Microsoft\Windows NT\CurrentVersion
               valueName: SystemRoot
               valueData:
                 String: Z:\Windows
             actualState:
[...]

PR #318, opened 2024-02-18
Enable group resource output schema to simplify config output that uses groups

SteveL-MSFT:
# PR Summary

- New schema for group resources for `get`, `set`, and `test`
- `dsc` as a resource has new `--as-group` to use this new schema
- `dsc` as a resource has new `--as-get` to use with AssertionGroup so that a `test` operation still returns a `get` output to conform with new schema
- new debug and trace needed to identify issue
- new tests for Group which also uses nested groups
- fix input validation to use `validate` method for group resources instead of relying only on schema
- fix config validation so that resources only get their portion of the config instead of the whole config doc (which resulted in unintended infinite recursion when `dsc` was the resource)
- added test for brew example resource and yaml
- update `osinfo` readme to match code
- updated some existing debug to be traces conforming with new standard

## PR Context

Addresses the `get` part of https://github.com/PowerShell/DSC/issues/266, `set` and `test` will be addressed in separate PRs

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,8 @@ changes since the last release, see the [diff on GitHub][unreleased].
 <!-- Unreleased comparison link -->
 [unreleased]: https://github.com/PowerShell/DSC/compare/v3.0.0-alpha.4...main
 
+<!-- Add entries between releases under the appropriate section heading here  -->
+
 ### Changed
 
 - Updated the options for the `dsc` root command:
@@ -62,7 +64,16 @@ changes since the last release, see the [diff on GitHub][unreleased].
 
   </details>
 
-<!-- Add entries between releases under the appropriate section heading here  -->
+- Added support for authoring DSC Resource manifests in YAML. DSC now recognizes resource manifests
+  that use the `.dsc.resource.yml` or `.dsc.resource.yaml` file extension instead of only
+  `.dsc.resource.json`.
+
+  <details><summary>Related work Items</summary>
+
+  - Issues: [#129][#129]
+  - PRs: [#311][#311]
+
+  </details>
 
 ## [v3.0.0-alpha.4][release-v3.0.0-alpha.4] - 2023-11-14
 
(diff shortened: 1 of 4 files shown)


dsc/examples/groups.dsc.yaml
```diff
@@ -0,0 +1,38 @@
...
+- name: Last Group
+  type: DSC/Group
+  properties:
+    $schema: https://raw.githubusercontent.com/PowerShell/DSC/main/schemas/2023/08/config/document.json
+    resources:
+    - name: Last
+      type: Test/Echo
+      properties:
+        output: Last
+  dependsOn:
```
anmenaga, line 13: seems like too short ident for `dependsOn`.

dsc/assertion.dsc.resource.json
```diff
@@ -7,14 +7,17 @@
     "executable": "dsc",
     "args": [
       "config",
-      "test"
+      "--as-group",
```
anmenaga, line 10: It would help if there were short examples of format for both `--as-group` and `--as-get` in PR description (and docs).

dsc_lib/src/configure/config_result.rs
```diff
@@ -102,6 +147,27 @@ pub struct ResourceTestResult {
     pub result: TestResult,
 }
 
+#[derive(Debug, Clone, PartialEq, Deserialize, Serialize, JsonSchema)]
+#[serde(deny_unknown_fields)]
```
anmenaga, line 151: Is there a strong reason to deny_unknown_fields ?

anmenaga (approved):

dsc/assertion.dsc.resource.json
SteveL-MSFT: I provided descriptions of those two switches in the PR description.  They won't be doc'd since they are hidden and intended to be only used by dsc itself.

dsc_lib/src/configure/config_result.rs
SteveL-MSFT: We can evaluate if we want to allow arbitrary fields later if a scenario comes up

dsc/examples/groups.dsc.yaml
SteveL-MSFT: This `dependsOn` is for the `Last Group` whereas the lines above are the resources within the group, so the indent here is correct

PR #321, opened 2024-02-21
Add run command resource

### LINKED ISSUE #302: `RunCommand` resource

There are cases where a user already has scripts they've written and they want to incorporate it into their config as they start to transition to declarative.  We don't want to promote usage of this, but also not have barriers to prevent adoption.  This resource would implement `get`, `set`, and `test` which simply executes the defined command-line.

The `RunCommand` resource takes a single command-line to execute and returns the exit code and writes a WARNING message containing any STDERR output and an INFO message containing any STDOUT output.  An ERROR is returned if the exit code is non-zero.  This means that any non-standard command that emits a successful non-zero exit code needs to be wrapped in a script.

tgauth:
# PR Summary
- fix #302 

## PR Context
- adds `RunCommandOnSet` resource, without tracing for stdout/stderr propagated back to DSC


runcommandonset/src/utils.rs
```diff
@@ -0,0 +1,178 @@
...
+///
+/// * `executable` - The command to execute
+/// * `args` - Optional arguments to pass to the command
+///
+/// # Errors
+///
+/// Error message then exit if the command fails to execute or stdin/stdout/stderr cannot be opened.
+pub fn invoke_command(executable: &str, args: Option<Vec<String>>) -> (i32, String, String) {
+    // originally implemented in dsc_lib/src/dscresources/command_resource.rs
+    debug!("Invoking command {} with args {:?}", executable, args);
```
SteveL-MSFT, line 123: This should probably be a trace as this is user supplied args rather than a resource being executed so may contain sensitive data

runcommandonset/tests/runcommandonset.get.tests.ps1
```diff
@@ -0,0 +1,53 @@
...
+    BeforeAll {
+        $oldPath = $env:DSC_RESOURCE_PATH
+        $env:DSC_RESOURCE_PATH = Join-Path $PSScriptRoot ".."
+    }
+
+    AfterAll {
+        $env:DSC_RESOURCE_PATH = $oldPath
+    }
+
+    It 'Input can be sent to the resource via stdin json' {
```
SteveL-MSFT, line 14: You really only need either hte JSON or YAML test as it's really exercising `dsc` as `dsc` always passes JSON to the resource.

runcommandonset/tests/runcommandonset.set.tests.ps1
```diff
@@ -0,0 +1,72 @@
...
+
+    It 'Input can be sent to the resource via stdin json' {
+        $json = @"
+        {
+            "executable": "pwsh",
+            "arguments": ["-Command", "echo hello world"],
+        }
+"@
+
+        $json | dsc resource set -r Microsoft.DSC.Transitional/RunCommandOnSet > $TestDrive/output.txt
```
SteveL-MSFT, line 34: You don't have to redirect the output to a file, you can always assign it to a variable: $output = $json | dsc ...

runcommandonset/tests/runcommandonset.set.tests.ps1
```diff
@@ -0,0 +1,72 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+Describe 'tests for runcommandonset set' {
+    BeforeAll {
+        $oldPath = $env:DSC_RESOURCE_PATH
+        $env:DSC_RESOURCE_PATH = Join-Path $PSScriptRoot ".."
+        $yaml = @"
+executable: pwsh
+arguments:
```
SteveL-MSFT, line 10: I think you're missing a test with an executable, but no arguments.

runcommandonset/tests/runcommandonset.set.tests.ps1
```diff
@@ -0,0 +1,72 @@
...
+
+    It 'Executable is a required input via CLI arguments' {
+        $null = runcommandonset set -a foo
+        $LASTEXITCODE | Should -Be 4
+    }
+
+    It 'Executable is a required input via STDIN' {
+        '{ "arguments": "foo" }' | dsc resource set -r Microsoft.DSC.Transitional/RunCommandOnSet
+        $LASTEXITCODE | Should -Be 2
+    }
```
SteveL-MSFT, line 71: I think you're missing the executable doesn't exist test case

SteveL-MSFT (changes requested):

runcommandonset/tests/runcommandonset.set.tests.ps1
```diff
@@ -0,0 +1,93 @@
...
+
+    It 'STDOUT captured via STDERR when calling resource directly' {
+        $input_json = @"
+        {
+            "executable": "pwsh",
+            "arguments": ["-Command", "echo hello world"]
+        }
+"@
+        $input_json | runcommandonset set 2> $TestDrive/output.txt
```
SteveL-MSFT, line 39: Change all redirection to file to just assigning to variable

runcommandonset/tests/runcommandonset.set.tests.ps1
```diff
@@ -0,0 +1,93 @@
...
+        $LASTEXITCODE | Should -Be 4
+    }
+
+    It 'Executable is a required input via STDIN' {
+        $null = '{ "arguments": "foo" }' | dsc resource set -r Microsoft.DSC.Transitional/RunCommandOnSet
+        $LASTEXITCODE | Should -Be 2
+    }
+
+    It 'Executable can be provided without arguments' {
+        $null = '{ "executable": "pwsh" }' | dsc resource set -r Microsoft.DSC.Transitional/RunCommandOnSet
```
SteveL-MSFT, line 67: instead of assigning to `$null`, you should store it in a variable and confirm that the executed binary has the expected exit code

SteveL-MSFT (changes requested):
address the comments

runcommandonset/tests/runcommandonset.set.tests.ps1
tgauth: Can I capture `STDERR` when it's assigned to a variable? I'm making sure the expected `STDOUT` and `STDERR` are written there

runcommandonset/src/runcommand.rs
```diff
@@ -0,0 +1,31 @@
...
+
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Deserialize, Clone, PartialEq, Serialize)]
+pub struct RunCommand {
+    pub executable: String,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub arguments: Option<Vec<String>>,
+    // default value for exit code is 0
+    #[serde(default, skip_serializing_if = "is_default")]
```
SteveL-MSFT, line 12: Does this mean it doesn't serialize `exit_code` if the value is 0?  Also, for serialization, this should be renamed to `exitCode` to align with JSON naming convention.

runcommandonset/tests/runcommandonset.set.tests.ps1
SteveL-MSFT: `stderr` would still need to be redirected unfortunately.  There's a feature Jim was writing to redirect to a variable, but it doesn't exist yet

SteveL-MSFT (changes requested):
One more minor feedback

runcommandonset/src/runcommand.rs
tgauth: Yes - I did that so `exit_code` was not required as an input but would still be initialized internally to `0`, but might make more sense to accomplish that with `Option` depending on user feedback?

If `exit_code` isn't provided, I see no need to write it back out for a `get` command, but I could see it being useful for a `set` command. 

As is, for `set` the `changedProperties` will capture any difference in `exit_code`

SteveL-MSFT (approved):

Issue #139, opened 2023-08-04
Extend schema manifest to include `kind`


michaeltlombardi:
### Summary of the new feature / enhancement

As a resource author and integrating vendor, I want to be able to quickly distinguish between standard command-based DSC Resources, resource groups, and resource providers without having to inspect the full manifest in detail

### Proposed technical implementation details (optional)

Add a `kind` property to the DSC Resource manifest that supports three enum values:

- `standard` - a "normal" command-based DSC Resource, not a group or provider. This is the default value if the key isn't specified.
- `group` - a resource group, always has nested resources as properties under the `resources` keyword. _May_ have other properties that influence how the resource group behaves. When `kind` is set to `group`, the `validate` property is required.
- `provider` - a resource provider, a special resource group that enables the use of non-command-based DSC Resources. When `kind` is set to `provider`, the `validate` and `provider` properties are required.

After `kind` is included in the resource manifest, the `dsc resource list` command should surface the resource kind in the default output and at the top-level.

SteveL-MSFT:
Alternate proposal, in the resource.json, have the `kind` be the property name instead of the property value:

So instead of what we have today:

```json
{
    "$schema": "https://raw.githubusercontent.com/PowerShell/DSC/main/schemas/2023/08/bundled/resource/manifest.json",
    "type": "DSC/PowerShellGroup",
    "version": "0.1.0",
    "description": "Resource provider to classic DSC Powershell resources.",
    "tags": [
      "PowerShell"
    ],
    "provider": {
      "list": {
        "executable": "pwsh",
        "args": [
          "-NoLogo",
          "-NonInteractive",
          "-NoProfile",
          "-Command",
          "./powershellgroup.resource.ps1 List"
        ]
        },
      "config": "full"
    },
    "get": {
      "executable": "pwsh",
      "args": [
        "-NoLogo",
        "-NonInteractive",
        "-NoProfile",
        "-Command",
        "$Input | ./powershellgroup.resource.ps1 Get"
      ],
      "input": "stdin"
      },
```

This becomes:

```json
{
    "$schema": "https://raw.githubusercontent.com/PowerShell/DSC/main/schemas/2023/08/bundled/resource/manifest.json",
    "type": "DSC/PowerShellGroup",
    "version": "0.1.0",
    "description": "Resource provider to classic DSC Powershell resources.",
    "tags": [
      "PowerShell"
    ],
    "provider": {
      "list": {
        "executable": "pwsh",
        "args": [
[...]

PR #338, opened 2024-03-01
add `kind` to dscresource

SteveL-MSFT:
# PR Summary

Add `kind` property to dscresource.  Only `Group` needs to be explicitly specified.  `Adapter` can be derived and `Resource` is default.

## PR Context

Fix https://github.com/PowerShell/DSC/issues/139


SteveL-MSFT:
Merging as I need these changes for some other coming PRs.  Any feedback we can address as separate PRs.

PR #340, opened 2024-03-02
enable building msix package

SteveL-MSFT:
# PR Summary

- new `-msix` switch to `build.ps1` along with `-skipbuild` to not perform a full build to produce msix
- all assets and manifest to build msix including placeholder logos
- add home directory of `dsc.exe` as first path to search for resources so that the built-in groups are found along with powershell and wmi
- removed add-path.ps1 as it's no longer needed with home directory searching

## PR Context

Build msix to publish to Store and WinGet making it easier to deploy dsc.  This package is currently published unlisted at https://www.microsoft.com/store/apps/9PCX3HX4HZ0Z


build.ps1
```diff
@@ -294,4 +295,116 @@ if ($Test) {
...
+        if ($architecture -eq 'aarch64-pc-windows-msvc') {
+            $arch = 'arm64'
+        }
+        else {
+            $arch = 'x64'
+        }
+
+        $makeappx = Get-ChildItem -Recurse -Path (Join-Path ${env:ProgramFiles(x86)} 'Windows Kits\10\bin\*\' $arch) -Filter makeappx.exe | Sort-Object FullName -Descending | Select-Object -First 1
+        if ($null -eq $makeappx) {
+            throw "makeappx not found"
```
anmenaga, line 319: It would help if there was a download link or instruction here how to get makeappx.

anmenaga (approved):

build.ps1
SteveL-MSFT: Fair point, I'll update the text, but it's basically installing Windows SDK

PR #344, opened 2024-03-04
RebootPending resource

### LINKED ISSUE #333: `RebootRequired` resource

From some discussion out of https://github.com/PowerShell/DSC/issues/50, it may make sense to have a `RebootRequired` resource that has knowledge of registry/file/api queries to indicate if (Windows in this case) the OS has a pending reboot.  This resource would be used in an `Assertion` so that config is not applied until the system no longer needs a reboot.

_No response_

anmenaga:
# PR Summary

Fix #333

Quick and dirty implementation of `RebootPending` resource.
- Windows only;
- requires Windows PowerShell;
- tested manually.

Example:
```
PS C:\DSCv3> dsc resource get -r Microsoft/RebootPending 
actualState:
  RebootPending: true
```


reboot_pending/reboot_pending.dsc.resource.json
```diff
@@ -0,0 +1,16 @@
+{
+    "$schema": "https://raw.githubusercontent.com/PowerShell/DSC/main/schemas/2023/08/bundled/resource/manifest.json",
+    "description": "Returns info about pending reboot.",
+    "type": "Microsoft/RebootPending",
```
SteveL-MSFT, line 4: Should be `Microsoft.Windows` since it's Windows specific.

reboot_pending/reboot_pending.resource.ps1
```diff
@@ -0,0 +1,12 @@
+ if (Get-ChildItem "HKLM:\Software\Microsoft\Windows\CurrentVersion\Component Based Servicing\RebootPending" -EA Ignore) { return @{ RebootPending = $true } | ConvertTo-Json -Compress }
```
SteveL-MSFT, line 1: Do we know if these reg paths are all publicly documented?

SteveL-MSFT (changes requested):

reboot_pending/reboot_pending.resource.ps1
anmenaga: Yes; I found them mentioned here:
https://learn.microsoft.com/en-us/mem/configmgr/core/servers/deploy/install/list-of-prerequisite-checks#pending-system-restart

reboot_pending/reboot_pending.dsc.resource.json
anmenaga: Updated.

build.ps1
```diff
@@ -122,6 +122,7 @@ $projects = @(
     "y2j",
     "wmi-adapter",
     "resources/brew",
+    "reboot_pending",
```
SteveL-MSFT, line 125: this needs to be moved to `$windows_projects` along with wmi-adapter

reboot_pending/reboot_pending.resource.ps1
SteveL-MSFT, line 1: Add a comment to the docs on the reboot registry locations

SteveL-MSFT (changes requested):

build.ps1
anmenaga: Updated.

reboot_pending/reboot_pending.resource.ps1
anmenaga: Updated.

PR #346, opened 2024-03-05
use bit flags for methods in table view

### LINKED ISSUE #329: DSC resource table change for `methods`

The current `Methods` column takes up lots of space, we should change it to "bit flags":

```console
Type                        Version  Methods                 Requires        Description
---------------------------------------------------------------------------------------------------------------------------------------
DSC.PackageManagement/Brew  0.1.0    gs-e                        DSC resource to manage Homebrew packages
DSC/AssertionGroup          0.1.0    gst-                          `test` will be invoked for all resources in the supplied …
DSC/Group                   0.1.0    gst-                          All resources in the supplied configuration is treated as
DSC/ParallelGroup           0.1.0    gst-                          All resources in the supplied configuration run concurren…
```

_No response_

tgauth:
# PR Summary

- fix #329
## PR Context
- new rendering:

image


dsc/src/subcommand.rs
```diff
@@ -436,18 +436,18 @@ pub fn resource(subcommand: &ResourceSubCommand, stdin: &Option<String>) {
                         if !found { continue; }
                     }
 
-                    methods = vec!["get".to_string()];
-                    if manifest.set.is_some() { methods.push("set".to_string()); }
-                    if manifest.test.is_some() { methods.push("test".to_string()); }
-                    if manifest.export.is_some() { methods.push("export".to_string()); }
+                    methods = vec!["g".to_string()];
```
SteveL-MSFT, line 439: Each `to_string()` will result in an allocation.  I wonder if we can preallocate one string `----` and replace the characters instead to reduce allocations

SteveL-MSFT (changes requested):

dsc/src/subcommand.rs
```diff
@@ -394,7 +394,7 @@ pub fn resource(subcommand: &ResourceSubCommand, stdin: &Option<String>) {
         ResourceSubCommand::List { resource_name, description, tags, format } => {
 
             let mut write_table = false;
-            let mut methods: Vec<String> = Vec::new();
+            let mut methods = String::new();
```
SteveL-MSFT, line 397: Should allocate your string from line 439 here

SteveL-MSFT (changes requested):

SteveL-MSFT (approved):

PR #349, opened 2024-03-06
add div function

tgauth:
# PR Summary
- implement https://learn.microsoft.com/en-us/azure/azure-resource-manager/templates/template-functions-numeric#div

## PR Context
- part of #57

### DIFF
--- a/dsc_lib/src/functions/div.rs
+++ b/dsc_lib/src/functions/div.rs
@@ -1,0 +1,77 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+use crate::DscError;
+use crate::configure::context::Context;
+use crate::functions::{AcceptedArgKind, Function};
+use serde_json::Value;
+use tracing::debug;
+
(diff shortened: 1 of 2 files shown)


dsc_lib/src/functions/div.rs
```diff
@@ -0,0 +1,77 @@
...
+        2
+    }
+
+    fn accepted_arg_types(&self) -> Vec<AcceptedArgKind> {
+        vec![AcceptedArgKind::Number]
+    }
+
+    fn invoke(&self, args: &[Value], _context: &Context) -> Result<Value, DscError> {
+        debug!("div function");
+        if let Some(value) = args[0].as_i64().unwrap().checked_div(args[1].as_i64().unwrap()) {
```
SteveL-MSFT, line 28: In general, use `?` operator instead of just having `unwrap()` panic.

SteveL-MSFT (changes requested):

SteveL-MSFT (approved):

Issue #258, opened 2023-11-08
Add metadata in config to indicate a config requires admin or non-admin


SteveL-MSFT:
### Summary of the new feature / enhancement

Some configurations may require admin or may require user to not be admin to apply.  The actual configuration can use an AssertionGroup to a resource that checks if user is elevated/root.  However, it might make sense to also have this info in the `metadata` property of the configuration, but this would just be documentation.

denelon:
WinGet package manifests have:
https://github.com/microsoft/winget-cli/blob/296a53dbedffabfd14b60f43e1a7abae0200cb75/schemas/JSON/manifests/v1.6.0/manifest.installer.1.6.0.json#L512

Something similar for a DSC Resource would mean errors caused by mismatched elevation requirements could be caught before execution.

JohnMcPMS:
A consistent request from our customers is to have a single configuration document that can be used in multiple contexts to achieve the overall desired configuration.  On Windows, this means the ability to group the configuration elements to run under the `SYSTEM` user and an individual user (as well as that user's unrestricted token if applicable).  I believe that this also means that ability to state the desired action to take when the orchestrator is not run in the desired context; for example, being able to simply ignore the entire group.  They are generally accepting of the need to apply the configuration multiple times, so long as it is a single document.  I believe that the major reason for this single document stance is to shift the burden of splitting the document more into the authoring side, removing the need to store/transport multiple data streams in the backend systems.  Given the benefits of having a tool like this beyond these customers though, the additional requests are minimal and thus this seems like the best method to provide.

My understanding of the requirements on any such group to provide this behavior are thus:
1. The ability to specify the target user context in which the group applies
  a. With analogous values to: `SYSTEM`, `User`, `RestrictedUser`, `UnrestrictedUser`
  b. `SYSTEM` targets the actual `SYSTEM` user context on Windows.
  c. `User` targets any standard user context
  d. `RestrictedUser` targets the restricted token for a standard user when UAC is enabled
  e. `UnrestrictedUser` targets the unrestricted token for a standard user when UAC is enabled
3. The ability to specify the behavior of the group when *not* in the desired context
  a. With analogous values to: `Ignore`, `Switch`
  b. `Ignore` would result in a no-op when attempting to `Set` the state, and presumably a no-op when attempting to `Test` as well.
  c. `Switch` would run the group's sub elements in the target context is possible.

My expectation on the average use case would be:
```YAML
resources:
  - name: SYSTEM group
[...]

PR #351, opened 2024-03-07
Enable metadata in config to declare if admin is required or not

SteveL-MSFT:
# PR Summary

Enable using metadata to declare if a config requires admin or non-admin.  A helper function library added using different methods depending on Windows and non-Windows.  This can be used later with RunAsGroup resource.

## PR Context

Fix https://github.com/PowerShell/DSC/issues/258

### DIFF
--- a/dsc_lib/src/configure/mod.rs
+++ b/dsc_lib/src/configure/mod.rs
@@ -9,7 +9,7 @@ use crate::DscResource;
 use crate::discovery::Discovery;
 use crate::parser::Statement;
 use self::context::Context;
-use self::config_doc::{Configuration, DataType, SecurityContextKind};
+use self::config_doc::{Configuration, DataType, Metadata, SecurityContextKind};
 use self::depends_on::get_resource_invocation_order;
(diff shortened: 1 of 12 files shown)


dsc_lib/src/configure/mod.rs
```diff
@@ -161,6 +162,36 @@ fn add_metadata(kind: &Kind, mut properties: Option<Map<String, Value>> ) -> Res
     Ok(serde_json::to_string(&properties)?)
 }
 
+fn check_security_context(config: &Configuration) -> Result<(), DscError> {
```
tgauth, line 165: is the entire `Configuration` needed by `check_security_context` or can just the `metadata` be passed in here?

SteveL-MSFT: good question.  `Configuration` is passed by reference, so there's no penalty.  It could just require the `metadata` which might make it a bit more clear where the check is happening.  I'll make the change.

tgauth (approved):

Issue #337, opened 2024-03-01
Tracing and progress get interleaved


SteveL-MSFT:
### Prerequisites

- [X] Write a descriptive title.
- [X] Make sure you are able to repro it on the latest version
- [X] Search the existing issues.

### Steps to reproduce

Tracing and progress get rendered on separate threads so not aware of each other and get interleaved.  Should use https://docs.rs/tracing-indicatif/latest/tracing_indicatif/ to sync them

repro:

```
dsc -l trace resource list
```

### Expected behavior

```console
Progress show shift to render most recent as traces get written
```

### Actual behavior

```console
Traces get written while progress tries to write so there's leftover progress state on the screen
```

### Environment data

```PowerShell
NA
```

### Version

main branch

PR #359, opened 2024-03-08
Fix indicatif prog bar

tgauth:
# PR Summary
- readdress #337 
- uses `warn_span` for progress bars since that is the default tracing level and can be turned off by specifying `error` trace level
- no longer displays progress bars once progress is completed, so `done` messages are removed

## PR Context

- `dsc resource list` demo:

https://github.com/PowerShell/DSC/assets/14894321/c98ad0d2-6581-4f88-a593-f3379132317b

- `dsc config get` demo:

https://github.com/PowerShell/DSC/assets/14894321/82948b41-dc58-45f8-aee5-798182ee04d2

### DIFF
--- a/dsc_lib/src/configure/mod.rs
+++ b/dsc_lib/src/configure/mod.rs
@@ -136,7 +136,7 @@ fn escape_property_values(properties: &Map<String, Value>) -> Result<Option<Map<
 }
 
 fn get_progress_bar_span(len: u64) -> Result<Span, DscError> {
-    let pb_span = warn_span!("progress bar");
+    let pb_span = warn_span!("");
     pb_span.pb_set_style(&ProgressStyle::with_template(
         "{spinner:.green} [{elapsed_precise:.cyan}] [{bar:40.cyan/blue}] {pos:>7}/{len:7} {msg:.yellow}"
     )?);
--- a/dsc_lib/src/configure/mod.rs
+++ b/dsc_lib/src/configure/mod.rs
@@ -16,9 +16,9 @@ use self::contraints::{check_length, check_number_limits, check_allowed_values};
 use indicatif::ProgressStyle;
 use security_context_lib::{SecurityContext, get_security_context};
 use serde_json::{Map, Value};
-use tracing_indicatif::span_ext::IndicatifSpanExt;
 use std::{collections::HashMap, mem};
 use tracing::{debug, trace, warn_span, Span};
+use tracing_indicatif::span_ext::IndicatifSpanExt;
 
 pub mod context;
 pub mod config_doc;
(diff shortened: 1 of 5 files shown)


dsc_lib/src/configure/mod.rs
```diff
@@ -135,13 +135,13 @@ fn escape_property_values(properties: &Map<String, Value>) -> Result<Option<Map<
     Ok(Some(result))
 }
 
-fn get_progress_bar(len: u64) -> Result<ProgressBar, DscError> {
-    let pb = ProgressBar::new(len);
-    pb.enable_steady_tick(Duration::from_millis(120));
-    pb.set_style(ProgressStyle::with_template(
+fn get_progress_bar_span(len: u64) -> Result<Span, DscError> {
+    let pb_span = warn_span!("progress bar");
```
SteveL-MSFT, line 139: If we just have this be an empty string `warn_span!("")` would that suppress the unnecessary `progress bar` in the traces?

SteveL-MSFT (changes requested):

dsc_lib/src/configure/mod.rs
```diff
@@ -246,7 +247,8 @@ impl Configurator {
             result.results.push(resource_result);
         }
 
-        pb.finish_with_message("Get configuration completed");
+        mem::drop(pb_span_enter);
```
SteveL-MSFT, line 250: is this needed explicitly?  I would expect both to be dropped when going out of scope

dsc_lib/src/configure/mod.rs
```diff
@@ -135,13 +135,13 @@ fn escape_property_values(properties: &Map<String, Value>) -> Result<Option<Map<
     Ok(Some(result))
 }
 
-fn get_progress_bar(len: u64) -> Result<ProgressBar, DscError> {
-    let pb = ProgressBar::new(len);
-    pb.enable_steady_tick(Duration::from_millis(120));
-    pb.set_style(ProgressStyle::with_template(
+fn get_progress_bar_span(len: u64) -> Result<Span, DscError> {
+    let pb_span = warn_span!("");
```
SteveL-MSFT, line 139: Presumably using a warn span means that if tracing level is error, then progress isn't displayed right?  That seems fine but would need to be documented.

SteveL-MSFT (changes requested):

dsc_lib/src/configure/mod.rs
tgauth: it's needed within the config functions but I removed it from resource list

dsc_lib/src/configure/mod.rs
tgauth: Correct - added a comment. Wasn't sure where to call that out initially, so it was just in the PR description.

SteveL-MSFT (approved):

PR #361, opened 2024-03-08
add min function

tgauth:
# PR Summary
- implement https://learn.microsoft.com/en-us/azure/azure-resource-manager/templates/template-functions-numeric#min

## PR Context
- part of #57

### DIFF
--- a/dsc_lib/src/functions/min.rs
+++ b/dsc_lib/src/functions/min.rs
@@ -27,12 +27,7 @@ impl Function for Min {
         debug!("min function");
         if args.len() == 1 {
             if let Some(array) = args[0].as_array() {
-                if array.len() > 1 {
-                    find_min(array)
-                }
-                else {
-                    Err(DscError::Parser("Array must contain more than 1 integer".to_string()))
-                }
+                find_min(array)
             }
             else {
                 Err(DscError::Parser("Array cannot be empty".to_string()))
@@ -79,14 +74,14 @@ mod tests {
     #[test]
     fn array_single_value() {
         let mut parser = Statement::new().unwrap();
-        let result = parser.parse_and_execute("[min(createArray(0)]", &Context::new());
-        assert!(result.is_err());
+        let result = parser.parse_and_execute("[min(createArray(0)]", &Context::new()).unwrap();
+        assert_eq!(result, 0);
     }
 
     #[test]
     fn arrays() {
         let mut parser = Statement::new().unwrap();
-        let result = parser.parse_and_execute("[min(createArray('0','3'), createArray('2','5'))]", &Context::new());
+        let result = parser.parse_and_execute("[min(createArray(0, 3), createArray(2, 5))]", &Context::new());
         assert!(result.is_err());
     }
 
@@ -102,13 +97,6 @@ mod tests {
         let mut parser = Statement::new().unwrap();
         let result = parser.parse_and_execute("[min(8, min(2, 5), 3)]", &Context::new()).unwrap();
         assert_eq!(result, 2);
-    }
-
-    #[test]
-    fn invalid_one_parameter() {
-        let mut parser = Statement::new().unwrap();
-        let result = parser.parse_and_execute("[min(createArray(1))]", &Context::new());
-        assert!(result.is_err());
     }
 
     #[test]
--- a/dsc_lib/src/functions/min.rs
+++ b/dsc_lib/src/functions/min.rs
@@ -1,0 +1,136 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+use crate::DscError;
+use crate::configure::context::Context;
+use crate::functions::{AcceptedArgKind, Function};
+use serde_json::Value;
+use tracing::debug;
+
+#[derive(Debug, Default)]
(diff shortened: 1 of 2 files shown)


dsc_lib/src/functions/min.rs
```diff
@@ -0,0 +1,136 @@
...
+
+    fn invoke(&self, args: &[Value], _context: &Context) -> Result<Value, DscError> {
+        debug!("min function");
+        if args.len() == 1 {
+            if let Some(array) = args[0].as_array() {
+                if array.len() > 1 {
+                    find_min(array)
+                }
+                else {
+                    Err(DscError::Parser("Array must contain more than 1 integer".to_string()))
```
SteveL-MSFT, line 34: Is this true or would it just return the one integer?

dsc_lib/src/functions/min.rs
```diff
@@ -0,0 +1,136 @@
...
+                }
+            }
+            else {
+                Err(DscError::Parser("List must contain more than 1 integer".to_string()))
```
SteveL-MSFT, line 38: This is probably "Array cannot be empty"

dsc_lib/src/functions/min.rs
```diff
@@ -0,0 +1,136 @@
...
+            }
+        }
+        else {
+            find_min(args)
+        }
+    }
+}
+
+fn find_min(args: &[Value]) -> Result<Value, DscError> {
+    let mut min_value = i64::MAX;
```
SteveL-MSFT, line 48: Looks like you can just use built-in Rust function: https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.min

dsc_lib/src/functions/min.rs
```diff
@@ -0,0 +1,136 @@
...
+    fn min_args(&self) -> usize {
+        1
+    }
+
+    fn max_args(&self) -> usize {
+        usize::MAX
+    }
+
+    fn accepted_arg_types(&self) -> Vec<AcceptedArgKind> {
+        vec![AcceptedArgKind::Number, AcceptedArgKind::Array]
```
SteveL-MSFT, line 23: It looks like the function should only take integers or an array of integers, so you can't mix them.  Should add a check for this.

SteveL-MSFT (changes requested):

dsc_lib/src/functions/min.rs
tgauth: I based this on the description within the documentation since it only refers to "integers" but yes, the code would just return the one integer

dsc_lib/src/functions/min.rs
```diff
@@ -0,0 +1,127 @@
...
+    fn array() {
+        let mut parser = Statement::new().unwrap();
+        let result = parser.parse_and_execute("[min(createArray(0, 3, 2, 5, 4)]", &Context::new()).unwrap();
+        assert_eq!(result, 0);
+    }
+
+    #[test]
+    fn array_single_value() {
+        let mut parser = Statement::new().unwrap();
+        let result = parser.parse_and_execute("[min(createArray(0)]", &Context::new());
```
SteveL-MSFT, line 82: Seems like this should return `0` unless it's explicitly documented to erro?

dsc_lib/src/functions/min.rs
```diff
@@ -0,0 +1,127 @@
...
+        assert!(result.is_err());
+    }
+
+    #[test]
+    fn arrays() {
+        let mut parser = Statement::new().unwrap();
+        let result = parser.parse_and_execute("[min(createArray('0','3'), createArray('2','5'))]", &Context::new());
```
SteveL-MSFT, line 89: Shouldn't these arrays not be quoted?  `createArray(0,3)`?

dsc_lib/src/functions/min.rs
```diff
@@ -0,0 +1,127 @@
...
+    fn nested() {
+        let mut parser = Statement::new().unwrap();
+        let result = parser.parse_and_execute("[min(8, min(2, 5), 3)]", &Context::new()).unwrap();
+        assert_eq!(result, 2);
+    }
+
+    #[test]
+    fn invalid_one_parameter() {
+        let mut parser = Statement::new().unwrap();
+        let result = parser.parse_and_execute("[min(createArray(1))]", &Context::new());
```
SteveL-MSFT, line 110: isn't this the same as above with an array of single number?

dsc_lib/src/functions/min.rs
```diff
@@ -0,0 +1,115 @@
...
+
+#[cfg(test)]
+mod tests {
+    use crate::configure::context::Context;
+    use crate::parser::Statement;
+
+    #[test]
+    fn list() {
+        let mut parser = Statement::new().unwrap();
+        let result = parser.parse_and_execute("[min(3,2,5,4)]", &Context::new()).unwrap();
```
SteveL-MSFT, line 56: I believe the integers are signed, so you should have one negative number in one of the tests to verify that it works correctly

SteveL-MSFT (changes requested):

SteveL-MSFT (approved):

PR #363, opened 2024-03-11
PowerShell adapter PR

mgreenegit:


dsc/tests/dsc_config_get.tests.ps1
```diff
@@ -27,7 +27,8 @@ Describe 'dsc config get tests' {
     It 'will fail if resource schema does not match' -Skip:(!$IsWindows) {
         $jsonPath = Join-Path $PSScriptRoot '../examples/invalid_schema.dsc.yaml'
         $config = Get-Content $jsonPath -Raw
-        $null = $config | dsc config get | ConvertFrom-Json
+        $testError = & {$config | dsc config get get 2>&1}
+        $testError | Select-String '^error:' -Quiet | Should -BeTrue
```
SteveL-MSFT, line 31: If you're just comparing the output, it's better to use something like:

```powershell
$testError | Should -match '^error:'
```

As when it fails, you'll see both the expected and actual vs true/false.

dsc/tests/dsc_get.tests.ps1
```diff
@@ -45,7 +45,8 @@ Describe 'config get tests' {
             "Name": "ProductName"
         }
 '@
-        $json | dsc resource get -r Microsoft.Windows/registry
+        $testError = & {$json | dsc resource get -r Microsoft.Windows/registry get 2>&1}
+        $testError | Select-String '^error:' -Quiet | Should -BeTrue
```
SteveL-MSFT, line 49: Same thing here

dsc/tests/dsc_parameters.tests.ps1
```diff
@@ -81,7 +81,8 @@ Describe 'Parameters tests' {
 "@
         $params_json = @{ parameters = @{ param1 = $value }} | ConvertTo-Json
 
-        $null = $config_yaml | dsc config -p $params_json get
+        $testError = & {$config_yaml | dsc config -p $params_json get 2>&1}
+        $testError | Select-String 'Parameter input failure:' -Quiet | Should -BeTrue
```
SteveL-MSFT, line 85: Same in all these cases

powershell-adapter/Tests/PSTestModule/1.0.0/DscResources/TestPSRepository/TestPSRepository.psm1
```diff
@@ -14,7 +14,7 @@ function Get-TargetResource {
     )
 
     $returnValue = @{
-        Ensure                    = [EnsureEnumeration]::Absent
+        Ensure                    = [EnsureEnumeration].GetEnumName(0)
```
SteveL-MSFT, line 17: This seems like a reduction in readability of the code?

powershell-adapter/Tests/TestClassResource/0.0.1/TestClassResource.psd1
```diff
@@ -0,0 +1,134 @@
+#
```
SteveL-MSFT, line 1: Replace this top part with the copyright notice

powershell-adapter/Tests/TestClassResource/0.0.1/TestClassResource.psd1
```diff
@@ -0,0 +1,134 @@
...
+ModuleVersion = '0.0.1'
+
+# Supported PSEditions
+# CompatiblePSEditions = @()
+
+# ID used to uniquely identify this module
+GUID = 'b267fa32-e77d-48e6-9248-676cc6f2327f'
+
+# Author of this module
+Author = 'migreene'
```
SteveL-MSFT, line 24: This should be Microsoft

powershell-adapter/Tests/TestClassResource/0.0.1/TestClassResource.psd1
```diff
@@ -0,0 +1,134 @@
...
+
+# Company or vendor of this module
+CompanyName = 'Unknown'
```
SteveL-MSFT, line 27: Same

powershell-adapter/Tests/TestClassResource/0.0.1/TestClassResource.psd1
```diff
@@ -0,0 +1,134 @@
...
+
+# Copyright statement for this module
+Copyright = '(c) migreene. All rights reserved.'
```
SteveL-MSFT, line 30: Same

powershell-adapter/Tests/TestClassResource/0.0.1/TestClassResource.psd1
```diff
@@ -0,0 +1,134 @@
...
+
+# Description of the functionality provided by this module
```
SteveL-MSFT, line 32: We should just delete all the generated comment/placeholders that aren't useful

powershell-adapter/Tests/TestClassResource/0.0.1/TestClassResource.psd1
```diff
@@ -0,0 +1,134 @@
...
+# TypesToProcess = @()
+
+# Format files (.ps1xml) to be loaded when importing this module
+# FormatsToProcess = @()
+
+# Modules to import as nested modules of the module specified in RootModule/ModuleToProcess
+# NestedModules = @()
+
+# Functions to export from this module, for best performance, do not use wildcards and do not delete the entry, use an empty array if there are no functions to export.
+FunctionsToExport = '*'
```
SteveL-MSFT, line 72:
```suggestion
FunctionsToExport = @()
```

powershell-adapter/Tests/TestClassResource/0.0.1/TestClassResource.psd1
```diff
@@ -0,0 +1,134 @@
...
+
+# Cmdlets to export from this module, for best performance, do not use wildcards and do not delete the entry, use an empty array if there are no cmdlets to export.
+CmdletsToExport = '*'
```
SteveL-MSFT, line 75:
```suggestion
CmdletsToExport = @()
```

powershell-adapter/Tests/TestClassResource/0.0.1/TestClassResource.psd1
```diff
@@ -0,0 +1,134 @@
...
+
+# Variables to export from this module
+VariablesToExport = '*'
```
SteveL-MSFT, line 78:
```suggestion
VariablesToExport = @()
```

powershell-adapter/Tests/TestClassResource/0.0.1/TestClassResource.psd1
```diff
@@ -0,0 +1,134 @@
...
+
+# Aliases to export from this module, for best performance, do not use wildcards and do not delete the entry, use an empty array if there are no aliases to export.
+AliasesToExport = '*'
```
SteveL-MSFT, line 81:
```suggestion
AliasesToExport = @()
```

powershell-adapter/Tests/powershellgroup.config.tests.ps1
```diff
@@ -5,8 +5,8 @@ Describe 'PowerShell adapter resource tests' {
 
     BeforeAll {
         $OldPSModulePath  = $env:PSModulePath
-        $env:PSModulePath += ";" + $PSScriptRoot
-
+        $env:PSModulePath += [System.IO.Path]::PathSeparator + $PSScriptRoot + [System.IO.Path]::DirectorySeparatorChar
+
```
SteveL-MSFT, line 9:
```suggestion
        $env:PSModulePath += [System.IO.Path]::PathSeparator + $PSScriptRoot
```

Appending the dir separator shouldn't be needed

powershell-adapter/Tests/powershellgroup.resource.tests.ps1
```diff
@@ -5,7 +5,7 @@ Describe 'PowerShell adapter resource tests' {
 
     BeforeAll {
         $OldPSModulePath  = $env:PSModulePath
-        $env:PSModulePath += [System.IO.Path]::PathSeparator + $PSScriptRoot
+        $env:PSModulePath += [System.IO.Path]::PathSeparator + $PSScriptRoot + [System.IO.Path]::DirectorySeparatorChar
```
SteveL-MSFT, line 8:
```suggestion
        $env:PSModulePath += [System.IO.Path]::PathSeparator + $PSScriptRoot
```

Was there a need to add this?

resources/brew/brew.dsc.resource.sh
SteveL-MSFT, line 1: Why is this here?

powershell-adapter/powershell.resource.ps1
```diff
@@ -1,427 +1,349 @@
-# Copyright (c) Microsoft Corporation.
```
SteveL-MSFT, line 1: We need to keep the copyright notice on all our source files

powershell-adapter/powershell.resource.ps1
```diff
@@ -1,427 +1,349 @@
...
 param(
-    [ValidateSet('List','Get','Set','Test','Export','Validate')]
-    $Operation = 'List',
-    [Switch]
-    $WinPS = $false,
-    [Parameter(ValueFromPipeline)]
-    $stdinput
+    [Parameter(Mandatory = $true, Position = 0, HelpMessage = 'Operation to perform. Choose from List, Get, Set, Test, Export, Validate.')]
+    [ValidateSet('List', 'Get', 'Set', 'Test', 'Export', 'Validate')]
+    [string]$Operation = 'Default',
```
SteveL-MSFT, line 5: Why is this `Default` and not one of the actual operations?

powershell-adapter/powershell.resource.ps1
```diff
@@ -1,427 +1,349 @@
...
+    [Parameter(Mandatory = $false, Position = 1, ValueFromPipeline = $true, HelpMessage = 'Configuration or resource input in JSON format.')]
+    [string]$stdinput = '@{}'
```
SteveL-MSFT, line 7: Would be better to call this `$jsonInput` as it's not always being read from STDIN.  Also, may be better to not initialize it to empty JSON and instead check for empty string later.

powershell-adapter/powershell.resource.ps1
```diff
@@ -1,427 +1,349 @@
...
-    $DscResources= Get-DscResource
-
-    foreach ($r in $DscResources)
-    {
-        if ($r.ImplementedAs -eq "Binary")
-        {
+# Cache the results of Get-DscResource to optimize performance
+function Invoke-CacheRefresh {
+    # cache the results of Get-DscResource
+    [resourceCache[]]$resourceCache = @()
```
SteveL-MSFT, line 63: Created https://github.com/PowerShell/DSC/issues/371 to persist cache to disk in the future

powershell-adapter/powershell.resource.ps1
```diff
@@ -1,427 +1,349 @@
...
-        if ($WinPS) {$requiresString = "Microsoft.Windows/WindowsPowerShell"} else {$requiresString = "Microsoft.DSC/PowerShell"}
-
-        $t = [Type]$r.ResourceType
-        $exportMethod = $t.GetMethod('Export')
-
-        $capabilities = @('Get', 'Set', 'Test')
-        if ($null -ne $exportMethod) {
-            $capabilities += 'Export'
+        elseif ($dsc.ParentPath) {
+            $moduleName = Split-Path $dsc.ParentPath | Split-Path | Split-Path -Leaf
```
SteveL-MSFT, line 78: Can you explain why `split-path` is needed 3 times here?

powershell-adapter/powershell.resource.ps1
```diff
@@ -1,427 +1,349 @@
...
+            $DscResourceInfo.Module = $moduleName
+            $DscResourceInfo.ModuleName = $moduleName
+            # workaround: populate module version from psmoduleinfo if available
+            if ($moduleInfo = Get-Module -Name $moduleName -ListAvailable -ErrorAction Ignore) {
```
SteveL-MSFT, line 82: Rather than calling `Get-Module` each time through the loop, should we just cache it once?

powershell-adapter/powershell.resource.ps1
```diff
@@ -1,427 +1,349 @@
...
+$resourceCache = Invoke-CacheRefresh
+
+# Convert the INPUT to a configFormat object so configuration and resource are standardized as moch as possible
+function Get-ConfigObject {
+    param(
+        [Parameter(Mandatory = $true, ValueFromPipeline = $true)]
+        $stdinput
+    )
+    # normalize the INPUT object to an array of configFormat objects
+    $inputObj = $stdInput | ConvertFrom-Json -Depth 10
```
SteveL-MSFT, line 104: Since we are converting FROM json and not TO json, we shouldn't set explicit depth as we may be losing important parts of the input

```suggestion
    $inputObj = $stdInput | ConvertFrom-Json
```

powershell-adapter/powershell.resource.ps1
```diff
@@ -1,427 +1,349 @@
...
+    $desiredState = @()
+
+    # catch potential for improperly formatted configuration input
+    if ($inputObj.resources -and -not $inputObj.metadata.'Microsoft.DSC'.context -eq 'configuration') {
```
SteveL-MSFT, line 108: I don't think this check is needed since you can rely on the metadata to determine it is config otherwise just treat it as resource

powershell-adapter/powershell.resource.ps1
```diff
@@ -1,427 +1,349 @@
...
-                    exit 1
-                }
-                $result += $op_result
-            }
-            else
-            {
-                $errmsg = "Can not find type " + $r.type + "; please ensure that Get-DscResource returns this resource type"
-                Write-Error $errmsg
-                exit 1
+    if ($inputObj.metadata.'Microsoft.DSC'.context -eq 'configuration') {
```
SteveL-MSFT, line 112: It's a bit tedious, but you need to check each nested object isn't $null before you use it otherwise the script will have an exception

```suggestion
    if ($null -ne $inputObject.metadata -and $null -ne $inputObj.metadata.'Microsoft.DSC' -and $inputObj.metadata.'Microsoft.DSC'.context -eq 'configuration') {
```

powershell-adapter/powershell.resource.ps1
```diff
@@ -1,427 +1,349 @@
...
+        switch ([moduleType]$cachedResourceInfo.ImplementationDetail) {
+            'ScriptBased' {
+
+                # If the OS is Windows, import the latest installed PSDesiredStateConfiguration module. For Linux/MacOS, only class based resources are supported and are called directly.
+                if ($IsWindows) {
+                    $DscModule = Get-Module -Name PSDesiredStateConfiguration -ListAvailable | Sort-Object -Property Version -Descending | Select-Object -First 1
+                    Import-Module $DscModule -DisableNameChecking -ErrorAction Ignore
+        
+                    if ($null -eq $DscModule) {
+                        Write-Error 'The PowerShell adapter was called but the module PSDesiredStateConfiguration could not be found in PSModulePath. To install the module, run Install-PSResource -Name PSDesiredStateConfiguration'
```
SteveL-MSFT, line 167:
```suggestion
                        Write-Error 'The PowerShell adapter was called but the module PSDesiredStateConfiguration could not be found in PSModulePath. To install the module, run: Install-PSResource -Name PSDesiredStateConfiguration'
```

powershell-adapter/powershell.resource.ps1
```diff
@@ -1,427 +1,349 @@
...
+        # cache was refreshed on script load
+        foreach ($Type in $resourceCache.Type) {
+        
+            # https://learn.microsoft.com/dotnet/api/system.management.automation.dscresourceinfo
+            $r = $resourceCache | Where-Object Type -EQ $Type | ForEach-Object DscResourceInfo
+
+            # Provide a way for existing resources to specify their capabilities, or default to Get, Set, Test
+            $module = Get-Module -Name $r.ModuleName -ListAvailable | Sort-Object -Property Version -Descending | Select-Object -First 1
+            if ($module.PrivateData.PSData.Capabilities) {
+                $capabilities = $module.PrivateData.PSData.Capabilities
```
SteveL-MSFT, line 264: Perhaps we should call it `DscCapabilities` to make it clear it's only for DSC?

SteveL-MSFT (changes requested):

dsc/tests/dsc_config_get.tests.ps1
mgreenegit: fixed

powershell-adapter/Tests/PSTestModule/1.0.0/DscResources/TestPSRepository/TestPSRepository.psm1
mgreenegit: Otherwise, the index of the enum value is returned

powershell-adapter/Tests/TestClassResource/0.0.1/TestClassResource.psd1
mgreenegit: Can we update this in new-modulemanifest?

powershell-adapter/powershell.resource.ps1
mgreenegit: the user should always explicitly choose. otherwise if something changes in the future, it could be a more severe breaking change.

powershell-adapter/powershell.resource.ps1
mgreenegit: this carried over from previous work, but I believe that is the proper number of steps to get back to the module name folder

resources/brew/brew.dsc.resource.sh
mgreenegit: Git might be picking up on a change to file encoding after cloning to Windows?

powershell-adapter/powershell.resource.ps1
mgreenegit: The intention here is to return helpful feedback if the author forgot (or didn't know to) set the metadata

powershell-adapter/Tests/PSTestModule/1.0.0/DscResources/TestPSRepository/TestPSRepository.psm1
SteveL-MSFT: I think you can use `([EnsureEnumeration]::Absent).ToString()`

powershell-adapter/powershell.resource.ps1
SteveL-MSFT: If `Mandatory=$true`, then the user has to specify.  Maybe just leave it `[string]$Operation` to avoid confusion

powershell-adapter/powershell.resource.ps1
SteveL-MSFT: I missed that this was a warning and not an error.  Seems ok then.

powershell-adapter/powershell.resource.ps1
```diff
@@ -1,427 +1,349 @@
...
-
-    foreach ($r in $DscResources)
-    {
-        if ($r.ImplementedAs -eq "Binary")
-        {
+# Cache the results of Get-DscResource to optimize performance
+function Invoke-CacheRefresh {
+    # cache the results of Get-DscResource
+    [resourceCache[]]$resourceCache = @()
+    $DscResources = Get-DscResource
```
SteveL-MSFT, line 64: The goal should be to not require PSDSC module, but this is still relying on `Get-DscResource` cmdlet.  Perhaps for now we just copy that code out of PSDSC and improve it later?

dsc/tests/dsc_get.tests.ps1
mgreenegit: fixed

powershell-adapter/Tests/TestClassResource/0.0.1/TestClassResource.psd1
mgreenegit: fixed

powershell-adapter/Tests/TestClassResource/0.0.1/TestClassResource.psd1
mgreenegit: fixed

powershell-adapter/Tests/TestClassResource/0.0.1/TestClassResource.psd1
mgreenegit: https://github.com/PowerShell/PSDesiredStateConfiguration/issues/101

powershell-adapter/Tests/TestClassResource/0.0.1/TestClassResource.psd1
mgreenegit: fixed

powershell-adapter/Tests/powershellgroup.config.tests.ps1
mgreenegit: fixed

powershell-adapter/Tests/powershellgroup.resource.tests.ps1
mgreenegit: was part of troubleshooting. not needed.

powershell-adapter/powershell.resource.ps1
mgreenegit: good suggestion. fixed.

powershell-adapter/powershell.resource.ps1
mgreenegit: fixed

powershell-adapter/Tests/PSTestModule/1.0.0/DscResources/TestPSRepository/TestPSRepository.psm1
mgreenegit: fixed

powershell-adapter/Tests/TestClassResource/0.0.1/TestClassResource.psd1
mgreenegit: fixed

powershell-adapter/Tests/TestClassResource/0.0.1/TestClassResource.psd1
mgreenegit: fixed

powershell-adapter/powershell.resource.ps1
mgreenegit: fixed

powershell-adapter/powershell.resource.ps1
mgreenegit: fixed

powershell-adapter/powershell.resource.ps1
mgreenegit: implemented for review

powershell-adapter/powershell.resource.ps1
mgreenegit: fixed

powershell-adapter/powershell.resource.ps1
mgreenegit: fixed

powershell-adapter/powershell.resource.ps1
mgreenegit: fixed

dsc/tests/dsc_config_get.tests.ps1
mgreenegit: tested. unfortunately, Pester -match operation does not seem to work in this pattern.

SteveL-MSFT:
@mgreenegit there's 6 test failures due to the PS adapter, can you take a look and address before I re-review?

build.ps1
```diff
@@ -193,6 +193,9 @@ $skip_test_projects_on_windows = @("tree-sitter-dscexpression")
             }
 
             Copy-Item "*.dsc.resource.json" $target -Force -ErrorAction Ignore
+            
+            $psdscFolder = New-Item -Path $target -Name 'psDscAdapter' -Force -ItemType Directory -ErrorAction Ignore
```
SteveL-MSFT, line 197: Instead of this, you should use a `copy_files.txt` file which just lists the files to be copied to the release folder

dsc/tests/dsc_parameters.tests.ps1
```diff
@@ -108,8 +109,9 @@ Describe 'Parameters tests' {
 "@
         $params_json = @{ parameters = @{ param1 = $value }} | ConvertTo-Json
 
-        $null = $config_yaml | dsc config -p $params_json get
-        $LASTEXITCODE | Should -Be 4
+        $testError = & {$config_yaml | dsc config -p $params_json get get 2>&1}
+        $testError | Select-String -Pattern '^error:' -Quiet | Should -Be $true
```
SteveL-MSFT, line 113: Use the `Should -match` pattern instead

dsc/tests/dsc_parameters.tests.ps1
```diff
@@ -134,8 +136,9 @@ Describe 'Parameters tests' {
...
```
SteveL-MSFT, line 140: Use the `Should -match` pattern instead

dsc/tests/dsc_parameters.tests.ps1
```diff
@@ -158,8 +161,9 @@ Describe 'Parameters tests' {
...
```
SteveL-MSFT, line 165: Use the `Should -match` pattern instead

dsc/tests/dsc_parameters.tests.ps1
```diff
@@ -184,8 +188,9 @@ Describe 'Parameters tests' {
 "@
         $params_json = @{ parameters = @{ param1 = $value }} | ConvertTo-Json
 
-        $null = $config_yaml | dsc config -p $params_json get | ConvertFrom-Json
-        $LASTEXITCODE | Should -Be 4
+        $testError = & {$config_yaml | dsc config -p $params_json get get 2>&1}
+        $testError | Select-String -Pattern '^error:' -Quiet | Should -Be $true
```
SteveL-MSFT, line 192: Use the `Should -match` pattern instead

powershell-adapter/Tests/TestClassResource/0.0.1/TestClassResource.psd1
```diff
@@ -0,0 +1,50 @@
...
+CompanyName = 'Microsoft Corporation'
+
+# Copyright statement for this module
+Copyright = '(c) Microsoft. All rights reserved.'
+
+# Functions to export from this module, for best performance, do not use wildcards and do not delete the entry, use an empty array if there are no functions to export.
+FunctionsToExport = @()
+
+# Cmdlets to export from this module, for best performance, do not use wildcards and do not delete the entry, use an empty array if there are no cmdlets to export.
+CmdletsToExport = '*'
```
SteveL-MSFT, line 28: Are there any cmdlets being exported?  If so, you should list them explicitly instead of using the wildcard

powershell-adapter/Tests/TestClassResource/0.0.1/TestClassResource.psm1
```diff
@@ -43,7 +43,7 @@ class TestClassResource
         {
             $this.Prop1 = $env:DSC_CONFIG_ROOT
         }
-        $this.EnumProp = [EnumPropEnumeration]::Expected
+        $this.EnumProp = [EnumPropEnumeration].GetEnumName(1)
```
SteveL-MSFT, line 46:
```suggestion
        $this.EnumProp = ([EnumPropEnumeration]::Expected).ToString()
```

powershell-adapter/Tests/TestClassResource/0.0.1/TestClassResource.psm1
```diff
@@ -15,7 +15,7 @@ class TestClassResource
     [string] $Prop1
 
     [DscProperty()]
-    [EnumPropEnumeration] $EnumProp
+    [string] $EnumProp
```
SteveL-MSFT, line 18: Why was this changed from the enum type?

powershell-adapter/Tests/native_and_powershell.dsc.yaml
```diff
@@ -21,3 +21,6 @@ resources:
     keyPath: HKLM\Software\Microsoft\Windows NT\CurrentVersion
     valueName: ProductName
     _ensure: Present
+metadata:
```
SteveL-MSFT, line 24: `dsc` will inject this into the config so shouldn't have it explicitly here unless there is some reason?

powershell-adapter/Tests/powershellgroup.config.tests.ps1
```diff
@@ -5,8 +5,8 @@ Describe 'PowerShell adapter resource tests' {
 
     BeforeAll {
         $OldPSModulePath  = $env:PSModulePath
-        $env:PSModulePath += ";" + $PSScriptRoot
-
+        $env:PSModulePath += [System.IO.Path]::PathSeparator + $PSScriptRoot
+
```
SteveL-MSFT, line 9: whitespace

```suggestion

```

powershell-adapter/powershell.resource.ps1
SteveL-MSFT: The diff still shows the copyright header as deleted

powershell-adapter/powershell.resource.ps1
```diff
@@ -1,427 +1,367 @@
...
-
-    $DscResources = Get-DscResource
-
-    foreach ($r in $DscResources)
-    {
-        $moduleName = "";
-        if ($r.ModuleName) { $moduleName = $r.ModuleName }
-        elseif ($r.ParentPath) { $moduleName = Split-Path $r.ParentPath | Split-Path | Split-Path -Leaf }
+# load private functions of psDscAdapter stub module
+Import-Module './psDscAdapter/psDscAdapter.psd1' -Force
```
SteveL-MSFT, line 11:
```suggestion
Import-Module '$PSScriptRoot/psDscAdapter/psDscAdapter.psd1' -Force
```

powershell-adapter/powershell.resource.ps1
SteveL-MSFT: I see.  Please add a comment in the code explaining this.

resources/brew/brew.dsc.resource.sh
SteveL-MSFT, line 1: Why is this here?

SteveL-MSFT (changes requested):

build.ps1
mgreenegit: fixed

powershell-adapter/Tests/TestClassResource/0.0.1/TestClassResource.psd1
mgreenegit: https://github.com/PowerShell/PSDesiredStateConfiguration/issues/101

powershell-adapter/Tests/TestClassResource/0.0.1/TestClassResource.psm1
mgreenegit: fixed

dsc/tests/dsc_parameters.tests.ps1
mgreenegit: fixed

dsc/tests/dsc_parameters.tests.ps1
mgreenegit: fixed

dsc/tests/dsc_parameters.tests.ps1
mgreenegit: fixed

powershell-adapter/Tests/native_and_powershell.dsc.yaml
mgreenegit: didn't know

mgreenegit: it does make testing adapters tricky because you have to just know that this happens. I favor the convenience, but we should make sure this is documented in the future.

powershell-adapter/Tests/powershellgroup.config.tests.ps1
mgreenegit: fixed

powershell-adapter/powershell.resource.ps1
mgreenegit: added

powershell-adapter/powershell.resource.ps1
mgreenegit: fixed

resources/brew/brew.dsc.resource.sh
mgreenegit: has to be the file encoding through vscode?

powershell-adapter/Tests/TestClassResource/0.0.1/TestClassResource.psm1
mgreenegit: something causes the output to return the index instead of the string. I can't seem to reproduce it in PowerShell. It could be something I am doing in Get-ActualState, or the class for output type, but I haven't been able to find it. It could also be related to how the input/output is handled outside of powershell.

resources/brew/brew.dsc.resource.sh
SteveL-MSFT: I guess that makes sense

dsc/tests/dsc_config_get.tests.ps1
```diff
@@ -27,7 +27,8 @@ Describe 'dsc config get tests' {
     It 'will fail if resource schema does not match' -Skip:(!$IsWindows) {
         $jsonPath = Join-Path $PSScriptRoot '../examples/invalid_schema.dsc.yaml'
         $config = Get-Content $jsonPath -Raw
-        $null = $config | dsc config get | ConvertFrom-Json
+        $testError = & {$config | dsc config get get 2>&1}
+        $testError | Select-String '^error:' -Quiet | Should -Be $true
```
SteveL-MSFT, line 31: I thought you had fixed these?  Did you push?

powershell-adapter/copy_files.txt
```diff
@@ -1 +1,8 @@
 powershell.resource.ps1
+\psDscAdapter\psDscAdapter.psd1
```
SteveL-MSFT, line 2: I don't think the leading slash is needed, makes it seem like you are copying from root?

SteveL-MSFT (changes requested):
Almost there

resources/brew/brew.dsc.resource.sh
SteveL-MSFT: @mgreenegit can you just make sure in your branch that file isn't empty (which is what github shows to me)?

powershell-adapter/copy_files.txt
mgreenegit: fixed

SteveL-MSFT (approved):
LGTM, @anmenaga can you review?

mgreenegit:
Just fyi, working out a few changes in advance of starting the Windows adapter
________________________________

mgreenegit:
@anmenaga ready!

build.ps1
```diff
@@ -188,7 +188,14 @@ $skip_test_projects_on_windows = @("tree-sitter-dscexpression")
 
             if (Test-Path "./copy_files.txt") {
                 Get-Content "./copy_files.txt" | ForEach-Object {
-                    Copy-Item $_ $target -Force -ErrorAction Ignore
+                    # copy the file to the target directory, creating the directory path if needed
```
SteveL-MSFT, line 191: throw an error if a backslash is used as separator so that they are all consistently using forward slash

powershell-adapter/windowspowershell.resource.ps1
```diff
@@ -0,0 +1,165 @@
+# Copyright (c) Microsoft Corporation.
```
SteveL-MSFT, line 1: Do we need a separate script?  The previous version I believe had a parameter to know which PS to spawn and I think that was the only functional difference?

SteveL-MSFT (changes requested):

build.ps1
mgreenegit: added

powershell-adapter/windowspowershell.resource.ps1
mgreenegit: not sure yet. even if they start the same, it will be easier to support long term if we seperate them. we might want to consider completely seperating them into two distinct modules/adapters. I can remove them if you prefer?

resources/brew/brew.dsc.resource.sh
mgreenegit: @stevel-msft is there a good way to fix this?

powershell-adapter/windowspowershell.resource.ps1
SteveL-MSFT: It may make sense to separate once they start to diverge, but I think they are the same right now so it may be better to avoid confusion why there's two same scripts to maintain (like when we find bugs)

mgreenegit:
Sounds good to me
________________________________

anmenaga (approved):
LGTM. That is a solid base change for PS Adapter.

PS. As Steve mentions, `brew.dsc.resource.sh` should not be part of this PR.

mgreenegit:
:shipit:

mgreenegit:
HOLD OFF
working on windows powershell support

mgreenegit:
@anmenaga @SteveL-MSFT
added WindowsPowerShell adapter. ready for another review.

mgreenegit:
@SteveL-MSFT @anmenaga 
I don't think we can work around this?
https://github.com/PowerShell/DSC/actions/runs/8608080015/job/23589768568#step:4:1043

SteveL-MSFT:
@anmenaga can you help @mgreenegit figure out the test failures?

anmenaga:
Failing tests involve Binary resources.
I was under impression that we do **not** declare support for those in the fist version of upcoming release of DSC?
If that is the case, then these tests should be removed.

mgreenegit:
The underlying reason is for v1, it needed to run winrm quickconfig. I don’t think we can do that within a build runner?
________________________________

anmenaga:
> The underlying reason is for v1, it needed to run winrm quickconfig. I don’t think we can do that within a build runner?

I don't see a technical problem with that; we can run anything within a build script;
However, the real question is - do we intend to declare support for old binary resources? That will require additional work (and we already have plenty with the need to remove script-based and class resources dependency on PSDesiredStateConfiguration module/PS DSC subsystem.

mgreenegit:
I didn’t know you could perform admin level actions. Thanks @anmenaga. Ready for review.

build.ps1
```diff
@@ -14,6 +14,10 @@ param(
     [switch]$SkipLinkCheck
 )
 
+if ($isWindows) {
+    winrm quickconfig -quiet
```
SteveL-MSFT, line 18: We should do this in the test and not the build script.  Probably in a `BeforeAll {}` block

powershell-adapter/Tests/TestClassResource/0.0.1/TestClassResource.psd1
SteveL-MSFT, line 28:
```suggestion
# this wildcard is required due to https://github.com/PowerShell/PSDesiredStateConfiguration/issues/101
CmdletsToExport = '*'
```

powershell-adapter/Tests/powershellgroup.resource.tests.ps1
```diff
@@ -16,91 +16,121 @@ Describe 'PowerShell adapter resource tests' {
...
+
+        $r = dsc resource list --adapter Microsoft.DSC/WindowsPowerShell
+        $LASTEXITCODE | Should -Be 0
+        $resources = $r | ConvertFrom-Json
+        ($resources | ? {$_.Type -eq 'PSDesiredStateConfiguration/File'}).Count | Should -Be 1
+    }
+
+    It 'Get works on Binary "File" resource' -Skip:(!$IsWindows){
+
+        $testFile = 'c:\test.txt'
```
SteveL-MSFT, line 33:
```suggestion
        $testFile = '$testdrive\test.txt'
```

powershell-adapter/Tests/powershellgroup.resource.tests.ps1
```diff
@@ -16,91 +16,121 @@ Describe 'PowerShell adapter resource tests' {
...
+        'test' | Set-Content -Path $testFile -Force
+        $r = '{"DestinationPath":"' + $testFile.replace('\','\\') + '"}' | dsc resource get -r 'PSDesiredStateConfiguration/File'
+        $LASTEXITCODE | Should -Be 0
+        $res = $r | ConvertFrom-Json
+        $res.actualState.result.properties.DestinationPath | Should -Be "$testFile"
+    }
+
+    It 'Get works on traditional "Script" resource' -Skip:(!$IsWindows){
+
+        $testFile = 'c:\test.txt'
```
SteveL-MSFT, line 43:
```suggestion
        $testFile = '$testdrive\test.txt'
```

powershell-adapter/Tests/winps_resource.dsc.yaml
```diff
@@ -1,11 +1,14 @@
-# Example configuration mixing native app resources with classic PS resources
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
 $schema: https://raw.githubusercontent.com/PowerShell/DSC/main/schemas/2023/10/config/document.json
 resources:
 - name: Get info from classic DSC resources
-  type: Microsoft.Windows/WindowsPowerShell
+  type: Microsoft.DSC/WindowsPowerShell
```
SteveL-MSFT, line 7: Since WindowsPowerShell is part of Windows, we want the typename of this resource to be `Microsoft.Windows/WindowsPowerShell` whereas PowerShell 7 being cross platform would be `Microsoft.DSC/PowerShell`

powershell-adapter/windowspowershell.dsc.resource.json
```diff
@@ -1,8 +1,12 @@
 {
-    "manifestVersion": "1.0",
-    "type": "Microsoft.Windows/WindowsPowerShell",
+    "$schema": "https://raw.githubusercontent.com/PowerShell/DSC/main/schemas/2023/08/bundled/resource/manifest.json",
+    "type": "Microsoft.DSC/WindowsPowerShell",
```
SteveL-MSFT, line 3: As noted above 
```suggestion
    "type": "Microsoft.Windows/WindowsPowerShell",
```

powershell-adapter/~/.config/powershell/Microsoft.PowerShell_profile.ps1
```diff
@@ -0,0 +1 @@
+$env:psmodulepath = $env:psmodulepath.trimend(';')
```
SteveL-MSFT, line 1: Why is this file here? seems accidentally included?

SteveL-MSFT (changes requested):

mgreenegit:
Good point
________________________________

build.ps1
mgreenegit: fixed

powershell-adapter/Tests/TestClassResource/0.0.1/TestClassResource.psd1
mgreenegit: breaks dsc

powershell-adapter/Tests/powershellgroup.resource.tests.ps1
mgreenegit: File resource does not support relative paths

powershell-adapter/Tests/powershellgroup.resource.tests.ps1
mgreenegit: File resource does not support relative paths

powershell-adapter/Tests/winps_resource.dsc.yaml
mgreenegit: fixed

mgreenegit:
Addressed all comments

PR #364, opened 2024-03-14
Enable passing of secure parameters

SteveL-MSFT:
# PR Summary

- Add ability to pass `secureString` and `secureObject` parameter types
- These get passed as JSON (aligned with how you'd define them in an ARM parameters file) to resources where properties match:

```json
{
  "secureString": "string"
}
```

and

```json
{
  "secureObject": { }
}
```

- See the test `Echo` resource as an example
- Enable execution of expressions for parameter default values
- Enable type validation of default values for parameters
- Removed redundant input JSON schema validation since it's already being done in the command resource method invocation


dsc/tests/dsc_parameters.tests.ps1
```diff
@@ -268,4 +268,39 @@ Describe 'Parameters tests' {
...
     }
+
+    It 'secure types can be passed as objects to resources' {
+      $out = dsc config -f $PSScriptRoot/../examples/secure_parameters.parameters.yaml get -p $PSScriptRoot/../examples/secure_parameters.dsc.yaml | ConvertFrom-Json
+      $LASTEXITCODE | Should -Be 0
+      $out.results[0].result.actualState.output | Should -BeExactly 'mySecret'
+      $out.results[1].result.actualState.output | Should -BeExactly 'mySecretProperty'
+    }
+
+    It 'paramter types are validated for <type>' -TestCases @(
```
tgauth, line 279:
```suggestion
    It 'parameter types are validated for <type>' -TestCases @(
```

PR #370, opened 2024-03-23
Update release pipeline

SteveL-MSFT:
# PR Summary

- Move release branch to OneBranch which has automated compliance validation
- Generate msix, msixbundle, and file hashes during release process

There's some minor clean-up work left, but it currently works enough for a preview release.


.pipelines/DSC-Official.yml
```diff
@@ -0,0 +1,283 @@
...
+        - pwsh: |
+            ./build.ps1 -Release -Architecture x86_64-unknown-linux-gnu
+          displayName: 'Build x86_64-unknown-linux-gnu'
+          condition: succeeded()
+        - pwsh: |
+            tar czf '$(ob_outputDirectory)/DSC-$(PackageVersion)-x86_64-unknown-linux-gnu.tar.gz' -C $(Build.SourcesDirectory)/bin/x86_64-unknown-linux-gnu/$(BuildConfiguration) .
+          displayName: 'Compress x86_64-unknown-linux-gnu'
+          condition: succeeded()
+
+      - job: BuildLinuxArm64
```
tgauth, line 195: Not blocking but can the Linux build steps utilize `matrix` like the Mac and Windows builds do?

tgauth (approved):

.pipelines/DSC-Official.yml
SteveL-MSFT: Windows and macOS is able to use matrix because they both can cross compile from a x64 instance arm64 code.  The problem with Linux is that I couldn't get certain OS dependencies like libssl_dev to install both x64 and arm64 so I can cross compile and gave up on it.

PR #376, opened 2024-03-29
add int function

tgauth:
# PR Summary
- implement https://learn.microsoft.com/en-us/azure/azure-resource-manager/templates/template-functions-numeric#int

## PR Context
- part of #57

### DIFF
--- a/dsc_lib/src/functions/int.rs
+++ b/dsc_lib/src/functions/int.rs
@@ -30,7 +30,7 @@ impl Function for Int {
         if arg.is_string() {
             let input = arg.as_str().ok_or(DscError::Function("int".to_string(), "invalid input string".to_string()))?;
             let result = input.parse::<f64>().map_err(|_| DscError::Function("int".to_string(), "unable to parse string to int".to_string()))?;
-            value = NumCast::from(result).ok_or(DscError::Parser("unable to cast to int".to_string()))?;
+            value = NumCast::from(result).ok_or(DscError::Function("int".to_string(), "unable to cast to int".to_string()))?;
         } else if arg.is_number() {
             value = arg.as_i64().ok_or(DscError::Function("int".to_string(), "unable to parse number to int".to_string()))?;
         } else {
--- a/dsc_lib/src/functions/int.rs
+++ b/dsc_lib/src/functions/int.rs
@@ -28,13 +28,13 @@ impl Function for Int {
         let arg = &args[0];
         let value: i64;
         if arg.is_string() {
-            let input = arg.as_str().ok_or(DscError::Function("int".to_string(), "invalid input string".to_string()))?;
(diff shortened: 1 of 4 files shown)


dsc_lib/src/functions/int.rs
```diff
@@ -0,0 +1,73 @@
...
+
+#[cfg(test)]
+mod tests {
+    use crate::configure::context::Context;
+    use crate::parser::Statement;
+
+    #[test]
+    fn string() {
+        let mut parser = Statement::new().unwrap();
+        let result = parser.parse_and_execute("[int('4')]", &Context::new()).unwrap();
```
SteveL-MSFT, line 49: Add a test for `4.0` string

SteveL-MSFT (changes requested):

dsc_lib/src/functions/int.rs
tgauth: Good idea - it's a little more in depth than anticipated because I think the treesitter grammar only expects ints (no `.`). 
Leaning toward refactoring that in a separate PR.

dsc_lib/src/functions/int.rs
```diff
@@ -0,0 +1,73 @@
...
+        let mut parser = Statement::new().unwrap();
+        let result = parser.parse_and_execute("[int(int('-1'))]", &Context::new()).unwrap();
+        assert_eq!(result, -1);
+    }
+
+    #[test]
+    fn error() {
+        let mut parser = Statement::new().unwrap();
+        let result = parser.parse_and_execute("[int('foo')]", &Context::new());
+        assert!(result.is_err());
```
SteveL-MSFT, line 71: instead of `is_err()` which can return any error, you should do something like:

```rust
assert!(matches!(result, Err(<specific error>)))
```

SteveL-MSFT (changes requested):

dsc_lib/src/functions/int.rs
SteveL-MSFT: as a separate PR, we should fix the other unit tests that simply check for `is_err()`

dsc_lib/src/functions/int.rs
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    #[test]
+    fn error() {
+        let mut parser = Statement::new().unwrap();
+        let err = parser.parse_and_execute("[int('foo')]", &Context::new()).unwrap_err();
+        match err {
+            DscError::Function(func, msg) => {
+                assert_eq!(func, "int".to_string());
+                assert_eq!(msg, "unable to parse string to int".to_string());
```
SteveL-MSFT, line 84: Eventually, we may need to localize error messages so comparing the text isn't a good thing.  Instead, perhaps we should add a more specific error type like `DscError:FunctionArg(func, msg)` and check for that and not the specifics of the error content.

SteveL-MSFT (approved):

PR #377, opened 2024-03-29
Added adapter filter

### LINKED ISSUE #274: Need a way to selectively disable individual resources/providers

### LINKED ISSUE #368: Adapters in resource path get evaluated twice
### Prerequisites

### Steps to reproduce

dsc resource list

### Expected behavior

```console
each adapter is only processed once
```

### Actual behavior

```console
each adapter is processed twice
```

### Error details

```console
This happens because we are adding the path to dsc.exe to paths, even if it is already there
https://github.com/mgreenegit/DSC/blob/4bb4b01e0eae26836c95b9fa339a32dec33fad85/dsc_lib/src/discovery/command_discovery.rs#L77-L78
```

### Environment data

```PowerShell
7
```

### Version

latest

### Visuals

_No response_

anmenaga:
# PR Summary

Fix #274 
Fix #368 

1) Removed duplicates from path list, so that same resources would not be processed multiple times;
2) re-enabled WMI provider 
3) Added filter for adapters, so that in default scenarios 'List' operation on adapters is not executed, therefore reducing time.
```
Usage: dsc.exe resource list [OPTIONS] [RESOURCE_NAME]

Arguments:
  [RESOURCE_NAME]  Optional filter to apply to the list of resources

Options:
  -a, --adapter <ADAPTER_NAME>     Adapter filter to limit the resource search
  -d, --description <DESCRIPTION>  Description keyword to search for in the resource description
  -t, --tags <TAGS>                Tag to search for in the resource tags
  -f, --format <FORMAT>            The output format to use [possible values: json, pretty-json, yaml]
  -h, --help                       Print help
```
Examples:

1. `dsc resource list` same as `dsc resource list *` - show all native resources, adapters, but not adapter-based resources (fastest);
2. `dsc resource list *Microsoft*` - same as previous but also apply `*Microsoft*` filter;
3. `dsc resource list * -a *` - show all adapter-based resources;
4. `dsc resource list * -a *PowerShell*` - show all resources of adapters that match `*PowerShell*` filter
5. `dsc resource list *MSFT_PackageManagement* *PowerShell*` - same as previous but also apply `*MSFT_PackageManagement*` filter to resource types;


dsc_lib/src/discovery/command_discovery.rs
```diff
@@ -2,18 +2,22 @@
...
+use crate::discovery::convert_wildcard_to_regex;
 use crate::dscresources::dscresource::{Capability, DscResource, ImplementedAs};
 use crate::dscresources::resource_manifest::{import_manifest, Kind, ResourceManifest};
 use crate::dscresources::command_resource::invoke_command;
 use crate::dscresources::command_resource::log_resource_traces;
 use crate::dscerror::DscError;
 use indicatif::ProgressStyle;
+use regex::{RegexBuilder, Regex};
 use std::collections::BTreeMap;
+use std::collections::HashSet;
```
SteveL-MSFT, line 14:
```suggestion
use std::collections::{BTreeMap, HashSer};
```

dsc_lib/src/discovery/command_discovery.rs
```diff
@@ -2,18 +2,22 @@
...
 use std::env;
 use std::ffi::OsStr;
 use std::fs::File;
 use std::io::BufReader;
 use std::path::Path;
+use std::path::PathBuf;
```
SteveL-MSFT, line 20:
```suggestion
use std::path::{Path, PathBuf};
```

dsc_lib/src/discovery/command_discovery.rs
```diff
@@ -78,74 +62,241 @@ impl CommandDiscovery {
...
+        Self::new()
+    }
+}
+
+impl ResourceDiscovery for CommandDiscovery {
+
+    #[allow(clippy::too_many_lines)]
+    fn list_available_resources(&mut self, type_name_filter: &str, adapter_name_filter: &str) -> Result<BTreeMap<String, DscResource>, DscError> {
+
+        debug!("Listing resources with type_name_filter/adapter_name_filter: {:?}/{:?}", type_name_filter, adapter_name_filter);
```
SteveL-MSFT, line 84: I think you can just have the variables in the formatted string

```suggestion
        debug!("Listing resources with type_name_filter/adapter_name_filter: {type_name_filter}/{adapter_name_filter}");
```

dsc_lib/src/discovery/command_discovery.rs
```diff
@@ -78,74 +62,241 @@ impl CommandDiscovery {
...
+        )?);
+        pb_span.pb_set_message("Searching for resources");
+        let _ = pb_span.enter();
+
+        let mut resources: BTreeMap<String, DscResource> = BTreeMap::new();
+        let mut adapter_resources: BTreeMap<String, DscResource> = BTreeMap::new();
+
+        let regex_str = convert_wildcard_to_regex(type_name_filter);
+        debug!("Using regex {regex_str} as filter for resource type");
+        let mut regex_builder = RegexBuilder::new(regex_str.as_str());
```
SteveL-MSFT, line 98: For consistency, prefer we use the reference syntax than `as_str()`

```suggestion
        let mut regex_builder = RegexBuilder::new(&regex_str);
```

dsc_lib/src/discovery/command_discovery.rs
```diff
@@ -78,74 +62,241 @@ impl CommandDiscovery {
...
+        regex_builder.case_insensitive(true);
+        let type_regex: Regex;
+        if let Ok(reg_v) = regex_builder.build() {
+            type_regex = reg_v;
+        } else {
+            let err_str = "Could not build Regex filter for resource type";
+            error!(err_str);
+            return Err(DscError::Operation(err_str.to_string()));
+        }
```
SteveL-MSFT, line 107: I think this could be simplified as

```suggestion
        let Ok(type_regex) = regex_builder.build() else {
            let err_str = "Could not build Regex filter for resource type";
            error!(err_str);
            return Err(DscError::Operation(err_str.to_string()));
        };
```

dsc_lib/src/discovery/command_discovery.rs
```diff
@@ -78,74 +62,241 @@ impl CommandDiscovery {
...
+
+        if let Ok(paths) = CommandDiscovery::get_resource_paths() {
+            for path in paths {
+                debug!("Searching in {:?}", path);
+                if path.exists() && path.is_dir() {
+                    for entry in path.read_dir().unwrap() {
```
SteveL-MSFT, line 113: I'd prefer to not have any `unwrap()` calls as they are potential for panics.  If they return a `Result`, use `?` and add to DscError for conversion as needed.  If it's an `Option`, then return an `Err` when `None`

dsc_lib/src/discovery/command_discovery.rs
```diff
@@ -78,74 +62,241 @@ impl CommandDiscovery {
...
+        let type_regex: Regex;
+        if let Ok(reg_v) = regex_builder.build() {
+            type_regex = reg_v;
+        } else {
+            let err_str = "Could not build Regex filter for resource type";
+            error!(err_str);
+            return Err(DscError::Operation(err_str.to_string()));
+        }
+
+        if let Ok(paths) = CommandDiscovery::get_resource_paths() {
```
SteveL-MSFT, line 109: This is too deeply nested.  Can we split out some of it into helper functions and remove the line 81?

dsc_lib/src/discovery/command_discovery.rs
```diff
@@ -78,74 +62,241 @@ impl CommandDiscovery {
...
-                                           At this point we can't determine whether or not the bad manifest contains resource that is requested by resource/config operation
-                                           if it is, then "ResouceNotFound" error will be issued later
-                                           and here we just record the error into debug stream.*/
-                                        debug!("{}", e);
+                                        continue;
+                                    },
+                                };
+
+                                if resource.manifest.is_some() {
+                                    let manifest = import_manifest(resource.manifest.clone().unwrap())?;
```
SteveL-MSFT, line 133:
```suggestion
                                if Some(manifest) = resource.manifest {
                                    let manifest = import_manifest(manifest)?;
```

dsc_lib/src/discovery/command_discovery.rs
```diff
@@ -78,74 +62,241 @@ impl CommandDiscovery {
...
+            let mut regex_builder = RegexBuilder::new(regex_str.as_str());
+            regex_builder.case_insensitive(true);
+            let adapter_regex: Regex;
+            if let Ok(reg_v) = regex_builder.build() {
+                adapter_regex = reg_v;
+            } else {
+                let err_str = "Could not build Regex filter for adapter name";
+                error!(err_str);
+                return Err(DscError::Operation(err_str.to_string()));
+            }
```
SteveL-MSFT, line 162:
```suggestion
            let Ok(adapter_regex) = regex_builder.build() else {
                let err_str = "Could not build Regex filter for adapter name";
                error!(err_str);
                return Err(DscError::Operation(err_str.to_string()));
            };
```

dsc_lib/src/discovery/command_discovery.rs
```diff
@@ -78,74 +62,241 @@ impl CommandDiscovery {
...
+
+            // now go through the adapter resources and add them to the list of resources
+            for adapter in adapter_resources {
+                if adapter_regex.is_match(&adapter.1.type_name) {
+                    debug!("Enumerating resources for adapter {}", adapter.1.type_name);
+                    let pb_adapter_span = warn_span!("");
+                    pb_adapter_span.pb_set_style(&ProgressStyle::with_template(
+                        "{spinner:.green} [{elapsed_precise:.cyan}] {msg:.white}"
+                    )?);
+                    pb_adapter_span.pb_set_message(format!("Enumerating resources for adapter {}", adapter.1.type_name).as_str());
```
SteveL-MSFT, line 172:
```suggestion
                    pb_adapter_span.pb_set_message(format!("Enumerating resources for adapter '{}'", adapter.1.type_name).as_str());
```

dsc_lib/src/discovery/command_discovery.rs
```diff
@@ -78,74 +62,241 @@ impl CommandDiscovery {
...
+                    let _ = pb_adapter_span.enter();
+                    let adapter_resource = adapter.1;
+                    let adapter_type_name = adapter_resource.type_name.clone();
+                    let manifest = import_manifest(adapter_resource.manifest.clone().unwrap())?;
```
SteveL-MSFT, line 176: Again, rethink all of your uses of `unwrap()`.  This can be:

```suggestion
                    let manifest = if let Some(manifest) = adapter_resource.manifest {
                      import_manifest(manifest)
                    } else {
                      return Err(DscError::MissingManifest(adapter_resource.type_name.clone()));
                    };
```

dsc_lib/src/discovery/command_discovery.rs
```diff
@@ -78,74 +62,241 @@ impl CommandDiscovery {
...
+            "{spinner:.green} [{elapsed_precise:.cyan}] [{bar:40.cyan/blue}] {pos:>7}/{len:7} {msg:.yellow}"
+        )?);
+        pb_span.pb_set_message("Searching for resources");
+        let _ = pb_span.enter();
+
+        let mut resources: BTreeMap<String, DscResource> = BTreeMap::new();
+        let mut adapter_resources: BTreeMap<String, DscResource> = BTreeMap::new();
+        let mut remaining_required_resource_types = required_resource_types.to_owned();
+
+        if let Ok(paths) = CommandDiscovery::get_resource_paths() {
```
SteveL-MSFT, line 238: Seems like there's some redundancy with `list_availble_resource()` above?  Should refactor and have one depend on the other

SteveL-MSFT (changes requested):
Seems like we need to makes sure we have test cases covering all your bullet points

dsc_lib/src/discovery/command_discovery.rs
anmenaga: Fixed.

dsc_lib/src/discovery/command_discovery.rs
anmenaga: Fixed.

dsc_lib/src/discovery/command_discovery.rs
anmenaga: Updated.

dsc_lib/src/discovery/command_discovery.rs
anmenaga: Updated

dsc_lib/src/discovery/command_discovery.rs
anmenaga: This is actually old code; the diff is misleading.
We have 258 `unwrap()` calls throughout the code; I've opened #378 to address those.

dsc_lib/src/discovery/command_discovery.rs
anmenaga: Updated.

dsc_lib/src/discovery/command_discovery.rs
anmenaga: I don't think this is a problem in this case - a simple file search.
Plus, I'm not sure how this can be done shorter.

dsc_lib/src/discovery/command_discovery.rs
anmenaga: updated.

dsc_lib/src/discovery/command_discovery.rs
anmenaga: updated.

dsc_lib/src/discovery/command_discovery.rs
anmenaga: updated.

anmenaga:
Added tests.

SteveL-MSFT:
After trying out your branch, we should discuss some user experience options:

- Currently, to use an adapter, you have to specify `*` or some resource name.  I would prefer something like `dsc resource list --adapter Microsoft.DSC/PowerShell` and have it show all the PowerShell resources
- When an adapter is specified, should the list ONLY contain resources for that adapter.  I would think this would be expected unless you use a wildcard for the adapter, of course.

cc @mgreenegit, @michaeltlombardi

anmenaga:
My thoughts:
1.
> * Currently, to use an adapter, you have to specify `*` or some resource name.  I would prefer something like `dsc resource list --adapter Microsoft.DSC/PowerShell` and have it show all the PowerShell resources

It is shorter to type `*` for type filter than `--adapter` of the named parameter. I'd leave it as is.

2.
> * When an adapter is specified, should the list ONLY contain resources for that adapter.  I would think this would be expected unless you use a wildcard for the adapter, of course.

I agree with this; it is more intuitive than current behavior.

SteveL-MSFT:
> My thoughts: 1.
> 
> > * Currently, to use an adapter, you have to specify `*` or some resource name.  I would prefer something like `dsc resource list --adapter Microsoft.DSC/PowerShell` and have it show all the PowerShell resources
> 
> It is shorter to type `*` for type filter than `--adapter` of the named parameter. I'd leave it as is.
> 

We would have a `-a` short name for `--adapter`.  Particularly on Unix where you have to single-quote the wildcard so that it doesn't get globbed, my preference is to have adapter be a named arg and not require `*`.

anmenaga:
> * Currently, to use an adapter, you have to specify `*` or some resource name.  I would prefer something like `dsc resource list --adapter Microsoft.DSC/PowerShell` and have it show all the PowerShell resources
> * When an adapter is specified, should the list ONLY contain resources for that adapter.  I would think this would be expected unless you use a wildcard for the adapter, of course.

Updated code and PR description for both points.

dsc/tests/dsc_args.tests.ps1
```diff
@@ -180,9 +180,9 @@ resources:
         $LASTEXITCODE | Should -Be 0
     }
 
-    It 'resource tracing shows up' {
+    It 'resource tracing shows up' -Skip:(!$IsWindows) {
```
SteveL-MSFT, line 183: Maybe (as a separate PR) we should have our dsctest resource emit output to stderr and test against that in addition to the ps-adapter

dsc/tests/dsc_args.tests.ps1
```diff
@@ -215,4 +215,60 @@ resources:
...
+            $a_obj.type | Should -Be $b_obj.type
+            # adapter-based resources should Not be in the results
+            $a_obj.requireAdapter | Should -BeNullOrEmpty
+            $b_obj.requireAdapter | Should -BeNullOrEmpty
+        }
+    }
+
+    It 'verify `dsc resource list resource_filter`' {
+        # same as previous but also apply resource_filter filter
+        $a = dsc resource list Test* -f json
```
SteveL-MSFT, line 237: Similar to `*`, should put in single quotes so the shell doesn't try to interpret it and `dsc` receives it

```suggestion
        $a = dsc resource list 'Test*' -f json
```

dsc_lib/src/discovery/mod.rs
```diff
@@ -6,9 +6,8 @@ mod discovery_trait;
 
 use crate::discovery::discovery_trait::ResourceDiscovery;
 use crate::{dscresources::dscresource::DscResource, dscerror::DscError};
-use regex::{RegexBuilder, Regex};
 use std::collections::BTreeMap;
-use tracing::{debug, error};
+use tracing::{error};
```
SteveL-MSFT, line 10: Surprised clippy didn't complain about unnecessary braces

```suggestion
use tracing::error;
```

tools/test_group_resource/tests/provider.tests.ps1
```diff
@@ -5,7 +5,7 @@ Describe 'Resource adapter tests' {
 
     It 'Can list adapter resources' {
 
-        $out = dsc resource list *testresource* | ConvertFrom-Json | Sort-Object -Property type
+        $out = dsc resource list *testresource* -a Test/TestGroup | ConvertFrom-Json | Sort-Object -Property type
```
SteveL-MSFT, line 8:
```suggestion
        $out = dsc resource list '*testresource*' -a Test/TestGroup | ConvertFrom-Json | Sort-Object -Property type
```

tools/test_group_resource/tests/provider.tests.ps1
```diff
@@ -55,7 +56,7 @@ Describe 'Resource adapter tests' {
             Set-Content -Path testdrive:/invalid.dsc.resource.json -Value $invalid_manifest
             $env:PATH += [System.IO.Path]::PathSeparator + (Resolve-Path (Resolve-Path $TestDrive -Relative))
 
-            $out = dsc resource list *invalid* 2>&1
+            $out = dsc resource list *invalid* -a *InvalidTestGroup* 2>&1
```
SteveL-MSFT, line 59:
```suggestion
            $out = dsc resource list '*invalid*' -a '*InvalidTestGroup*' 2>&1
```

SteveL-MSFT (changes requested):

dsc/tests/dsc_args.tests.ps1
anmenaga: Yes, that would make tests cross-platform.

SteveL-MSFT (approved):

Issue #283, opened 2023-11-21
Microsoft.Windows/Registry fails to open key in alpha 4 release


msftrubengu:
### Prerequisites

- [X] Write a descriptive title.
- [X] Make sure you are able to repro it on the latest version
- [X] Search the existing issues.

### Steps to reproduce

Download v3.0.0-alpha.4 x86_64 windows.
Use Microsoft.Windows/Registry or config file with the resource

Using the resources built from v3.0.0-alpha.4 works, so I'm assuming this is an issue with the binaries in the release.

### Expected behavior

```console
PS D:\Dev\DSC> dsc --input-file d:\dev\DscV3\configFiles\resources.yml config test
  2023-11-21T19:53:27.008710Z  INFO dsc: Reading input from file d:\dev\DscV3\configFiles\resources.yml
    at src\main.rs:53

  2023-11-21T19:53:29.617538Z  INFO dsc_lib::dscresources::command_resource: Invoking test Microsoft.Windows/Registry using registry
    at D:\Dev\DSC\dsc_lib\src\dscresources\command_resource.rs:232

  2023-11-21T19:53:29.662573Z  INFO dsc_lib::dscresources::command_resource: Invoking test Microsoft.Windows/Registry using registry
    at D:\Dev\DSC\dsc_lib\src\dscresources\command_resource.rs:232

results:
- name: test 1
  type: Microsoft.Windows/Registry
  result:
    desiredState:
      keyPath: HKCU\DscTest
      _exist: true
    actualState:
      $id: https://developer.microsoft.com/json-schemas/windows/registry/20230303/Microsoft.Windows.Registry.schema.json      keyPath: HKCU\DscTest
      _exist: false
      _inDesiredState: false
    inDesiredState: false
    differingProperties:
    - _exist
- name: test 2
  type: Microsoft.Windows/Registry
  result:
    desiredState:
      keyPath: HKCU\DscTest2
      _exist: true
    actualState:
      $id: https://developer.microsoft.com/json-schemas/windows/registry/20230303/Microsoft.Windows.Registry.schema.json      keyPath: HKCU\DscTest2
      _exist: false
      _inDesiredState: false
    inDesiredState: false
    differingProperties:
    - _exist
messages: []
hadErrors: false
```

### Actual behavior

```console
D:\Dev>dsc --input-file DscV3\configFiles\resources.yml config test
  2023-11-21T18:47:20.393816Z  INFO dsc: Reading input from file DscV3\configFiles\resources.yml
[...]

Issue #290, opened 2023-12-26
Expose `delete` as a direct dsc resource operation


SteveL-MSFT:
### Summary of the new feature / enhancement

It may be simpler for DSC resource implementations to implement `delete` rather than have logic as part of `set` to handle `_exist`.  We can still support `_exist` as a special keyword in configuration where `dsc` knows to call `delete` if implemented.

PowerShell resources already implement `ensure` differently, so that won't change either way.

michaeltlombardi:
I'm not sure about this - it does move a little bit away from the underlying state management model of an instance + properties - if you have `delete`, why not `create` (don't update if exist) or `modify` (don't create if not exist)?

It seems like the sort of thing I might expect a higher order tool to implement, rather than a resource or DSC itself.

SteveL-MSFT:
@michaeltlombardi the idea is that this can simplify resource authoring as you can have clear separation of `set` (CreateOrUpdate) and `delete` whereas, for example, in the registry resource, I basically have a bunch of `if` statements to check if `_exist` is passed in and it would simplify the code by having `delete` a separate code path.  For end users, they can call `delete` directly via `dsc resource delete ...`, but if using config, it would be abstracted so deletes can occur with create/update within the same config.

SteveL-MSFT:
We should allow resources to implement `_exist` directly, but if the resource implements `delete` then `dsc` will use that to handle `_exist` rather than passing that to the resource.

SteveL-MSFT:
Discussed with @mgreenegit and we agreed to move forward with this proposal

PR #379, opened 2024-04-02
Rewrite registry resource to support delete operation

SteveL-MSFT:
# PR Summary

- Moved old `registry` resource to archive folder and excluded from build (no need to review, files moved without changes)
- Wrote new `registry` resource using Registry crate simplifying the code, important behaviors to doc for resources:
  - When performing a `get` to a resource that doesn't exist, don't error.  Instead return keys and `_exist = false`
  - When performing a `delete` to a resource that already doesn't exist, don't error.  Instead return keys and `_exist = false`
- Add `delete` as valid `dsc resource` operation
- New property for `set` method `handlesExist` indicating that the resource will handle `_exist` directly so `dsc` will pass it as-is to the resource with test resource to validate.  The engine work to change `_exist` to `delete` will come later after this is merged.
- Updated table writer for `dsc resource list` to show `SetHandlesExist` and `Delete` capabilities
- Updated existing tests using `registry` resource as it takes JSON via argument instead of stdin

## PR Context

Fix https://github.com/PowerShell/DSC/issues/283
Partially address https://github.com/PowerShell/DSC/issues/290


build.ps1
```diff
@@ -117,7 +117,7 @@ if (!$SkipBuild) {
     New-Item -ItemType Directory $target > $null
 
 # make sure dependencies are built first so clippy runs correctly
-$windows_projects = @("pal", "ntreg", "ntstatuserror", "ntuserinfo", "registry", "reboot_pending", "wmi-adapter")
```
anmenaga, line 120: Why "ntstatuserror" and "ntuserinfo" are removed?

anmenaga (approved):
LGTM.

What was causing #283 ?

build.ps1
SteveL-MSFT: Those libs were used by the old registry resource for handling NT API errors (translating to human readable errors) and for getting the current user for HKCU (it's not a real hive).  None of these are needed as the new crate uses win32 apis that abstract all of this.

SteveL-MSFT:
#283 was probably due to the old resource using libraries that called the NT APIs using lots of unsafe code that wasn't written correctly.

PR #382, opened 2024-04-04
Update engine to call `delete` for resources that don't support `_exist` directly

SteveL-MSFT:
# PR Summary

- A configuration set operation will now call `delete` on a resource (if it supports it) if `_exist = false` is one of the properties and the resource doesn't have the `SetHandlesExist` capability.
- If the resource has the `SetHandlesExist` capability, then it will get the raw properties
- Updated `Test/Exist` resource to better validate case that `_exist` is passed through
- Added `Test/Delete` resource to validate `delete` gets called
- Group/Adapter resources are not supported for the `delete` path as all resources would need to support `delete` (you'll get an error message if the resource returns a group response).
- Since `delete` is a synthetic `set` with `_exist`, had to write code to convert the `get` response into a `set` response
- Added synthetic `test` during configuration `set` operation if the resource doesn't directly implement it

## PR Context

Fix https://github.com/PowerShell/DSC/issues/290


anmenaga (approved):
LGTM

Issue #218, opened 2023-10-04
Refactor `input: arg` for improved UX/Schema


michaeltlombardi:
### Summary of the new feature / enhancement

In #213 we added functionality for defining an input kind that passes the instance property bag JSON to an executable by replacing a templated argument in the list:

https://github.com/PowerShell/DSC/blob/5b92b584f9fd8a1919c0bafd75058584948a1d95/dsc_lib/src/dscresources/resource_manifest.rs#L51-L62

I'm not sure about this change/implementation from the user-facing perspective and the schema - it moves the schema from a set of enum values (in JSON Schema syntax) to a `oneOf` that is more complicated and harder for users to adjust/grok.

Compare the three available input kinds for the `get` command:

```json
"get": {
  "executable": "registry",
  "args": ["config", "get"],
  "input": "env"
}
```

```json
"get": {
  "executable": "registry",
  "args": ["config", "get"],
  "input": "stdin"
}
```

```json
"get": {
  "executable": "registry",
  "args": ["config", "get", "--input", "{{JSON}}"],
  "input": { "arg": "{{JSON}}" }
}
```

The latter option requires the author to:

1. Decide on a string to replace with interpolation.
2. Define `input` as an object with the property `arg` set to that string value instead of as a string.

This both complicates the JSON schema and adds a (small) complexity of choice where I'm not sure it _fully_ matters. It also makes supporting full IntelliSense and contextual help during the authoring process more difficult, as we can't rely on the `markdownEnumDescriptions` metadata in the same way.

### Proposed technical implementation details (optional)

I would propose an enum value like `Arg` (`arg` in the JSON) that _always_ replaces a static string argument in the list, instead of allowing an arbitrary string. If that static string argument isn't included, we could either error (the manifest is invalid) or add the JSON as the last argument, after all defined arguments.

We could have that argument string be something like `{{instance_json}}` to be very unlikely to conflict with any actual arguments.

Then the manifest would look like:

```jsonc
// explicit argument
"set": {
  "executable": "registry",
  "args": ["config", "set", "--input", "{{instance_json}}"],
  "input": "arg"
}
```
```jsonc
// implicit argument, if supported
"set": {
  "executable": "registry",
  "args": ["config", "set"],
  "input": "arg"
}
```

From a schema perspective, we can then validate that when `input` is `arg` that `args` includes the static string and not when `input` is any other value.

Here's a sample of the schema definition, assuming no implicit handling:

```yaml
allOf:
[...]

PR #385, opened 2024-04-05
Refactor arguments to command resources an mixed strings and objects for extensibility

SteveL-MSFT:
# PR Summary

- args are now always processed regardless if how `input` is specified, so it's technically possible to have the input passed as both an arg and stdin/env
- args now take a string or an object where the object currently supported looks like:

```json
{
  "jsonInputArg": "<parameter>",
  "mandatory": true
}
```

The value of `jsonInputArg` is what the input gets passed to, so `"jsonInputArg": "--foo"` would be passed as `command --foo $json` effectively.

The `mandatory` indicates whether to pass the JSON even if empty.  Default is false so it skips providing this parameter completely, but if true, it will always provide the parameter and an empty string if input is empty.

## PR Context

Fix https://github.com/PowerShell/DSC/issues/218


dsc_lib/src/dscresources/resource_manifest.rs
```diff
@@ -80,11 +80,23 @@ pub enum ManifestSchemaUri {
...
+        /// Indicates if argument is mandatory which will pass an empty string if no JSON input is provided.  Default is false.
+        mandatory: Option<bool>,
+    }
+}
+
 #[derive(Debug, Clone, PartialEq, Deserialize, Serialize, JsonSchema)]
 pub enum InputKind {
-    /// The input replaces arguments with this token in the command.
-    #[serde(rename = "arg")]
-    Arg(String),
```
michaeltlombardi, line 87: Rather than deleting this option entirely, I think it makes sense to allow resource authors to explicitly define the input kind as `arg` to preclude being sent the property bag as environment variables or over stdin and accounting for double processing.

If we do this, we could also update the JSON Schema (and code) to infer the input kind from the manifest - if the args array includes an object and `input` isn't defined, `input` is `arg`. If the args array is only made of strings and `input` isn't defined, `input` is `stdin`. The `input` is only ever `env` if it's explicitly defined.

We could then also use JSON schema to validate the `args` array - if it's explicitly defined as `env` or `stdin`, is must _not_ include an object, only strings. If it's not explicitly defined or is defined as `arg`, it may only include one object in the array.

We could use this to provide better author-time feedback and validation for the resource manifest and direct authors towards best practices.

dsc_lib/src/dscresources/command_resource.rs
```diff
@@ -146,24 +143,21 @@ pub fn invoke_set(resource: &ResourceManifest, cwd: &str, desired: &str, skip_te
 
     let mut get_env: Option<HashMap<String, String>> = None;
     let mut get_input: Option<&str> = None;
-    let mut get_args = resource.get.args.clone();
+    let args = process_args(&resource.get.args, desired);
     match &resource.get.input {
```
tgauth, line 147: would it be simpler to add this `match` block to `process_args` or as its own function since it's repeated within the other commands?

dsc_lib/src/dscresources/resource_manifest.rs
michaeltlombardi: Quick json schema example (a little pseudo code, bc it's a little complex to write initially but functional/useful):

```yaml
# set property in manifest
input:
  type: string
  enum: [arg, env, stdin]
args:
  type: array
  contains: { type: object } # check if array includes objects
  minContains: 0 # allow string-only
  maxContains: 1 # allow json arg only once
  items:
    anyOf:
      - type: string
      - type: object # validate object structure
        required: [jsonInputArg]
        properties:
          jsonInputArg:
            type: string
          mandatory:
            type: boolean
            default: false
# Helper for validation/authoring
allOf:
  - if: # args items are only strings
    then:
      properties:
        input: { default: stdin }
  - if: # args items includes one object
    then:
      properties:
        input: { default: arg }
  - if: # input is arg (default or explicit)
      properties: { input: { const: arg } }
    then: # require object arg exactly once
      properties:
        args: { minContains: 1 }
    else: # forbid object arg entirely
      properties:
        args: { maxContains: 0 }
```

dsc/tests/dsc_discovery.tests.ps1
```diff
@@ -77,13 +77,8 @@ Describe 'tests for resource discovery' {
...
                 "executable": "dsctest",
                 "args": [
-                    "echo",
-                    "--input",
-                    "{json}"
-                ],
-                "input": {
-                    "arg": "{json}"
-                }
+                    "echo"
```
anmenaga, line 80: Why it's not matching the
```
{
  "jsonInputArg": "<parameter>",
  "mandatory": true
}
```
syntax?

dsc_lib/src/dscresources/command_resource.rs
```diff
@@ -634,24 +669,30 @@ pub fn invoke_command(executable: &str, args: Option<Vec<String>>, input: Option
     Ok((exit_code, stdout, stderr))
 }
 
-fn replace_token(args: &mut Option<Vec<String>>, token: &str, value: &str) -> Result<(), DscError> {
+fn process_args(args: &Option<Vec<ArgKind>>, value: &str) -> Option<Vec<String>> {
     let Some(arg_values) = args else {
-        return Err(DscError::Operation("No args to replace".to_string()));
+        debug!("No args to process");
```
anmenaga, line 674: Kind of too generic message.

anmenaga (approved):

dsc_lib/src/dscresources/resource_manifest.rs
SteveL-MSFT: The resource author has to opt into any acceptance of the JSON whether it's env, stdin, or args so there isn't a case where it's accidentally sent.  However, there might be a case where the resource might want the simplified reception of the JSON as env, but also keep a raw copy as an arg to pass to something else.

dsc/tests/dsc_discovery.tests.ps1
SteveL-MSFT: This is just a test resource that doesn't get run to validate the semver.  Maybe I'll rename the type to make this clear and remove the args since they aren't used at all.

dsc_lib/src/dscresources/command_resource.rs
SteveL-MSFT: This function is only for processing args, but you're right there's some redundancy.  Let me pull out the common part as a function.

dsc_lib/src/dscresources/command_resource.rs
SteveL-MSFT: Wanted a debug message that may be useful in the future to know that there weren't any args

Issue #390, opened 2024-04-09
The `int()` function incorrectly handles non-integer numbers


michaeltlombardi:
### Prerequisites

- [X] Write a descriptive title.
- [X] Make sure you are able to repro it on the latest version
- [X] Search the existing issues.

### Steps to reproduce

When using the `int()` function, I expect it to return the same value for a fractional number as the string representation of that fractional number. However, instead of returning `4` for `4.7`, it returns `7`.

### Expected behavior

```powershell
@'
$schema: https://raw.githubusercontent.com/PowerShell/DSC/main/schemas/2023/10/config/document.json
resources:
- name: Echo integer value
  type: Test/Echo
  properties:
    output:
      fromString: "[int('4.7')]"
      fromNumber: "[int(4.7)]"
'@ | dsc config get
```
```yaml
results:
- name: Echo integer value
  type: Test/Echo
  result:
    actualState:
      output:
        fromString: 4
        fromNumber: 4
messages: []
hadErrors: false
```

### Actual behavior

```powershell
@'
$schema: https://raw.githubusercontent.com/PowerShell/DSC/main/schemas/2023/10/config/document.json
resources:
- name: Echo integer value
  type: Test/Echo
  properties:
    output:
      fromString: "[int('4.7')]"
      fromNumber: "[int(4.7)]"
'@ | dsc config get
```
```yaml
results:
- name: Echo integer value
  type: Test/Echo
  result:
    actualState:
      output:
        fromString: 4
        fromNumber: 7
messages: []
hadErrors: false
```

### Environment data

```PowerShell
Name                           Value
----                           -----
PSVersion                      7.4.1
PSEdition                      Core
GitCommitId                    7.4.1
OS                             Microsoft Windows 10.0.22631
Platform                       Win32NT
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0…}
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
WSManStackVersion              3.0
```

### Version

Latest build from `main`

tgauth:
The root cause of this issue is within the tree-sitter-dscexpression grammar, so this will apply to other functions as well. For numbers, it's only looking for `int` based on https://learn.microsoft.com/en-US/azure/azure-resource-manager/templates/data-types.

michaeltlombardi:
[...]

PR #391, opened 2024-04-09
(DOCS) Document new and changed config functions

michaeltlombardi:
# PR Summary

This change adds reference documentation and changelog entries for the new and updated configuration functions in DSC.

It also:

- Defines a few helper snippets for docs authoring, scoped to the Markdown files in the repository.
- Adds maintainer comments to the changelog for using the snippets and for updating the changelog, particularly around releases.

## PR Context

Prepare for next release.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,29 @@ ms.date: 03/06/2024
 
 <!-- markdownlint-disable-file MD033 -->
 
+<!--
+    Helpful docs snippets
+
+    You can use the following snippets in this file to make authoring new change entries and
+    releases easier.
+
+    docs-changelog-entry-single-issue-pr: Adds a new changelog list entry with one related issue
+    and PR. Use this when a change entry only has one related issue and/or pull request. Always
+    use this snippet under a change kind heading, like 'Added' or 'Changed.'
+
+    docs-changelog-entry-multi-issue-pr: Adds a new changelog list entry with sub-lists for issues
+    and PRs. Use this when a change entry has more than one related issue and/or pull request.
+    Always use this snippet under a change kind heading, like 'Added' or 'Changed.'
+
+    docs-changelog-release-heading: Adds a new changelog release heading, following our existing
+    format. Use this when a new release is created to ensure that the new release heading has the
+    required links and synopsis.
+
+    docs-gh-link: Adds a new link to an issue or pull request on GitHub. Use this when adding a new
+    work item link reference and it will automatically construct the URL and reference link ID for
+    you from the work item ID.
+-->
+
 All notable changes to DSCv3 are documented in this file. The format is based on
 [Keep a Changelog][m1], and DSCv3 adheres to [Semantic Versioning][m2].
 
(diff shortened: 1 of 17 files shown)


docs/reference/schemas/config/functions/int.md
```diff
@@ -0,0 +1,87 @@
...
+ms.date:     04/09/2024
+ms.topic:    reference
+title:       int
+---
+
+# int
+
+## Synopsis
+
+Returns an integer from an input string or non-integer number.
```
tgauth, line 12: I'm not sure if `int()` needs to be capable of converting from a _non-integer number_, as https://learn.microsoft.com/en-us/azure/azure-resource-manager/templates/template-functions-numeric#int specifies the input as `string` or `int` (...though I can't really think of a use case for calling `int()` with an `int`)

michaeltlombardi: I think there's some value in getting return data from a resource like `1.3` and converting it to `1`, but I don't fully disagree with you.

It's also worth noting that, right now, converting a number with a fractional part to an integer with this configuration function is broken - see #390 for more details.

CHANGELOG.md
```diff
@@ -21,10 +44,89 @@ All notable changes to DSCv3 are documented in this file. The format is based on
...
+
+### Added
+
+- Added configuration functions:
+
+  - New mathematics functions include [add()][ur.01], [div()][ur.02], [max()][ur.03],
+    [min()][ur.04], [mod()][ur.05], [mul()][ur.06], and [sub()][ur.07]. The mathematics functions
+    only operate on integer values.
+
+  - The [reference()][ur.08] function enables you to reference the result output for other
```
SteveL-MSFT, line 94: Might be worth calling out that you can only reference a resource that has already executed previously, so order is important.  You'll get an error if you try to reference a resource that hasn't executed yet or doesn't exist in the configuration.

michaeltlombardi: Updated to clarify recommendation for adding the referenced instance to the `dependsOn` property when using `reference()`.

CHANGELOG.md
```diff
@@ -124,11 +231,12 @@ in this release, see the [diff on GitHub][compare-v3.0.0-alpha.5].
...
   </details>
 
-- Added the [DSCConfigRoot][42] environment variable and the [envvar() configuration function][43]
-  to enable users to reference files and folders relative to the folder containing the
-  configuration document. DSC automatically creates the `DSCConfigRoot` environment variable when
-  you use the `--path` option to specify the configuration document instead of passing the document
-  as a string from stdin or with the `--document` option.
+- Added the [DSCConfigRoot][a5.11] environment variable and the
+  [envvar() configuration function][a5.12] to enable users to reference files and folders relative
+  to the folder containing the configuration document. DSC automatically creates the
```
SteveL-MSFT, line 236:
```suggestion
  to the folder containing the configuration document. DSC automatically and only creates the
```

SteveL-MSFT (changes requested):

docs/reference/schemas/config/functions/int.md
SteveL-MSFT: @michaeltlombardi can you open an issue to track this?

SteveL-MSFT (approved):

PR #393, opened 2024-04-09
(MAINT) Update GitHub templates

michaeltlombardi:
# PR Summary

This change copies the default templates for PowerShell repositories and customizes them for DSC.

In particular, the bug report template is updated to more usefully reflect errors and issues with DSC, rather than a PowerShell command.

The change includes the feature request template and template config, because adding any files to the `.github/ISSUE_TEMPLATE` directory requires all of them to be present - none of the defaults in the community repository are used when this folder exists.

## PR Context

The existing bug report doesn't apply the correct labels and can be confusing to fill out, as it seems more specific to PowerShell than to DSC.

### DIFF
--- a/.github/ISSUE_TEMPLATE/Bug_Report.yml
+++ b/.github/ISSUE_TEMPLATE/Bug_Report.yml
@@ -1,0 +1,117 @@
+name: Bug report 🐛
+description: Report errors or unexpected behavior 🤔
+labels:
+  - Issue-Bug
+  - Need-Review
+body:
+- type: checkboxes
+  attributes:
+    label: Prerequisites
+    options:
+    - label: Write a descriptive title.
+      required: true
+    - label: Make sure you are able to repro it on the latest version
+      required: true
+    - label: Search the existing issues.
+      required: true
+- type: textarea
+  attributes:
+    label: Summary
+    description: >-
+      Write a short description of the issue at a high-level.
+    placeholder: >-
+      I am experiencing a problem with X.
+      I think Y should be happening but Z is actually happening.
+  validations:
+    required: true
+- type: textarea
+  attributes:
+    label: Steps to reproduce
+    description: >
+      List of steps, sample code, failing test or link to a project that reproduces the behavior.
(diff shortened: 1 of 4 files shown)


.github/ISSUE_TEMPLATE/Bug_Report.yml
```diff
@@ -0,0 +1,122 @@
...
+    required: true
+- type: dropdown
+  validations:
+    required: true
+  attributes:
+    label: Version
+    description: Specify the version you are using.
+    options:
+      - Latest build from `main`
+      - alpha.5
```
SteveL-MSFT, line 108: have them run `dsc --version` instead

.github/ISSUE_TEMPLATE/Bug_Report.yml
```diff
@@ -0,0 +1,124 @@
...
+      messages: []
+      hadErrors: false
+  validations:
+    required: true
+- type: textarea
+  attributes:
+    label: Actual behavior
+    render: console
+    placeholder: |
+      PS> dsc config get --path ./repro.dsc.config.yaml
```
SteveL-MSFT, line 75: We should ask to have `debug` tracing enabled as it shouldn't leak any private data:

```powershell
# enable debug tracing which should not contain any private data
dsc -l debug config get ...`
```

SteveL-MSFT (changes requested):

.github/ISSUE_TEMPLATE/Bug_Report.yml
```diff
@@ -0,0 +1,126 @@
...
+- type: dropdown
+  validations:
+    required: true
+  attributes:
+    label: Version
+    description: >-
+      Specify the version you're using. Run the `dsc --version` command to get
+      the current version.
+    options:
+      - Latest build from `main`
```
SteveL-MSFT, line 110: If we still use `options` isn't that a fixed list that we have to keep updating?

SteveL-MSFT (changes requested):

SteveL-MSFT (approved):

PR #402, opened 2024-04-12
Example resource for working with existing InTune custom compliance scripts

mgreenegit:
To gather feedback and understand if this would be of value, or not


SteveL-MSFT:
Perhaps we should put this under a `demo` folder indicating it's an example not not something we're going to support in production?  The `brew` resource would be another one I'd put in this bucket

mgreenegit:
Smart
________________________________

Issue #407, opened 2024-04-15
Extend exitCodes to support Hexidecimal error codes


joshcorr:
### Summary of the new feature / enhancement

While I was playing around an idea for a DSC Resource, I received Exit code `-2147024891` (`0x80070005`) Access Denied when trying to execute a binary which required elevation. I tried to represent this error in the `exitCodes` section of a DSC Resource manifest, but it does not appear to match the regex that dsc is using to validated `exitCodes`.

The feature request is that the regex be updated to support negative/Hexidecimal exit codes.

michaeltlombardi:
You're correct, the JSON Schema forbids non-integers for the property names in the definition of `exitCodes`:

https://github.com/PowerShell/DSC/blob/7404ad81bb57da3195c16f8748b0e2c73b7a4184/schemas/src/resource/manifest.yaml#L383-L395

The implementation for DSC expects the exit codes to be an integer:

https://github.com/PowerShell/DSC/blob/7404ad81bb57da3195c16f8748b0e2c73b7a4184/dsc_lib/src/dscresources/resource_manifest.rs#L57-L59

You can add the following repro manifest to your path to see how DSC behaves:

```json
{
    "$schema": "https://raw.githubusercontent.com/PowerShell/DSC/main/schemas/2023/10/bundled/resource/manifest.json",
    "type": "DSC.Repro/Example",
    "version": "0.1.0",
    "description": "Repro behavior when exit codes are hexadecimal.",
    "get": {
        "executable": "repro",
        "input": "stdin"
    },
    "schema": {
        "command": {
            "executable": "repro",
            "args": [
                "--schema"
            ]
        }
    },
    "exitCodes": {
        "0": "Success",
        "1": "Error",
        "-2147024891": "testing hexadecimal as integer"
    }
}
```

```powershell
dsc resource list DSC.Repro/Example

dsc resource list DSC.Repro/Example
| ConvertFrom-Json
| Select-Object -ExpandProperty manifest
| Select-Object -ExpandProperty exitCodes
| ConvertTo-Json
```

```console
Type               Kind      Version  Caps    RequireAdapter  Description
-------------------------------------------------------------------------------------------------------------
DSC.Repro/Example  Resource  0.1.0    g-----                  Repro behavior when exit codes are hexadecimal.

{
  "0": "Success",
  "1": "Error",
  "-2147024891": "testing hexadecimal as integer"
}
```

However, if you update the manifest to use hex format for the exit code:

```jsonc
{
[...]

PR #410, opened 2024-04-17
(GH-407) Support negative integers in manifest `exitCodes`

michaeltlombardi:
# PR Summary

This change:

- Updates both the source and composed schemas to allow negative integers.
- Updates the reference documentation to clarify that you must specify integers, that you can't use any alternate formats for those integers, and that the keys in YAML must be wrapped in quotes for parsing.
- Addresses #407 by making the schema compliant with the implementation.

## PR Context

Prior to this change, the JSON schema for the `exitCodes` property of resource manifests only supported positive integers as exit codes.

DSC itself supports negative integers as exit codes, and some apps return negative integers as exit codes for hexadecimal exit codes, like `-2147024891` for `0x80070005`, "Access denied."

> [!NOTE]
> This PR builds on #397 and is blocked on that PR being reviewed and merged.

### DIFF
--- a/docs/reference/schemas/resource/manifest/root.md
+++ b/docs/reference/schemas/resource/manifest/root.md
@@ -229,10 +229,26 @@ information, see the [DSC Resource manifest provider property schema reference][
 The `exitCodes` property defines a set of valid exit codes for the resource and their meaning.
 Define this property as a set of key-value pairs where:
 
-- The key is a string containing an integer that maps to a known exit code for the resource.
+- The key is a string containing a signed integer that maps to a known exit code for the resource.
+  The exit code must be a literal signed integer. You can't use alternate formats for the exit
+  code. For example, instead of the hexadecimal value `0x80070005` for "Access denied", specify the
+  exit code as `-2147024891`.
 - The value is a string describing the semantic meaning of that exit code for a human reader.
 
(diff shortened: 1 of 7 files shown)


docs/reference/schemas/resource/manifest/root.md
```diff
@@ -229,11 +229,26 @@ information, see the [DSC Resource manifest provider property schema reference][
 The `exitCodes` property defines a set of valid exit codes for the resource and their meaning.
 Define this property as a set of key-value pairs where:
 
-- The key is a string containing an integer that maps to a known exit code for the resource.
+- The key is a string containing an integer that maps to a known exit code for the resource. You
```
SteveL-MSFT, line 232: Should we just say this is a signed int?

SteveL-MSFT (dismissed):
One suggestion to make it a bit more clear in the manifest docs

docs/reference/schemas/resource/manifest/root.md
SteveL-MSFT: Don't see this change yet?

michaeltlombardi: The change is here for me:

https://github.com/PowerShell/DSC/blob/916548c74fcef82ec3f3e84cf917d969f9d826ad/docs/reference/schemas/resource/manifest/root.md?plain=1#L232-L235

PR #422, opened 2024-04-24
Refactor resource discovery code

SteveL-MSFT:
# PR Summary

In preparation to add support for handle multiple and finding specific versions of resource, needed to refactor the existing code.

- New helper to simply find all command based resources
- Another helper to enumerate adapter resources
- All the discovery is cached in memory
- Rename `discover_resources()` to be `find_resources()` as that's really what it does based on a filter
- Functions now return a vector of resources of the same type name, but different versions

Because the changes moves around lots of existing code, might be easier to review the file rather than the diff.

### DIFF
--- a/dsc_lib/src/discovery/command_discovery.rs
+++ b/dsc_lib/src/discovery/command_discovery.rs
@@ -138,7 +138,7 @@ impl ResourceDiscovery for CommandDiscovery {
                                         // resource that is requested by resource/config operation
                                         // if it is, then "ResouceNotFound" error will be issued later
                                         // and here we just record the error into debug stream.
-                                        debug!("{e}");
+                                        warn!("{e}");
                                         continue;
                                     },
                                 };
--- a/dsc_lib/src/discovery/command_discovery.rs
+++ b/dsc_lib/src/discovery/command_discovery.rs
@@ -143,7 +143,7 @@ impl ResourceDiscovery for CommandDiscovery {
                                     },
                                 };
 
-                                if regex.is_match(filter) {
+                                if regex.is_match(&resource.type_name) {
                                     if let Some(ref manifest) = resource.manifest {
                                         let manifest = import_manifest(manifest.clone())?;
                                         if manifest.kind == Some(Kind::Adapter) {
--- a/dsc_lib/src/discovery/command_discovery.rs
+++ b/dsc_lib/src/discovery/command_discovery.rs
@@ -203,7 +203,7 @@ impl ResourceDiscovery for CommandDiscovery {
 
         for (adapter_name, adapters) in &self.adapters {
             for adapter in adapters {
-                if !regex.is_match(&adapter_name) {
+                if !regex.is_match(adapter_name) {
                     continue;
                 }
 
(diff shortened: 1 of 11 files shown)


dsc_lib/src/discovery/discovery_trait.rs
```diff
@@ -5,6 +5,8 @@ use crate::{dscresources::dscresource::DscResource, dscerror::DscError};
 use std::collections::BTreeMap;
 
 pub trait ResourceDiscovery {
```
anmenaga, line 7: To be clear on intent of the original 2 functions (it would be good to add this as a comment on these functions):

`list_available_resources` - used only in 'List' operation - it just returns the list of found classes (i.e. no caching);
`discover_resources` - used in all other operations - it searches for and caches requested classes in [Discovery.resources](https://github.com/PowerShell/DSC/blob/9ee037ae9a4491d1c8a932a3224c3a281a313cc0/dsc_lib/src/discovery/mod.rs#L13) so that later on (at the invocation time) class definition can be retrieved by simple/fast [`self.resources.get`](https://github.com/PowerShell/DSC/blob/9ee037ae9a4491d1c8a932a3224c3a281a313cc0/dsc_lib/src/discovery/mod.rs#L58).

SteveL-MSFT: I'll add a comment.  These two functions still exist, but they both now use helpers to remove the duplicated code

dsc_lib/src/discovery/command_discovery.rs
```diff
@@ -6,25 +6,32 @@ use crate::discovery::convert_wildcard_to_regex;
...
 use std::ffi::OsStr;
 use std::fs::File;
 use std::io::BufReader;
 use std::path::{Path, PathBuf};
-use tracing::{debug, error, trace, warn, warn_span, Span};
+use tracing::{debug, info, trace, warn, warn_span};
 use tracing_indicatif::span_ext::IndicatifSpanExt;
 
 pub struct CommandDiscovery {
+    // use BTreeMap so that the results are sorted
```
anmenaga, line 23: please extend the comment to something like :
... so that results are sorted by reousrce type name and vector values - are found resource versions of this type in order of decreasing of semversion.

anmenaga (approved):

dsc_lib/src/discovery/mod.rs
```diff
@@ -45,8 +53,10 @@ impl Discovery {
                 }
             };
 
-            for resource in discovered_resources {
-                    resources.push(resource.1);
+            for (_resource_name, found_resources) in discovered_resources {
+                for resource in found_resources {
+                    resources.push(resource.clone());
```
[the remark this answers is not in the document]
SteveL-MSFT: Correct, that's just in preparation for follow-up PR, didn't want to change `find_resource` yet

dsc_lib/src/discovery/discovery_trait.rs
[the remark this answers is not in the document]
SteveL-MSFT: We should do that separately and also remove the abstraction above command_resource since we only have that one

dsc_lib/src/discovery/command_discovery.rs
SteveL-MSFT: fixed

Issue #412, opened 2024-04-18
`IncludeGroup` to include an external configuration


SteveL-MSFT:
### Summary of the new feature / enhancement

Not sure yet if this is a good idea as I can see the main downside that you don't get a complete view of the configuration in one doc, but there might be use cases where you define separate config docs for network, storage, compute (as examples) and you want a higher level doc that applies all of them.  So it could look like:

```yaml
resources:
- name: networking
  type: Microsoft.DSC/Include
  properties:
    configurationFile: ./networking.yaml
- name: storage
  type: Microsoft.DSC/Include
  properties:
    configurationFile: ./storage.yaml
```

acangialosi:
One use case for this is configuring development environments like dev box where there are a common set of configurations used across multiple workloads a team or an individual developer has defined. Dev Teams often standardize some environment specific configurations like enabling dev mode, showing hidden files, enabling hyper-v, or installing VS Code across multiple project types. 
Then they have workload or project specific environment requirements. One project focused on intelligent Apps needs a specific version of python installed and the Python extension installed. 
Another project is a .net project that installs .NET Core and the C# DevKit extension for VS Code. 
Being able to DependsOn across files also useful. 

```yaml
common-dev-env.yaml
    - resource: Microsoft.Windows.Developer/DeveloperMode
    - resource: Microsoft.Windows.Developer/WindowsExplorer
      directives:
        allowPrerelease: true
      settings:
        FileExtensions: Show
        HiddenFiles: Show
    - resource: Microsoft.WinGet.DSC/WinGetPackage
      id: vscode
      directives:
        allowPrerelease: true
      settings:
        id: Microsoft.VisualStudioCode
``` 
```yaml
intelligent-app.yaml
    resources:
    - name: commondevenv
      type: Microsoft.DSC/Include
      properties:
        configurationFile: ./common-dev-env.yaml
    - resource: Microsoft.WinGet.DSC/WinGetPackage
      id: python311
      directives:
        allowPrerelease: true
      settings:
        id: Python.Python.3.11
    - resource: PSDscResources/Script
      id: VSCodemspython
      dependsOn:
        - python311
        - vscode
      settings:
        SetScript: |
          code --install-extension ms-python.python
        GetScript: |
          return $false
        TestScript: |
          return $false
``` 
```yaml
dotnet-app.yaml
    resources:
    - name: commondevenv
      type: Microsoft.DSC/Include
      properties:
        configurationFile: ./common-dev-env.yaml
[...]

PR #429, opened 2024-05-03
Add `Include` resource via new `Import` resource kind and `resolve` operation

SteveL-MSFT:
# PR Summary

Note that this solution is not quite complete yet as it only works for `get`.  However, through this work, I've realized I need to refactor resource `get`, `set`, `test` responses so it is always a list even if just one element.  The current separation of a singleton vs list makes it too complicated to complete the work for `set` and `test`.

- New `Microsoft.DSC/Include` resource which includes a file for configuration and also parameters
- The example `include.dsc.yaml` shows usage
- Needed to add new `DSC_TRACE_LEVEL` env var so that child `dsc` processes will emit the same level of traces as I needed it to diagnose problems
  - if `--trace-level` is specified, it will replace any existing `DSC_TRACE_LEVEL` env var with the new value, `dsc` now uses this env var unless `--trace-level` is specified
  - Renamed `warning` level to `warn` to match the code to keep it simple
- New `Import` resource kind that needs to support a `resolve` operation and return a `ResolveResult`
  - As part of adding `resolve`, I've also made `get` optional since it doesn't make sense for an `Import` resource
  - After the files have been resolved to JSON, then a new instance of `configurator` is created to perform the operation against the resolved content
- Needed to change how `find_resources()` works by internally handling case-insensitivity for searching.  This removed a bunch of code that made things lowercase before calling this function. 
- helper `parse_input_to_json()` that will accept JSON or YAML and emit JSON since the file pointed to by `Include` may either be JSON or YAML for the configuration and parameters file

## PR Context

Fix https://github.com/PowerShell/DSC/issues/412


dsc/src/resolve.rs
```diff
@@ -0,0 +1,152 @@
...
+                    return Err(format!("Error: Failed to read file '{include_path:?}': {err}"));
+                }
+            }
+        },
+        Err(err) => {
+            return Err(format!("Error: Failed to open included file '{include_path:?}': {err}"));
+        }
+    }
+    // convert the buffer to a string
+    let include_content = match String::from_utf8(buffer) {
```
anmenaga, line 71: This reminds me that we need to put somewhere in the docs, that input documents for dsc.exe have to be in UTF-8.
This is important for non-English users.

dsc/src/resolve.rs
```diff
@@ -0,0 +1,152 @@
...
+                    Ok(current_directory) => current_directory.to_string_lossy().into_owned(),
+                    Err(err) => {
+                        return Err(format!("Error: Failed to get current directory: {err}"));
+                    }
+                }
+            }
+        };
+
+        // combine the current directory with the Include path
+        Ok(Path::new(&current_directory).join(path))
```
anmenaga, line 150: What if the path in the include file is a fully-specified one?

dsc/src/args.rs
```diff
@@ -33,8 +33,8 @@ pub struct Args {
     /// The subcommand to run
     #[clap(subcommand)]
     pub subcommand: SubCommand,
-    #[clap(short = 'l', long, help = "Trace level to use", value_enum, default_value = "warning")]
-    pub trace_level: TraceLevel,
+    #[clap(short = 'l', long, help = "Trace level to use", value_enum)]
+    pub trace_level: Option<TraceLevel>,
```
anmenaga, line 37: please add a comment that default value will be assigned in the code as "warn".

dsc/tests/dsc_config_get.tests.ps1
```diff
@@ -37,15 +37,15 @@ Describe 'dsc config get tests' {
             `$schema: https://raw.githubusercontent.com/PowerShell/DSC/main/schemas/2024/04/config/document.json
             resources:
             - name: Echo
-              type: Test/Echo
+              type: test/echo
```
anmenaga, line 40: Why change to lower case?

dsc_lib/src/dscresources/command_resource.rs
```diff
@@ -706,3 +747,37 @@ fn json_to_hashmap(json: &str) -> Result<HashMap<String, String>, DscError> {
...
+/// * `process_id` - The ID of the process
+/// * `stderr` - The stderr output from the process
+pub fn log_resource_traces(process_name: &str, process_id: &u32, stderr: &str)
+{
+    if !stderr.is_empty()
+    {
+        for trace_line in stderr.lines() {
+            if let Result::Ok(json_obj) = serde_json::from_str::<Value>(trace_line) {
+                if let Some(msg) = json_obj.get("Error") {
+                    error!("Process '{process_name}' id {process_id} : {}", msg.as_str().unwrap_or_default());
```
anmenaga, line 765: It would be good to make `"Process '{process_name}' id {process_id}` part shorter; with it trace lines are too long.
Maybe `[<processname>:<procId>]: <text>`

anmenaga (approved):
LGTM

dsc/src/resolve.rs
SteveL-MSFT: cc @michaeltlombardi

dsc/src/resolve.rs
SteveL-MSFT: This is already handled on line 127 above

dsc/tests/dsc_config_get.tests.ps1
SteveL-MSFT: this was to explicitly test case-insensitivity

dsc_lib/src/dscresources/command_resource.rs
SteveL-MSFT: We can do that in a follow-up PR along with handling the `tracing` crate JSON format

PR #433, opened 2024-05-08
(DOCS) Update reference and changelog for preview.7

michaeltlombardi:
# PR Summary

These changes update the reference documentation for the CLI and schemas as well as the changelog for the updates in the `preview.7` release.

## PR Context

Document and announce changes for the latest release.


CHANGELOG.md
```diff
@@ -81,25 +164,178 @@ changes since the last release, see the [diff on GitHub][unreleased].
...
+    - [#382][#382]
+
+  </details>
+
+- <a id="add-elevation-requirement" /></a> Added the option to specify whether a configuration
+  document requires root or elevated permissions. Now, you can define the `securityContext`
+  metadata property under the `Microsoft.DSC` namespace in a configuration document to specify
+  which security context to use:
+
+  - `Current` - Any security context.
```
SteveL-MSFT, line 268: Is it worth saying this is the default if not specified?

CHANGELOG.md
```diff
@@ -117,20 +353,204 @@ changes since the last release, see the [diff on GitHub][unreleased].
...
+  a pending reboot. It can only be used for assertions, not to enforce state.
+
+  <details><summary>Related work items</summary>
+
+  - Issues: [#333][#333]
+  - PRs: [#344][#344]
+
+  </details>
+
+- Added the `Microsoft.DSC.Transitional/RunCommandOnSet` resource, which runs a specified
```
SteveL-MSFT, line 508: I know we don't want to promote use of this resource, but it might not be clear to novice users that this can be used to run a PowerShell script.

docs/reference/cli/resource/delete.md
```diff
@@ -0,0 +1,158 @@
...
+
+## Description
+
+The `delete` subcommand removes a resource instance.
+
+Any properties the resource requires for discerning which instance to delete must be passed to this
+command as a JSON or YAML object. The object can be passed to this command from stdin or with the
+`--input` option. You can also use the `--path` option to read the object from a JSON or YAML file.
+
+This command returns no output when successful. If it encounters an error, it surfaces the error to
```
SteveL-MSFT, line 48: I believe `dsc` also returns a non-zero exit code if `delete` fails as relying on STDERR isn't sufficient.

docs/reference/schemas/config/metadata.md
```diff
@@ -0,0 +1,80 @@
...
+```yaml
+SchemaDialect: https://json-schema.org/draft/2020-12/schema
+SchemaID:      https://raw.githubusercontent.com/PowerShell/DSC/main/schemas/2023/10/config/document.metadata.json
+Type:          object
+```
+
+## Description
+
+Defines a set of informational key-value pairs for the configuration. Except for the
+`Microsoft.DSC` property, this metadata isn't validated. You can pass any data into your configuration as a property of `metadata`.
```
SteveL-MSFT, line 25: Should we have stronger language that `Microsoft.DSC` is reserved to only be used by DSC itself?

docs/reference/schemas/definitions/resourceKind.md
```diff
@@ -0,0 +1,358 @@
...
+        # deeply nested OSInfo.
+        - name: Deeply nested echo
+          type: Test/Echo
+          properties:
+            output:  >-
+              [reference(
+                resourceId('Microsoft/OSInfo', 'Deeply nested OSInfo')
+              ).actualState]
+          dependsOn:
+            - "[resourceId('Microsoft/OSInfo', 'Deeply nested OSInfo')]"
```
SteveL-MSFT, line 291: would it be useful to have an example of depending on more than one resource?  LIke maybe a group and an individual resource?

SteveL-MSFT (changes requested):
A few suggestions/questions

SteveL-MSFT (approved):

PR #434, opened 2024-05-14
Create apt package management resource.

JamesWTruher:
# PR Summary

This creates an package manager for `apt`
it includes some tests


resources/apt/test/assets/apt_install_rolldice.dsc.yaml
```diff
@@ -0,0 +1,19 @@
...
+    resources:
+    - name: os_check
+      type: Microsoft/OSInfo
+      properties:
+        family: Linux
+- name: apt_rolldice
+  type: DSC.PackageManagement/Apt
+  properties:
+    packageName: rolldice
+    _exist: true
```
SteveL-MSFT, line 17: It would be better if this value was defined as a parameter so instead of having two similar config, you could just pass in `false` (and default to `true`).  See https://github.com/PowerShell/DSC/blob/main/dsc/tests/dsc_parameters.tests.ps1 for examples

resources/apt/test/assets/apt_wget.dsc.yaml
```diff
@@ -0,0 +1,19 @@
...
+    $schema: https://raw.githubusercontent.com/PowerShell/DSC/main/schemas/2024/04/config/document.json
+    resources:
+    - name: os_check
+      type: Microsoft/OSInfo
+      properties:
+        family: Linux
+- name: apt_wget
+  type: DSC.PackageManagement/Apt
+  properties:
+    packageName: wget
```
SteveL-MSFT, line 16: `packageName` should probably also be a parameter

dsc/tests/dsc_resource_list.tests.ps1
```diff
@@ -17,7 +17,7 @@ Describe 'Tests for listing resources' {
     }
 
     It 'dsc resource list --tags "<tags>" and --description "<description> work' -TestCases @(
-        @{ tags = 'linux'; description = $null; expectedCount = 1; expectedType = 'Microsoft/OSInfo' }
+        @{ tags = 'linux'; description = $null; expectedCount = 2; expectedType = @('DSC.PackageManagement/Apt', 'Microsoft/OSInfo') }
```
SteveL-MSFT, line 20: Since this test now fails on Windows and macOS, you'll probably need to modify this test.  Simplest might be something like:

```powershell
expectedType = if ($IsLinux) { @(...) } else { @(...)}
```

SteveL-MSFT (changes requested):

resources/apt/apt.dsc.resource.json
```diff
@@ -0,0 +1,73 @@
...
+                    "type": "string",
+                    "title": "Version",
+                    "description": "Defines the version of the package to install"
+                },
+                "Source": {
+                    "type": "string",
+                    "title": "Source",
+                    "description": "Defines the source of the package"
+                },
+                "_exist": {
```
michaeltlombardi, line 65: Standard implementation guidance for well-known properties is to define them as a reference to the schema for that property:

```suggestion
                "_exist": {
                    "$ref": "https://raw.githubusercontent.com/PowerShell/DSC/main/schemas/2024/04/resource/properties/exist.json"
                }
```

For reference, see [DSC Resource _exist property schema][01].

[01]: https://learn.microsoft.com/powershell/dsc/reference/schemas/resource/properties/exist?view=dsc-3.0&preserve-view=true

resources/apt/apt.dsc.resource.json
```diff
@@ -0,0 +1,73 @@
+{
+    "$schema": "https://raw.githubusercontent.com/PowerShell/DSC/main/schemas/2024/04/bundled/resource/manifest.json",
+    "type": "DSC.PackageManagement/Apt",
+    "description": "DSC resource to manage Apt packages",
```
michaeltlombardi, line 4:
```suggestion
    "description": "Manage packages with the advanced package tool (APT)",
```

resources/apt/apt.dsc.resource.json
```diff
@@ -0,0 +1,73 @@
...
+        ],
+        "input": "env"
+    },
+    "set": {
+        "executable": "apt.dsc.resource.sh",
+        "args": [
+            "set"
+        ],
+        "input": "env",
+        "implementsPretest": true,
```
michaeltlombardi, line 24: Does it handle pretest? These lines look to me like set will just call `apt install` or `apt remove` - when this property is set to `true`, DSC assumes that the resource is testing whether to run code that changes the system before doing so.

```sh
    if [[ $_exist = true ]]; then
        apt install -y "${packageName}"
    else
        apt remove -y "${packageName}"
    fi
```

resources/apt/apt.dsc.resource.sh
```diff
@@ -0,0 +1,61 @@
...
+    if [[ -z $packageName ]]; then
+        echo "packageName not set"
+        exit 1
+    fi
+}
+
+get_apt() {
+    pkgname=$1
+    InstalledSection=0
+    apt list --installed $pkgname 2>&1 | while read line; do
```
michaeltlombardi, line 21: Nit: Should this script use `apt` or `apt-get` - my understanding is that the latter is preferred for shell scripts and the former is more for interactive use.

resources/apt/apt.dsc.resource.sh
```diff
@@ -0,0 +1,61 @@
...
+}
+
+if [[ "$#" -eq "0" ]]; then
+    echo "Command not provided, valid commands: get, set, export"
+    exit 1
+elif [[ "$1" == "get" ]]; then
+    check_args
+    output="$(get_apt $packageName)"
+    if [[ -z $output ]]; then
+	    printf '{ "_exist": "false", "packageName": "%s", "Version": "", "Source": "" }\n' $packageName
```
michaeltlombardi, line 41: We should always return a compressed JSON Line from our resources.

If we're going to keep the version and source fields, the casing should match the schema:
```suggestion
	    printf '{"_exist":"false","packageName":"%s","version":"","source":""}\n' $packageName
```
However, because both version and source are non-required, we can simply omit them:
```suggestion
	    printf '{"_exist":"false","packageName":"%s"}\n' $packageName
```

resources/apt/apt.dsc.resource.sh
```diff
@@ -0,0 +1,61 @@
...
+    else
+	    echo $output
+    fi
+elif [[ "$1" == "set" ]]; then
```
michaeltlombardi, line 45: Do we need to check if the resource is invoked as root? If I recall correctly, install and remove fail as normal users.

resources/apt/apt.dsc.resource.json
```diff
@@ -0,0 +1,73 @@
...
+                },
+                "version": {
+                    "type": "string",
+                    "title": "Version",
+                    "description": "Defines the version of the package to install"
+                },
+                "Source": {
+                    "type": "string",
+                    "title": "Source",
+                    "description": "Defines the source of the package"
```
michaeltlombardi, line 63: We should mark this value as read-only, since a user can't control the source used for the package.

```suggestion
                    "description": "Defines the source of the package",
                    "readOnly": true
```

resources/apt/apt.dsc.resource.sh
```diff
@@ -0,0 +1,61 @@
...
+    check_args
+    if [[ -z $_exist ]]; then
+        # if $_exist is not defined in the input, it defaults to `true`
+        _exist=true
+    fi
+    if [[ $_exist = true ]]; then
+        apt install -y "${packageName}"
```
michaeltlombardi, line 52: Should this handle installing a specific version, since we have the `version` property? Or should `version` be marked as `readOnly`?

SteveL-MSFT: I think this resource should support specifying the version, but limit it to apt syntax

resources/apt/apt.dsc.resource.sh
JamesWTruher: `apt-get list` is not valid on my system

resources/apt/apt.dsc.resource.sh
SteveL-MSFT: @michaeltlombardi perhaps we should have a general issue opened to add in the resource manifest when specific operations require elevated and DSC can fail fast

resources/apt/test/assets/apt_install_rolldice.dsc.yaml
JamesWTruher: I'm happy to leave that as an exercise for the reader

resources/apt/apt.dsc.resource.sh
SteveL-MSFT: We can add this later

SteveL-MSFT (approved):
We can iterate and improve separately

PR #438, opened 2024-05-22
update expressions to error for non-ints

### LINKED ISSUE #390: The `int()` function incorrectly handles non-integer numbers

tgauth:
# PR Summary
- fix #390 

## PR Context
- existing grammar parses only the decimal digit(s) of floating point input as a number, resulting in incorrect handling by functions, like `int()`
- adds check for `ERROR` nodes in Rust parser for function name and args

### DIFF
--- a/dsc_lib/src/functions/int.rs
+++ b/dsc_lib/src/functions/int.rs
@@ -75,6 +75,13 @@ mod tests {
     }
 
     #[test]
+    fn incomplete_float() {
+        let mut parser = Statement::new().unwrap();
+        let err = parser.parse_and_execute("[int(.2)]", &Context::new()).unwrap_err();
+        assert!(matches!(err, DscError::IntegerConversion(_)));
+    }
+
+    #[test]
     fn nested() {
         let mut parser = Statement::new().unwrap();
         let result = parser.parse_and_execute("[int(int('-1'))]", &Context::new()).unwrap();
@@ -84,7 +91,7 @@ mod tests {
     #[test]
     fn error() {
         let mut parser = Statement::new().unwrap();
-        let err = parser.parse_and_execute("[int('foo')]", &Context::new()).unwrap_err();
+        let err = parser.parse_and_execute("[int('foo.1')]", &Context::new()).unwrap_err();
         assert!(matches!(err, DscError::FunctionArg(_, _)));
     }
 }
--- a/dsc_lib/src/functions/int.rs
+++ b/dsc_lib/src/functions/int.rs
@@ -68,6 +68,13 @@ mod tests {
     }
 
     #[test]
+    fn float() {
+        let mut parser = Statement::new().unwrap();
+        let err = parser.parse_and_execute("[int(1.2)]", &Context::new()).unwrap_err();
+        assert!(matches!(err, DscError::IntegerConversion(_)));
+    }
+
+    #[test]
     fn nested() {
         let mut parser = Statement::new().unwrap();
         let result = parser.parse_and_execute("[int(int('-1'))]", &Context::new()).unwrap();
--- a/dsc_lib/src/functions/int.rs
+++ b/dsc_lib/src/functions/int.rs
@@ -75,9 +75,16 @@ mod tests {
     }
 
     #[test]
-    fn incomplete_float() {
+    fn incomplete_float_missing_digit() {
         let mut parser = Statement::new().unwrap();
         let err = parser.parse_and_execute("[int(.2)]", &Context::new()).unwrap_err();
+        assert!(matches!(err, DscError::Parser(_)));
+    }
+
+    #[test]
+    fn incomplete_float_missing_decimal() {
+        let mut parser = Statement::new().unwrap();
+        let err = parser.parse_and_execute("[int(2.)]", &Context::new()).unwrap_err();
         assert!(matches!(err, DscError::Parser(_)));
     }
 
--- a/dsc_lib/src/functions/int.rs
+++ b/dsc_lib/src/functions/int.rs
@@ -70,15 +70,15 @@ mod tests {
     #[test]
     fn float() {
         let mut parser = Statement::new().unwrap();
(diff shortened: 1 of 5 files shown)


tree-sitter-dscexpression/grammar.js
```diff
@@ -29,7 +29,7 @@ module.exports = grammar({
     _quotedString: $ => seq('\'', $.string, '\''),
     // ARM strings do not allow to contain single-quote characters
     string: $ => /[^']*/,
-    number: $ => /-?\d+/,
+    number: $ => /[+-]?([0-9]*[.])?[0-9]+/,
```
SteveL-MSFT, line 32: Do we need to support `+1`?  Do we know if ARM allows a `+` symbol?

Why use `[0-9]` instead of `\d`?  Do we want to support `.1` vs `0.1`?

SteveL-MSFT (changes requested):

tree-sitter-dscexpression/grammar.js
tgauth: updated to continue using `\d` - will not support `.1` or `+`

SteveL-MSFT (changes requested):

tree-sitter-dscexpression/corpus/invalid_expressions.txt
```diff
@@ -110,3 +110,17 @@ Incomplete expression
         (function
           (functionName)))
       (MISSING "]"))
+
+=====
+Invalid float argument
+=====
+[myFunction(a.1)]
```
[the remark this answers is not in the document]
tgauth: Added tests. To clarify what I said before - `+1` is unsupported by the grammar. `.1` is supported by the grammar, but detected as float within the Rust parsing, so unsupported overall.

SteveL-MSFT (changes requested):

tree-sitter-dscexpression/corpus/invalid_expressions.txt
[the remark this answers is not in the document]
tgauth: The Rust parsing _could_ support floats - it's just not setup to right now, whether it's `0.1` or `.1`.

I'm having trouble trying to exclude leading with the decimal point in the grammar because then the grammar detects the decimal portion as the number, which results in the functions like `int()` seemingly handling the input incorrectly -perhaps the Rust parser needs to check for any error nodes and stop execution if it finds one?

tree-sitter-dscexpression/grammar.js
```diff
@@ -29,7 +29,7 @@ module.exports = grammar({
     _quotedString: $ => seq('\'', $.string, '\''),
     // ARM strings do not allow to contain single-quote characters
     string: $ => /[^']*/,
-    number: $ => /-?\d+/,
+    number: $ => /-?(\d*[.])?\d+/,
```
SteveL-MSFT: If we want to prevent `.1`, then we just need to require at least one digit, so something like:

```
^-?\d+([.]\d+)?$
```

See https://regex101.com/r/lynDYN/1 as my tests

SteveL-MSFT (changes requested):

tree-sitter-dscexpression/corpus/invalid_expressions.txt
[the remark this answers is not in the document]
SteveL-MSFT: See my new comment in the regex below

tree-sitter-dscexpression/grammar.js
tgauth: If we add the check to the Rust parser, then the grammar can stay as it was before!
I kept the tree-sitter tests to make sure the floating point numbers result in error nodes.

dsc_lib/src/functions/int.rs
```diff
@@ -67,6 +67,20 @@ mod tests {
...
+    fn float() {
+        let mut parser = Statement::new().unwrap();
+        let err = parser.parse_and_execute("[int(1.0)]", &Context::new()).unwrap_err();
+        assert!(matches!(err, DscError::Parser(_)));
+    }
+
+    #[test]
+    fn incomplete_float() {
+        let mut parser = Statement::new().unwrap();
+        let err = parser.parse_and_execute("[int(.2)]", &Context::new()).unwrap_err();
```
SteveL-MSFT, line 80: Should also have a test for `2.`

SteveL-MSFT (changes requested):

SteveL-MSFT (approved):

PR #441, opened 2024-05-30
Add resource support for what if

### LINKED ISSUE #70: Handle `WhatIf`

Need to have resources support this to work completely

_No response_

tgauth:
# PR Summary
- fix #70 
- example for resource participation, where `whatif` is implemented like `set` but with an additional switch arg:

```jsonc
{
    "$schema": "https://aka.ms/dsc/schemas/resource_manifest",
    // truncated
    "set": {
        "executable": "whatifexample",
        "args": [],
        "return": "state"
    },
    "whatIf": {
        "executable": "whatifexample",
        "args": [
            "--what-if"
        ],
        "return": "state"
    },
    // truncated
}
```
- example output from resource that implements `whatif`, format matches `set`:
image

## PR Context
- add resource support for `config set --what-if`

### DIFF
--- a/dsc_lib/src/configure/config_doc.rs
+++ b/dsc_lib/src/configure/config_doc.rs
@@ -30,6 +30,7 @@ pub enum Operation {
 #[derive(Debug, Clone, PartialEq, Deserialize, Serialize, JsonSchema)]
 pub enum ExecutionKind {
     Actual,
+    // differentiate internally whether what-if should be processed by the resource or dsc engine
     #[serde(rename = "WhatIf")]
     WhatIfDSC,
     #[serde(rename = "WhatIf")]
--- a/dsc_lib/src/configure/config_doc.rs
+++ b/dsc_lib/src/configure/config_doc.rs
@@ -30,11 +30,7 @@ pub enum Operation {
 #[derive(Debug, Clone, PartialEq, Deserialize, Serialize, JsonSchema)]
 pub enum ExecutionKind {
     Actual,
-    // differentiate internally whether what-if should be processed by the resource or dsc engine
-    #[serde(rename = "WhatIf")]
-    WhatIfDSC,
-    #[serde(rename = "WhatIf")]
-    WhatIfResource,
+    WhatIf,
 }
 
 #[derive(Debug, Default, Clone, PartialEq, Deserialize, Serialize, JsonSchema)]
--- a/dsc_lib/src/configure/config_doc.rs
+++ b/dsc_lib/src/configure/config_doc.rs
@@ -30,7 +30,10 @@ pub enum Operation {
 #[derive(Debug, Clone, PartialEq, Deserialize, Serialize, JsonSchema)]
 pub enum ExecutionKind {
     Actual,
-    WhatIf,
+    #[serde(rename = "WhatIf")]
+    WhatIfDSC,
+    #[serde(rename = "WhatIf")]
+    WhatIfResource,
 }
 
 #[derive(Debug, Default, Clone, PartialEq, Deserialize, Serialize, JsonSchema)]
--- a/dsc_lib/src/dscresources/dscresource.rs
+++ b/dsc_lib/src/dscresources/dscresource.rs
@@ -50,6 +50,8 @@ pub enum Capability {
     Set,
     /// The resource supports the `_exist` property directly.
     SetHandlesExist,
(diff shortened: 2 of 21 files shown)


dsc_lib/src/dscresources/dscresource.rs
```diff
@@ -50,6 +50,8 @@ pub enum Capability {
     Set,
     /// The resource supports the `_exist` property directly.
     SetHandlesExist,
+    /// The resource supports the `what-if` execution type directly.
+    SetHandlesWhatIf,
```
SteveL-MSFT, line 54: This shouldn't be specific to `set` although primarily used by `set`.  `get` can also fail if the user doesn't have the right permission.  Should generalize as just `WhatIf`.

dsc_lib/src/dscresources/resource_manifest.rs
```diff
@@ -103,6 +103,11 @@ pub enum ArgKind {
         /// Indicates if argument is mandatory which will pass an empty string if no JSON input is provided.  Default is false.
         mandatory: Option<bool>,
     },
+    WhatIf {
+        /// The argument that serves as the what-if switch.
+        #[serde(rename = "whatIfInputArg")]
```
SteveL-MSFT, line 108: Should probably be called `whatIfSwitchArg` as it should not take any input

dsc_lib/src/configure/config_doc.rs
```diff
@@ -30,7 +30,11 @@ pub enum Operation {
 #[derive(Debug, Clone, PartialEq, Deserialize, Serialize, JsonSchema)]
 pub enum ExecutionKind {
     Actual,
-    WhatIf,
+    // differentiate internally whether what-if should be processed by the resource or dsc engine
+    #[serde(rename = "WhatIf")]
+    WhatIfDSC,
+    #[serde(rename = "WhatIf")]
+    WhatIfResource,
```
SteveL-MSFT, line 37: Do we need to differentiate between config and resource what-if?

SteveL-MSFT (changes requested):

dsc_lib/src/dscresources/dscresource.rs
tgauth: So the expectation is if the resource implements `what-if` directly - it's for every command? 

Should I add `get` support for `what-if` separately? I haven't added `what-if` to any commands besides `config set` so the `get` scenario won't work yet.

dsc_lib/src/configure/config_doc.rs
tgauth: Not sure if this answers the question but this is differentiating whether the `what-if` should be processed synthetically by DSC or whether the resource itself implements `what-if`

SteveL-MSFT: I see.  I don't think we need to explicitly differentiate just like we don't indicate if `test` is implemented or synthetic.  Is there a specific scenario where you think this is useful?

dsc_lib/src/dscresources/dscresource.rs
SteveL-MSFT: No, a resource can implement it for whatever operation makes sense.  `set` would be recommended, but it could apply to `get`, `test`, and even `export`.

SteveL-MSFT:
Should there be a separate schema for what resources return for `what-if` that aligns with how the engine synthesizes it?  Perhaps with addition of `reason` field that is open text that the resource can fill in

anmenaga (commented):
2 questions:
1) please outline the interface/how resource can participate/enlist/declare 'what-if' support in PR description
2) are adapter-based resources supported?

dsc_lib/src/dscresources/dscresource.rs
mgreenegit: -whatif is not a common parameter for get-* cmdlets, by comparison.

anmenaga:
@tgauth please open an issue to add 'whatif' support to PS Adapter.

SteveL-MSFT:
Based on the review meeting:

- Add new `whatIf` section as a top-level operation in the resource manifest which only applies to `set`, all other changes in the code will flow from this

SteveL-MSFT:
@tgauth can you provide example of what-if output where a resource participates?

SteveL-MSFT (approved):

PR #444, opened 2024-06-04
retrieve exit code info from manifest if available

### LINKED ISSUE #431: Update internal `invoke_command()` to retrieve the exit code description from manifest

Currently, if the exit code is non-zero, the function returns it as an error.  however, the resource manifest can describe what specific exit codes mean.  The helper function should be updated so that if the exit code maps to one in the manifest, that information is shown, otherwise it's "unknown".

_No response_

tgauth:
# PR Summary
- fix #431 

## PR Context
- retrieve non-zero exit code description from manifest if available 
- create `Test/ExitCode` resource to test changes

### DIFF
--- a/dsc/tests/dsc.exit_code.tests.ps1
+++ b/dsc/tests/dsc.exit_code.tests.ps1
@@ -4,7 +4,7 @@
 Describe 'exit code tests' {
     It 'non-zero exit code in manifest has corresponding message' {
         $result = dsc resource get -r Test/ExitCode --input "{ exitCode: 8 }" 2>&1
-        $result | Should -Match 'ERROR.*?[Exit code 8].*?Placeholder from manifest for exit code 8'
+        $result | Should -Match 'ERROR.*?[Exit code 8].*?manifest description: Placeholder from manifest for exit code 8'
     }
     It 'non-zero exit code not in manifest has generic message' {
         $result = dsc resource get -r Test/ExitCode --input "{ exitCode: 1 }" 2>&1
--- a/dsc/tests/dsc.exit_code.tests.ps1
+++ b/dsc/tests/dsc.exit_code.tests.ps1
@@ -1,0 +1,18 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+Describe 'exit code tests' {
+    It 'non-zero exit code in manifest has corresponding message' {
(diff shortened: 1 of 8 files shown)


dsc_lib/src/dscresources/command_resource.rs
```diff
@@ -614,6 +614,11 @@ pub fn invoke_command(executable: &str, args: Option<Vec<String>>, input: Option
     };
 
     if exit_code != 0 {
+        if let Some(exit_codes) = exit_codes {
+            if let Some(error_message) = exit_codes.get(&exit_code) {
+                return Err(DscError::Command(executable.to_string(), exit_code, error_message.to_string()));
```
SteveL-MSFT, line 619: Should have a new error type for this in case we want to change the message or make it more clear it's mapped

SteveL-MSFT (changes requested):

dsc/tests/dsc.exit_code.tests.ps1
```diff
@@ -0,0 +1,18 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+Describe 'exit code tests' {
+    It 'non-zero exit code in manifest has corresponding message' {
+        $result = dsc resource get -r Test/ExitCode --input "{ exitCode: 8 }" 2>&1
```
anmenaga, line 6: Please add a test for `dsc resource list` operation.
FYI: it emits WARN instead of ERROR.

anmenaga (changes requested):

dsc/tests/dsc.exit_code.tests.ps1
tgauth: here? https://github.com/PowerShell/DSC/blob/main/dsc_lib/src/discovery/command_discovery.rs#L224-L238

SteveL-MSFT (approved):

dsc/tests/dsc.exit_code.tests.ps1
anmenaga: Yes.
I guess this applies only to adapters, because `list` for non-adapted resources is just reading a `manifest.json` file.

tgauth: That makes sense, thanks! Completed the manual verification

anmenaga (approved):

PR #448, opened 2024-06-05
VS Code DSCv3 completion provider

anmenaga:
# PR Summary

This is a prototype DSCv3 VS Code extension that provides completions for classic PS-class-based DSC resources when authoring `*.dsc.yaml` configuration files.

Completions are displayed when `type:` is typed in a `*.dsc.yaml` files.
This requires latest dsc build from `main` branch and relies on PS Adapter cache, which can be generated by `dsc resource list * -a Microsoft.DSC/PowerShell`.

https://github.com/PowerShell/DSC/assets/11860095/ec88ddb8-6296-4120-9c46-7748ef4a337c


vscode/DscExt/src/extension.ts
```diff
@@ -0,0 +1,67 @@
...
+					ps_cache_file_path = process.env.HOME + '/.dsc/PSAdapterCache.json';
+				}
+				console.log("Using cache path " + ps_cache_file_path);
+
+				if (!fs.existsSync(ps_cache_file_path)) {
+					console.log("Cache file does not exist");
+					return [];
+				}
+
+				const dataArray = JSON.parse(fs.readFileSync(ps_cache_file_path, 'utf-8'));
```
andyleejordan, line 35: What's the reason we made this synchronous? If not let's use `readFile` from Node's `fs/promises` and get this code back to being async and non-blocking (ditto above with the exist check).

vscode/DscExt/src/extension.ts
```diff
@@ -0,0 +1,67 @@
...
+import * as fs from 'fs';
+
+export function activate(context: vscode.ExtensionContext) {
+
+	const dsc_type_provider = vscode.languages.registerCompletionItemProvider({ language: 'yaml', pattern: '**/*.dsc.yaml' },
+		{
+			provideCompletionItems(document: vscode.TextDocument, position: vscode.Position) {
+
+				const linePrefix = document.lineAt(position).text.slice(0, position.character);
+				if (!linePrefix.endsWith('type:')) {
```
andyleejordan, line 15: I'm unsure how VS Code handles filtering provided completions based on partial input (which we're discarding here)...something to look into. Like, are we the provider supposed to filter it or not?

vscode/DscExt/src/extension.ts
```diff
@@ -0,0 +1,67 @@
...
+					return undefined;
```
andyleejordan, line 16: Is there more completion we want to do in the future, or is after "type:" really the only thing we want to / can provide completions for?

vscode/DscExt/src/extension.ts
```diff
@@ -0,0 +1,67 @@
...
+				console.log("Using cache path " + ps_cache_file_path);
+
+				if (!fs.existsSync(ps_cache_file_path)) {
+					console.log("Cache file does not exist");
+					return [];
+				}
+
+				const dataArray = JSON.parse(fs.readFileSync(ps_cache_file_path, 'utf-8'));
+				var completions:vscode.CompletionItem[] = [];
+				dataArray.ResourceCache.forEach(function (resouce: any) {
```
andyleejordan, line 37: Just noting `resouce` dropped an `r` 😅

andyleejordan: It's just an optimization but since (to my eye) everything in the lambda is...I don't know, is the word stateless? Not quite idempotent? But like the closure doesn't mutate anything, so we could fire all these completion builders off at once with `await Promise.all(dataArray.ResourceCache.map(async (resource: any)...` right?

vscode/DscExt/src/extension.ts
```diff
@@ -0,0 +1,67 @@
...
+						
+					const all_props_completion = new vscode.CompletionItem(resouce.Type);
```
andyleejordan, line 39: Should we specify a kind for the completion items? (One of these: https://vscode-api.js.org/enums/vscode.CompletionItemKind.html)

vscode/DscExt/src/extension.ts
```diff
@@ -0,0 +1,67 @@
...
+					const key_props_completion = new vscode.CompletionItem(resouce.Type + " [keys only]");
+
+					let all_props_comp_text: string = ' ' + resouce.Type;
+					let key_props_comp_text: string = ' ' + resouce.Type;
+					all_props_comp_text += '\nproperties:';
+					key_props_comp_text += '\nproperties:';
+					resouce.DscResourceInfo.Properties.forEach(function (prop: any) {
+						all_props_comp_text += '\n  ' + prop.Name + ': ';
```
andyleejordan, line 47: If you want, it's probably a lot more readable to write this snippet as multiple strings ("which will be joined as multiple lines upon insertion") and is there any insertions or variables we might want to provide inside the completion snippet? https://code.visualstudio.com/docs/editor/userdefinedsnippets#_create-your-own-snippets

vscode/DscExt/src/extension.ts
andyleejordan: Ok so technically you're pushing onto `completions` from outside the closure but that was just an implementation choice, since this could also just return the completions from each async lambda and join them together as the final set of results. Is there any particular sorting we're wanting to have happen?

andyleejordan (commented):
I like it! It's a great start.

ThomasNieto:
@anmenaga any chance we can get this one over the finish line? It would be a very useful feature for increasing the usability of creating configuration documents.

SteveL-MSFT:
@ThomasNieto We're prioritizing our Bicep extension work so I don't think we'll spend the effort to complete this one

ThomasNieto:
@SteveL-MSFT will the bicep extension have this feature?

andyleejordan:
@ThomasNieto there's an experimental preview of a Bicep extension to provide DSC completions in a Bicep file here: https://github.com/microsoft/bicep-types-dsc

It is very early work. It demonstrates that DSC resources can be discovered and then have their schemas converted on the fly to Bicep types, providing IntelliSense in VS Code. However, when enabled it results in ARMv2 templates from Bicep which are as yet incompatible with DSC (see #766) so it currently no more than a demo.

PR #453, opened 2024-06-07
Add example configs for winget and wmi

SteveL-MSFT:
# PR Summary

This PR adds example configs for `winget` and `wmi`.  Adds example of using `env()` function to config.

Decided not to add the "demo script" from PSConfEU as you need lots of context to understand the bullet points


dsc/examples/PSConfEU_2024_Demo.md
```diff
@@ -0,0 +1,103 @@
...
+
+### List DSC resources
+
+- List DSC resources
+
+```powershell
+dsc resource list
+```
+
+- List PowerShell Adapted DSC resources
```
michaeltlombardi, line 13:
```suggestion
- List adapted PowerShell DSC resources
```

dsc/examples/PSConfEU_2024_Demo.md
```diff
@@ -0,0 +1,103 @@
...
+```powershell
+dsc resource list --adapter Microsoft.DSC/PowerShell
+```
+
+- Find DSC resources
+
+```powershell
+dsc resource list '*DSC*'
+```
+
```
michaeltlombardi, line 24: We should also discuss searching for PSDSC resources specifically, highlighting that they use the `<ModuleName>/<ResourceName>` convention in the search
``````suggestion

- Find specific adapted PSDSC resources

```powershell
dsc resource list --adapter Microsoft.DSC/PowerShell PSDscResources/*
``````

File: dsc/examples/PSConfEU_2024_Demo.md [hunk 3]
```diff
@@ -0,0 +1,103 @@
...
+- List DSC resources
+
+```powershell
+dsc resource list
+```
+
+- List PowerShell Adapted DSC resources
+
+```powershell
+dsc resource list --adapter Microsoft.DSC/PowerShell
```
michaeltlombardi, line 16: Is it worth pointing out the Windows PowerShell adapter, too?

```suggestion
dsc resource list --adapter Microsoft.DSC/PowerShell
dsc resource list --adapter Microsoft.Windows/PowerShell
```

dsc/examples/PSConfEU_2024_Demo.md
```diff
@@ -0,0 +1,103 @@
...
+- JSON vs YAML
+- Args, JSON Input, Env Var
+- JSON Schema
+
+### Resource Types
+
+- Resource
+- Group
+- Adapter
+- Import
```
michaeltlombardi, line 59: We probably want to be careful about calling these kinds, since `type` in the syntax and usage means "specific resource."

```suggestion
### Resource Kinds

- Resource
- Group
- Adapter
- Import
```

dsc/examples/PSConfEU_2024_Demo.md
```diff
@@ -0,0 +1,103 @@
...
+### Expressions
+
+- Functions
+  - reference()
+  - env()
+- Dot-notation
+
+### Group Resources
+
+- DependsOn scope
```
michaeltlombardi, line 80: Should call out the effective limitations on reference function as well.

SteveL-MSFT: I'll reword this as discussing scope and have DependsOn and Reference examples

dsc/examples/wmi.dsc.yaml
```diff
@@ -0,0 +1,12 @@
+$schema: https://raw.githubusercontent.com/PowerShell/DSC/main/schemas/2024/04/config/document.json
+resources:
+- name: WMI
+  type: Microsoft.Windows/WMI
+  properties:
+    resources:
+      - name: computer system
+        type: root.cimv2/Win32_ComputerSystem
+        properties:
```
anmenaga, line 9: Looks like we need to modify the schema to make this 'properties' field optional if no are specified.

anmenaga (approved):

dsc/examples/wmi.dsc.yaml
SteveL-MSFT: Actually, `properties` is already optional, so removing from example rather than having it empty

PR #454, opened 2024-06-10
Linux Service module

anmenaga:
# PR Summary

A simple PowerShell class-based resource that lists services on Linux.
Supports `Get` and `Export`.
It uses `systemctl -l --type service --all` to access the data.

LinuxServiceModule


resources/Microsoft.DSC.Experimental/0.0.1/Microsoft.DSC.Experimental.psd1
```diff
@@ -0,0 +1,50 @@
...
+CmdletsToExport = '*'
+
+# Variables to export from this module
+VariablesToExport = @()
+
+# Aliases to export from this module, for best performance, do not use wildcards and do not delete the entry, use an empty array if there are no aliases to export.
+AliasesToExport = @()
+
+# DSC resources to export from this module
+DscResourcesToExport = 'Service'
```
michaeltlombardi, line 37: We should probably more explicitly indicate that the resource manages services through `systemctl` in the name - maybe `systemctlService` as the name? As-is, the generic class name `Service` seems likely to raise a conflict if someone else implements a `Service` resource in PSDSC.

resources/Microsoft.DSC.Experimental/0.0.1/Microsoft.DSC.Experimental.psm1
```diff
@@ -0,0 +1,76 @@
...
+            $obj.Sub = $_.Sub
+            $obj.Description = $_.Description
+            
+            $resultList.Add($obj)
+        }
+
+        return $resultList.ToArray()
+    }
+
+    static [pscustomobject[]] GetServices()
```
michaeltlombardi, line 57: Is there a reason we return an array of custom objects instead of instances of the class? I also think it might be useful to just shorten the method name to `Get()` or `GetAll()` since the class name already indicates that it's going to retrieve services.

resources/Microsoft.DSC.Experimental/0.0.1/Microsoft.DSC.Experimental.psm1
```diff
@@ -0,0 +1,76 @@
...
+
+    [bool] Test()
+    {
+        return $false
+    }
+
+    [Service] Get()
+    {
+       
+        return [Service]::GetServices() | ? {$_.Unit -eq $this.Unit }
```
michaeltlombardi, line 36:
```suggestion
        return [Service]::GetServices() | Where-Object {$_.Unit -eq $this.Unit }
```

resources/Microsoft.DSC.Experimental/0.0.1/Microsoft.DSC.Experimental.psd1
SteveL-MSFT: I think there's two routes we can go here, one is a `systemctl` specific resource and one that abstracts the different Linux's.  In the latter case, the resource can either have a fixed knowledge base of the different distros or search the system for the relevant command.  I can see pros/cons for each approach.

michaeltlombardi: My _instinct_ is to go with per-implementation resources and consider an alias[^1] for each implementation. The complexities of service management across implementations gets out of hand quickly. For example, the [Puppet service resource][01] abstracts managing services across a bunch of backends, but then that leads to having to [implement logic to support those backends][02], using the same data-model.

It also means being poked repeatedly to support new distros, or update support, and manage all the exponentially expanding complexities that introduces.

On the other hand, with PSDSC resources being class-based, I also don't see a particular problem with defining implementation-specific resources and a meta resource that handles mapping to the appropriate resource. You could have `ServiceSystemD` and `ServiceInit` as the more direct resources, and a `Service` resource that maps to the appropriate behavior behind the scenes, converting more readable / general properties to what those specific implementations expect/need.

There's also a knock-on effect for documentation and guidance - if we support first-party meta resources, we have to explain how it selects the implementation and define which resource properties are conditional based on the backend implementation, that sort of thing. For a first pass, I think explicit-to-one-implementation is simpler to implement, maintain, test, and document. If and when people ask for an init-supporting resource, we can use that and the intervening experience to judge whether to do per-backend resources only or if it's worth investing in the meta resource.

[^1]: By alias, for PSDSC resources I mean simply defining a class that inherits from the explicitly named class - so if we named this class resource `ServiceSystemD`, we might also do something like `[DscResource()] class ServiceUbuntu : ServiceSystemD {}` - probably need more thought put into it than this, but that was my first thought.

[01]: https://www.puppet.com/docs/puppet/7/types/service.html
[02]: https://github.com/puppetlabs/puppet/tree/main/lib/puppet/provider/service

resources/Microsoft.DSC.Experimental/0.0.1/Microsoft.DSC.Experimental.psd1
```diff
@@ -0,0 +1,50 @@
...
+
+# Private data to pass to the module specified in RootModule/ModuleToProcess. This may also contain a PSData hashtable with additional module metadata used by PowerShell.
+PrivateData = @{
+    PSData = @{
+        DscCapabilities = @(
+            'Get'
+            'Export'
+        )
```
michaeltlombardi, line 45: Is this wired up anywhere already? Does this indicate the capabilities of all resources in the module, or just this resource?

resources/Microsoft.DSC.Experimental/0.0.1/Microsoft.DSC.Experimental.psd1
SteveL-MSFT: I'm ok with being explicit here and calling this the sytemctl resource since we already started with an apt and brew specific resources

anmenaga: Renamed to `SystemctlService`.

resources/Microsoft.DSC.Experimental/0.0.1/Microsoft.DSC.Experimental.psm1
anmenaga: Changed to `[SystemctlService[]] GetAll()`.

resources/Microsoft.DSC.Experimental/0.0.1/Microsoft.DSC.Experimental.psm1
anmenaga: Updated.

resources/Microsoft.DSC.Experimental/0.0.1/Microsoft.DSC.Experimental.psd1
anmenaga: Yes, it goes to `Caps` column of `dsc list` output.
I think putting it in `PrivateData` of a psd1 was a temporary solution, and it applies to all resources in the module.

resources/Microsoft.DSC.Experimental/0.0.1/Microsoft.DSC.Experimental.psd1
```diff
@@ -0,0 +1,50 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+@{
+
+# Script module or binary module file associated with this manifest.
+RootModule = 'Microsoft.DSC.Experimental.psm1'
```
SteveL-MSFT, line 7: We may have more experimental resources in the future, you should call this `Microsoft.DSC.Experimental.Systemctl`

resources/Microsoft.DSC.Experimental/0.0.1/Microsoft.DSC.Experimental.psd1
```diff
@@ -0,0 +1,50 @@
...
+CmdletsToExport = '*'
+
+# Variables to export from this module
+VariablesToExport = @()
+
+# Aliases to export from this module, for best performance, do not use wildcards and do not delete the entry, use an empty array if there are no aliases to export.
+AliasesToExport = @()
+
+# DSC resources to export from this module
+DscResourcesToExport = 'SystemctlService'
```
SteveL-MSFT, line 37: Do we need the `Service` suffix?

SteveL-MSFT (changes requested):

resources/Microsoft.DSC.Experimental/0.0.1/Microsoft.DSC.Experimental.psd1
anmenaga: renamed.

resources/Microsoft.DSC.Experimental/0.0.1/Microsoft.DSC.Experimental.psd1
anmenaga: Looks like `systemctl` can manage other types, and here we are using `--type service` only, so I feel like `Service` suffix is appropriate.

SteveL-MSFT (approved):

PR #456, opened 2024-06-10
Fixed RootModule bug

anmenaga:
# PR Summary

Some PS modules do not specify `RootModule` in their `psd1`.
In such cases `Get-Module` fills in `RootModule` with full path to `psd1`, which is misleading for discovering class-based resources.
This was causing internal failures in PsDscAdapter.
This PR fixes that with explicit check for such condition and adds some tracing to make diagnostics easier.

### DIFF
--- a/powershell-adapter/psDscAdapter/psDscAdapter.psm1
+++ b/powershell-adapter/psDscAdapter/psDscAdapter.psm1
@@ -67,7 +67,7 @@ function FindAndParseResourceDefinitions
         return
     }
     
-    "Loading resources from '$filePath'" | Write-DscTrace -Operation Trace
+    "Loading resources from file '$filePath'" | Write-DscTrace -Operation Trace
     #TODO: Handle class inheritance 
     #TODO: Ensure embedded instances in properties are working correctly
     [System.Management.Automation.Language.Token[]] $tokens = $null
@@ -161,8 +161,18 @@ function LoadPowerShellClassResourcesFromModule
         [PSModuleInfo]$moduleInfo
     )
 
+    "Loading resources from module '$($moduleInfo.Path)'" | Write-DscTrace -Operation Trace
+    
     if ($moduleInfo.RootModule)
     {
+        if  (([System.IO.Path]::GetExtension($moduleInfo.RootModule) -ne ".psm1") -and
+            ([System.IO.Path]::GetExtension($moduleInfo.RootModule) -ne ".ps1") -and
+            (-not $z.NestedModules))
+        {
+            "RootModule is neither psm1 nor ps1 '$($moduleInfo.RootModule)'" | Write-DscTrace -Operation Trace
+            return [System.Collections.Generic.List[DscResourceInfo]]::new()
+        }
+
         $scriptPath = Join-Path $moduleInfo.ModuleBase  $moduleInfo.RootModule
     }
     else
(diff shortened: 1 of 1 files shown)


powershell-adapter/psDscAdapter/psDscAdapter.psm1
```diff
@@ -67,7 +67,7 @@ function FindAndParseResourceDefinitions
         return
     }
     
-    "Loading resources from '$filePath'" | Write-DscTrace -Operation Trace
+    "Loading resources from file '$filePath'" | Write-DscTrace -Operation Trace
```
SteveL-MSFT, line 70: Not a blocker, but using the pipeline is inherently slower than using the parameter to the helper function directly as it forces use of the parameter binder

powershell-adapter/psDscAdapter/psDscAdapter.psm1
```diff
@@ -177,7 +187,9 @@ function LoadPowerShellClassResourcesFromModule
         foreach ($nestedModule in $moduleInfo.NestedModules)
         {
             $resourcesOfNestedModules = LoadPowerShellClassResourcesFromModule $nestedModule
-            $Resources.AddRange($resourcesOfNestedModules)
+            if ($resourcesOfNestedModules) {
```
SteveL-MSFT, line 190: In this case, if there isn't a RootModule, then an empty collection is returned and we add that range which doesn't do anything but use CPU.  Would it be better to check `if ($resourcesOfNestedModules.Count -gt 0)` or is there a case where it might be `$null`?

SteveL-MSFT (changes requested):

michaeltlombardi:
What about cases where people split their module code without using `NestedModules`?

For example, the pattern built by the DSC Community has base classes etc defined in a sub-module and has using/import statements to retrieve the code:

- [SqlServerDsc root module][01]
- [SqlServerDsc module manifest][02]

[01]: https://www.powershellgallery.com/packages/SqlServerDsc/16.6.1-preview0002/Content/SqlServerDsc.psm1
[02]: https://www.powershellgallery.com/packages/SqlServerDsc/16.6.1-preview0002/Content/SqlServerDsc.psd1

powershell-adapter/psDscAdapter/psDscAdapter.psm1
anmenaga: Pipeline looks cooler though :)

powershell-adapter/psDscAdapter/psDscAdapter.psm1
anmenaga: 1) `if ($resourcesOfNestedModules) ` checks for both conditions: `-ne $null` **and** `.Count -gt 0`.
2) yes, `LoadPowerShellClassResourcesFromModule` can return `null` (in error cases).

anmenaga:
> What about cases where people split their module code without using `NestedModules`?
> 
> For example, the pattern built by the DSC Community has base classes etc defined in a sub-module and has using/import statements to retrieve the code:
> 
> * [SqlServerDsc root module](https://www.powershellgallery.com/packages/SqlServerDsc/16.6.1-preview0002/Content/SqlServerDsc.psm1)
> * [SqlServerDsc module manifest](https://www.powershellgallery.com/packages/SqlServerDsc/16.6.1-preview0002/Content/SqlServerDsc.psd1)

1. strictly speaking this is out of scope of this particular bug;
2. I've looked through the PSAdapter code and your example module; and looks like such cases should be handled just fine when...
3. ... [class-inheritance will be implemented](https://github.com/PowerShell/DSC/issues/462).
4. Thanks for a good example module; that will help with development.

SteveL-MSFT (approved):
We can address other issues as separate PRs

PR #457, opened 2024-06-12
remove support to retrieve schema from URL

SteveL-MSFT:
# PR Summary

The support for schema URL was initially intended for intellisense support.  However, we can still resolve resources for intellisense using existing other means to get the json schema.  We should not allow reaching out to internet during configuration deployment.

### DIFF
--- a/dsc_lib/Cargo.toml
+++ b/dsc_lib/Cargo.toml
@@ -11,7 +11,6 @@ indicatif = { version = "0.17" }
 jsonschema = "0.17"
 num-traits = "0.2"
 regex = "1.7"
-reqwest = { version = "0.12", features = ["blocking"] }
 schemars = { version = "0.8.12", features = ["preserve_order"] }
 serde = { version = "1.0", features = ["derive"] }
 serde_json = { version = "1.0", features = ["preserve_order"] }
--- a/dsc_lib/src/dscerror.rs
+++ b/dsc_lib/src/dscerror.rs
@@ -4,7 +4,6 @@
 use std::str::Utf8Error;
 
 use indicatif::style::TemplateError;
-use reqwest::StatusCode;
 use thiserror::Error;
 use tracing::error;
 use tree_sitter::LanguageError;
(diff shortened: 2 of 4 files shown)


michaeltlombardi:
What about resolving references in the embedded-or-command-returned schema? Will we require [bundling][01] for resource JSON schemas?

[01]: https://json-schema.org/blog/posts/bundling-json-schema-compound-documents

SteveL-MSFT:
@michaeltlombardi yeah, forgot about referenced schema.  We should discuss this further.

SteveL-MSFT:
@michaeltlombardi actually, I just remembered that the JsonSchema crate has its own built in way to retrieve web referenced schema so this PR is literally only removing it as an option the manifest so I think we're ok

michaeltlombardi:
I mean that, from a guidance perspective, if the concern is either offline-functionality or no-web-calls-on-apply, we should advise authors on whether and when to bundle their schemas.

SteveL-MSFT:
@michaeltlombardi with this change, we require schema to be shipped with the resource whether it's embedded or via command.  We should recommend that those schema should not use web references as that would mean their resource won't work on some systems that limit outbound connections.

michaeltlombardi:
@SteveL-MSFT One dangling question then:

In our docs, we recommend that users reference the well-known properties instead of redefining them, e.g.

```json
"_rebootRequested": {
  "$ref": "https://raw.githubusercontent.com/PowerShell/DSC/main/schemas/2024/04/resource/properties/rebootRequested.json"
}
```

This ensures that users don't accidentally mis-defined the properties. I think we can tell users either that they can reference DSC schemas and subschemas, or that they need to bundle these into their full definition. I prefer the former, for simplicity - but some bundling tools will also find and pull those references in for them, too.

SteveL-MSFT:
> @SteveL-MSFT One dangling question then:
> 
> In our docs, we recommend that users reference the well-known properties instead of redefining them, e.g.
> 
> ```json
> "_rebootRequested": {
>   "$ref": "https://raw.githubusercontent.com/PowerShell/DSC/main/schemas/2024/04/resource/properties/rebootRequested.json"
> }
> ```
> 
> This ensures that users don't accidentally mis-defined the properties. I think we can tell users either that they can reference DSC schemas and subschemas, or that they need to bundle these into their full definition. I prefer the former, for simplicity - but some bundling tools will also find and pull those references in for them, too.

Referencing schema should still be recommended and this change shouldn't affect that.  This should only affect having the `url` type of schema in the manifest.

Issue #284, opened 2023-11-21
DSC/PowerShellGroup incorrect group inDesiredState


msftrubengu:
### Prerequisites

- [X] Write a descriptive title.
- [X] Make sure you are able to repro it on the latest version
- [X] Search the existing issues.

### Steps to reproduce

When all the resources in a PowerShellGroup are in the desired state, the group's inDesired state property is false. Looking at the code, I think it is because dsc compares the `desiredState` with `actualState`, but in the case of a PowerShellGroup the results are always going to be a single element with InDesiredState or an array of with the resources InDesiredState. Maybe dsc has to special case this group's result.

### Expected behavior

```console
Group's inDesiredState should be true if all its resource are in the desired state.
```

### Actual behavior

```console
PS D:\Dev\DSC> dsc resource test -r DSC/PowerShellGroup -i '{ "resources": [ { "name": "test", "type": "xE2ETestResource/E2EFileResource", "properties": { "path": "D:\\Dev\\DscV3\\testv3.txt", "content": "test" } } ] }'
  2023-11-21T21:59:55.212660Z  INFO dsc_lib::dscresources::command_resource: Invoking test DSC/PowerShellGroup using pwsh
    at D:\Dev\DSC\dsc_lib\src\dscresources\command_resource.rs:232

desiredState:
  resources:
  - name: test
    type: xE2ETestResource/E2EFileResource
    properties:
      path: D:\Dev\DscV3\testv3.txt
      content: test
actualState:
  InDesiredState: false
inDesiredState: false
differingProperties:
- resources
PS D:\Dev\DSC> dsc resource set -r DSC/PowerShellGroup -i '{ "resources": [ { "name": "test", "type": "xE2ETestResource/E2EFileResource", "properties": { "path": "D:\\Dev\\DscV3\\testv3.txt", "content": "test" } } ] }'
  2023-11-21T22:00:40.620243Z  INFO dsc_lib::dscresources::command_resource: Getting current state for set by invoking get DSC/PowerShellGroup using pwsh
    at D:\Dev\DSC\dsc_lib\src\dscresources\command_resource.rs:132

  2023-11-21T22:00:49.253187Z  INFO dsc_lib::dscresources::command_resource: Invoking set DSC/PowerShellGroup using pwsh
    at D:\Dev\DSC\dsc_lib\src\dscresources\command_resource.rs:145

beforeState:
  Path: D:\Dev\DscV3\testv3.txt
  Ensure: Present
  Content: ''
[...]

PR #464, opened 2024-06-17
Enable `set` and `test` for Import resources

SteveL-MSFT:
# PR Summary

This follows the PR to support `Import` kind resources by enabling `set` and `test`.  As part of this change, the output for group resources should also address #266.  Note that `test` is not complete yet as the engine currently just uses the input config as the desired state, but even after initial resolving of the include, that config still needs to have expressions invoked and parameters resolved for a valid comparison.  That work will be addressed later.

As part of this change, the `set` group result and `test` group result now align with `get` which changes the schema.

A helper function was added for `test` to recursively determine if in desired state by looking for `inDesiredState` property and if any are false than not in desired state (I think this might address https://github.com/PowerShell/DSC/issues/284)

## PR Context

Fix https://github.com/PowerShell/DSC/issues/266

### DIFF
--- a/dsc/tests/dsc_include.tests.ps1
+++ b/dsc/tests/dsc_include.tests.ps1
@@ -10,15 +10,6 @@ Describe 'Include tests' {
         $osinfoParametersConfigPath = Get-Item (Join-Path $includePath 'osinfo.parameters.yaml')
 
         $logPath = Join-Path $TestDrive 'stderr.log'
-
-        $includeConfig = @'
-            `$schema: https://raw.githubusercontent.com/PowerShell/DSC/main/schemas/2024/04/config/document.json
-            resources:
-            - name: Echo
-              type: Test/Echo
-              properties:
-                output: Hello World
-'@
     }
 
     It 'Include config with default parameters' {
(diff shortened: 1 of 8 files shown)


dsc/tests/dsc_include.tests.ps1
```diff
@@ -183,4 +174,38 @@ resources:
...
+        $echoConfigPathLeaf = (Split-Path $echoConfigPath -Leaf).Replace('\', '\\')
+        $directorySeparator = [System.IO.Path]::DirectorySeparatorChar.ToString().Replace('\', '\\')
+
+        $includeConfig = @"
+`$schema: https://raw.githubusercontent.com/PowerShell/DSC/main/schemas/2024/04/config/document.json
+resources:
+- name: nested
+  type: Microsoft.DSC/Include
+  properties:
+    configurationFile: "[concat('$echoConfigPathParent', '$directorySeparator', '$echoConfigPathLeaf')]"
```
anmenaga, line 201: why using `concat` instead of using just full path, aka `$echoConfigPath` ?

dsc/tests/dsc_include.tests.ps1
```diff
@@ -183,4 +174,38 @@ resources:
         $out.results[1].result[0].result[0].type | Should -Be 'Test/Echo'
         $out.results[1].result[0].result[0].result[0].actualState.output | Should -Be 'one'
     }
+
+    It 'Set with include works' {
```
anmenaga, line 178: Is there a test for `test` operation?

anmenaga (approved):

dsc/tests/dsc_include.tests.ps1
SteveL-MSFT: It'll be added once `test` is complete

dsc/tests/dsc_include.tests.ps1
SteveL-MSFT: It was to test that expressions get resolved for Include

PR #465, opened 2024-06-17
Add registry whatif

### LINKED ISSUE #452: Implement `whatif` for Registry resource

`whatif` for Registry set should check permissions to determine if the set would actually work

_No response_

tgauth:
# PR Summary
- fix #452 
- example output for creating a new key:
image

- example output for clobbering an existing value:
image

- example output for an insufficient permissions error: 
image

## PR Context
- start of `whatIf` for registry resource, to be continued once metadata can be passed back from resources

### DIFF
--- a/registry/src/config.rs
+++ b/registry/src/config.rs
@@ -37,5 +37,5 @@ pub struct Registry {
 #[serde(deny_unknown_fields)]
 pub struct WhatIf {
     #[serde(skip_serializing_if = "Option::is_none")]
-    pub message: Option<String>
+    pub error: Option<String>
 }
--- a/registry/src/config.rs
+++ b/registry/src/config.rs
@@ -21,8 +21,8 @@ pub struct Registry {
     #[serde(rename = "keyPath")]
     pub key_path: String,
     /// The information from a config set --what-if operation.
-    #[serde(flatten, skip_serializing_if = "Option::is_none")]
-    pub what_if: Option<WhatIf>,
+    #[serde(rename = "_metadata", skip_serializing_if = "Option::is_none")]
+    pub metadata: Option<Metadata>,
     /// The name of the registry value.
     #[serde(rename = "valueName", skip_serializing_if = "Option::is_none")]
     pub value_name: Option<String>,
@@ -35,7 +35,7 @@ pub struct Registry {
 
 #[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize, Deserialize, JsonSchema)]
 #[serde(deny_unknown_fields)]
-pub struct WhatIf {
-    #[serde(skip_serializing_if = "Option::is_none")]
-    pub error: Option<String>
+pub struct Metadata {
+    #[serde(rename = "whatIf", skip_serializing_if = "Option::is_none")]
+    pub what_if: Option<Vec<String>>
 }
--- a/registry/src/config.rs
+++ b/registry/src/config.rs
@@ -14,12 +14,15 @@ pub enum RegistryValueData {
     QWord(u64),
 }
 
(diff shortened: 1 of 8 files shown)


registry/src/registry_helper.rs
```diff
@@ -67,82 +66,107 @@ impl RegistryHelper {
...
-                let (parent_key, subkeys) = self.get_valid_parent_key_and_subkeys()?;
-                let mut reg_key = parent_key;
-                for subkey in subkeys {
-                    let Ok(path) = UCString::<u16>::from_str(subkey) else {
-                        return Err(RegistryError::Utf16Conversion("subkey".to_string()));
-                    };
-
-                    reg_key = reg_key.create(path, Security::CreateSubKey)?;
+                if is_what_if {
+                    None
```
SteveL-MSFT, line 88: Do we want to return a message that this key would be created?  Maybe the messages should be an array of strings.

registry/src/main.rs
```diff
@@ -63,16 +63,21 @@ fn main() {
                         }
                     }
                 },
-                args::ConfigSubCommand::Set{input} => {
+                args::ConfigSubCommand::Set{input, what_if} => {
                     let reg_helper = match RegistryHelper::new(&input) {
```
SteveL-MSFT, line 67: Did you consider making `what_if` a member of `RegistryHelper` instead of passing it to the methods?

registry/src/config.rs
```diff
@@ -14,12 +14,15 @@ pub enum RegistryValueData {
...
-#[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize, Deserialize, JsonSchema)]
+#[derive(Debug, Default, Clone, PartialEq, Eq, Hash, Serialize, Deserialize, JsonSchema)]
 #[serde(rename = "Registry", deny_unknown_fields)]
 pub struct Registry {
     /// The path to the registry key.
     #[serde(rename = "keyPath")]
     pub key_path: String,
+    /// The information from a config set --what-if operation.
+    #[serde(flatten, skip_serializing_if = "Option::is_none")]
+    pub what_if: Option<WhatIf>,
```
SteveL-MSFT, line 25: I think for now, you should return this as:

```json
{
  "_metadata": {
    "whatIf": ["array of strings"]
  }
}
```

Then a follow-up PR to have the engine move `_metadata` member to the resource result `metadata`

SteveL-MSFT (changes requested):

registry/src/registry_helper.rs
```diff
@@ -215,6 +246,18 @@ impl RegistryHelper {
 
         Ok((parent_key, subkeys))
     }
+
+    fn handle_error_or_what_if(&self, error: RegistryError) -> Result<Option<Registry>, RegistryError> {
+        // TODO: return error message via metadata instead of as property within set state
```
SteveL-MSFT, line 251: this comment can be removed now, right?

registry/src/registry_helper.rs
```diff
@@ -3,17 +3,18 @@
...
 pub struct RegistryHelper {
     config: Registry,
     hive: Hive,
     subkey: String,
+    what_if: bool,
 }
 
 impl RegistryHelper {
-    pub fn new(config: &str) -> Result<Self, RegistryError> {
+    pub fn new(config: &str, what_if: bool) -> Result<Self, RegistryError> {
```
SteveL-MSFT, line 17: looking at the usage below, I think it would be better to NOT have this as part of the constructor, but instead have it default to `false` and a pub fn `enable_what_if()` that flips to `true`

SteveL-MSFT (changes requested):

SteveL-MSFT (approved):

PR #468, opened 2024-06-19
PSAdapter cache updates

anmenaga:
# PR Summary

1) Adding a parameter that users can use to clear PSAdapter cache:
`.\psDscAdapter\powershell.resource.ps1 -Operation ClearCache`

2) Adding a cache version and logic to rebuild the cache if version in the cache file does not match version in the running code. This will remove problems with updates to newer dsc versions.

### DIFF
--- a/powershell-adapter/Tests/powershellgroup.resource.tests.ps1
+++ b/powershell-adapter/Tests/powershellgroup.resource.tests.ps1
@@ -84,19 +84,19 @@ Describe 'PowerShell adapter resource tests' {
 
     It 'Verify that ClearCache works in PSAdapter' {
         # generate the cache
-        $null = dsc resource list * -a Microsoft.DSC/PowerShell
+        $null = dsc resource list '*' -a Microsoft.DSC/PowerShell
         # call the ClearCache operation
         $scriptPath = Join-Path $PSScriptRoot '..' 'psDscAdapter' 'powershell.resource.ps1'
-        $null = & $scriptPath -Operation ClearCache 2
+        $null = & $scriptPath -Operation ClearCache
         # verify that PSAdapter does not find the cache
-        dsc -l debug resource list * -a Microsoft.DSC/PowerShell 2> $TestDrive/tracing.txt
+        dsc -l debug resource list '*' -a Microsoft.DSC/PowerShell 2> $TestDrive/tracing.txt
         $LASTEXITCODE | Should -Be 0
         "$TestDrive/tracing.txt" | Should -FileContentMatchExactly 'Cache file not found'
     }
 
     It 'Verify that a new PS Cache version results in cache rebuid' {
         # generate the cache
-        $null = dsc resource list * -a Microsoft.DSC/PowerShell
+        $null = dsc resource list '*' -a Microsoft.DSC/PowerShell
         # update the version in the cache file
         $cacheFilePath = if ($IsWindows) {
             # PS 6+ on Windows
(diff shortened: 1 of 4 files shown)


powershell-adapter/psDscAdapter/win_psDscAdapter.psm1
```diff
@@ -68,47 +70,52 @@ function Invoke-DscCacheRefresh {
         "Reading from Get-DscResource cache file $cacheFilePath" | Write-DscTrace
 
         $cache = Get-Content -Raw $cacheFilePath | ConvertFrom-Json
-        $dscResourceCacheEntries = $cache.ResourceCache
-
-        if ($dscResourceCacheEntries.Count -eq 0) {
-            # if there is nothing in the cache file - refresh cache
+        if ($cache.CacheSchemaVersion -ne $script:CurrentCacheSchemaVersion) {
             $refreshCache = $true
+            "Incompartible version of cache in file '"+$cache.CacheSchemaVersion+"' (expected '"+$script:CurrentCacheSchemaVersion+"')" | Write-DscTrace
```
tgauth, line 75:
```suggestion
            "Incompatible version of cache in file '"+$cache.CacheSchemaVersion+"' (expected '"+$script:CurrentCacheSchemaVersion+"')" | Write-DscTrace
```

powershell-adapter/Tests/powershellgroup.resource.tests.ps1
```diff
@@ -81,4 +81,42 @@ Describe 'PowerShell adapter resource tests' {
...
+        } else {
+            # either WinPS or PS 6+ on Linux/Mac
+            if ($PSVersionTable.PSVersion.Major -le 5) {
+                Join-Path $env:LocalAppData "dsc\WindowsPSAdapterCache.json"
+            } else {
+                Join-Path $env:HOME ".dsc" "PSAdapterCache.json"
+            }
+        }
+        $cache = Get-Content -Raw $cacheFilePath | ConvertFrom-Json
+        $cache.CacheSchemaVersion = 0
```
SteveL-MSFT, line 113: would it be better to have the cache keep a timestamp of the last refresh to compare instead of the trace message?

powershell-adapter/psDscAdapter/psDscAdapter.psm1
```diff
@@ -1,6 +1,8 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT License.
 
+$script:CurrentCacheSchemaVersion = 1
```
SteveL-MSFT, line 4: did you consider using `[version]` type?

SteveL-MSFT (changes requested):

powershell-adapter/psDscAdapter/win_psDscAdapter.psm1
anmenaga: Fixed.

powershell-adapter/Tests/powershellgroup.resource.tests.ps1
anmenaga: When it comes to validation in a test case -  both methods are fine.
If you're talking about actually storing timestamp in the cache file - I'm not sure what value that brings for the product code, since main focus of this functionality is to easily mark and handle breaking changes in the schema;   btw, such timestamp already exists indirectly - LastWriteTime of the cache file.

powershell-adapter/psDscAdapter/psDscAdapter.psm1
anmenaga: Seems like an overkill for this simple case here;
Besides, the cache does not have major/minor/build concepts.

SteveL-MSFT:
@anmenaga please address the merge conflict

SteveL-MSFT (approved):

Issue #462, opened 2024-06-15
Remaining features of cross-platform PSAdapter


anmenaga:
### Prerequisites

- [X] Write a descriptive title.
- [X] Make sure you are able to repro it on the latest version
- [X] Search the existing issues.

### Summary

These features have to implemented:

- [ ] Class-inheritance
- [ ] Embedded types
- [ ] Enums `string<->number` conversions.

### Steps to reproduce

NA

### Expected behavior

```console
NA
```

### Actual behavior

```console
NA
```

### Environment data

```PowerShell
v1
```

### Version

v1

PR #469, opened 2024-06-19
PSAdapter support for class inheritance

anmenaga:
# PR Summary
This adds properties of base classes to class-based dsc resources.
This is part of the #462

### DIFF
--- a/powershell-adapter/psDscAdapter/psDscAdapter.psm1
+++ b/powershell-adapter/psDscAdapter/psDscAdapter.psm1
@@ -49,6 +49,56 @@ function Get-DSCResourceModules
     return $dscModulePsd1List
 }
 
+function Add-AstMembers {
+    param(
+        $AllTypeDefinitions,
+        $TypeAst,
+        $Properties
+    )
+
+    foreach($TypeConstraint in $TypeAst.BaseTypes) {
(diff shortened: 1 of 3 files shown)


powershell-adapter/psDscAdapter/psDscAdapter.psm1
```diff
@@ -49,6 +49,56 @@ function Get-DSCResourceModules
...
+        foreach($attr in $property.Attributes)
+        {
+            if ($attr.TypeName.Name -eq 'DscProperty')
+            {
+                $skipProperty = $false
+                foreach($attrArg in $attr.NamedArguments)
+                {
+                    if ($attrArg.ArgumentName -eq 'Key')
+                    {
+                        $isKeyProperty = $true
```
SteveL-MSFT, line 84: can't you break early after finding it's a key?

SteveL-MSFT (changes requested):
added comment

powershell-adapter/psDscAdapter/psDscAdapter.psm1
anmenaga: yes; added break.

SteveL-MSFT (approved):

PR #481, opened 2024-07-03
Detect if `dsc.exe` is started from MSStore or Explorer and show a message and wait for key press

SteveL-MSFT:
# PR Summary

As part of the certification process to publish to MSStore, the tester will install the app and click `Open` in the MSStore GUI and expect something to happen.  Since this is a command-line tool, it would not get required parameters, show help, and exit which to the tester looks like a crash so this prevents it from being published to the store publicly.

The change is to detect if the parent process is `sihost.exe` (name of the MSStore host that spawns apps) or `explorer.exe` and show a message directing them to https://aka.ms/dscv3-docs and wait for a key press before exiting.

To be able to test this, I had to create a new private version of the app so had to update `build.ps1` to build a private version of the msix.


tgauth (approved):

anmenaga:
Can you please explain more why new `$packageType`=`msix-private` was needed?

SteveL-MSFT:
> Can you please explain more why new `$packageType`=`msix-private` was needed?

Everytime I need to publish a test build to the Store, I need to increment the version number (you can't replace a published package even if it's not public).  So I needed a "private" product name in the Store to use for testing.

PR #482, opened 2024-07-03
Renamed field that specifies adapted type

anmenaga:
# PR Summary

Previously, when an adapter process is invoked (for the `dsc resource ...` command) the actual resource type name is added to the input JSON bag as a *special* `type` field.
This results in collisions if resource has an actual `type` property. (e.g. `PSDesiredStateConfiguration/File` resource)

This PR renames the *special* field in the json to `adapted_dsc_type` which is less likely to cause collisions.


SteveL-MSFT (changes requested):
Please add a test

anmenaga:
> Please add a test

This is already greatly covered by existing tests. Every non-List PSAdapter operation uses this code path.

SteveL-MSFT:
> > Please add a test
> 
> This is already greatly covered by existing tests. Every non-List PSAdapter (and other adapters) operation uses this code path.

But we don't have an explicit test to verify the behavior otherwise it would have failed before this change

anmenaga:
Added tests and a test adapter that verifies the renamed field.

SteveL-MSFT (approved):

PR #488, opened 2024-07-12
Fix input schema validation for RebootPending resource

### LINKED ISSUE #485: Pending reboot should have embedded schema for assertions

When working on DSC's material, I was exploring assertions in more depth. After adding the `Microsoft.Windows/RebootPending` configuration, I was hit with the following error:

image

I couldn't find anything directly in the docs about why this was, but after fiddling in the following schema into the JSON resource manifest:

```json
"schema": {
      "command": {
        "executable": "powershell",
        "args": [
          "-NoLogo",
          "-NonInteractive",
          "-NoProfile",
          "-Command",
          "reboot_pending.resource.ps1"
        ]
      }
    }
```

It returned the reboot pending state. I know the resource is old and might not be shipped in the next releases. However, if you want I can create the pull request, or just keep it in the book "as is".

_No response_

anmenaga:
# PR Summary

Fix #485 

When used in a configuration `Microsoft.Windows/RebootPending` was failing input schema validation.
This PR adds empty schema (because `RebootPending` resource expects no input) to resolve that issue.

@michaeltlombardi please let me know if this is not the right way to do this for resources that do not take input.


SteveL-MSFT:
@anmenaga I think the schema should also reflect the output so people can view the schema and know what to expect even if it takes no input

SteveL-MSFT (changes requested):
add static schema to manifest

anmenaga:
> add static schema to manifest

@SteveL-MSFT 
That is "schema/embedded/properties" node, right?
Updated.

reboot_pending/reboot_pending.dsc.resource.json
```diff
@@ -1,16 +1,27 @@
...
+      "-NonInteractive",
+      "-NoProfile",
+      "-Command",
+      "reboot_pending.resource.ps1"
+    ]
+  },
+  "schema": {
+    "embedded": {
+      "$schema": "https://json-schema.org/draft/2020-12/schema",
+      "type": "null",
```
SteveL-MSFT, line 19: I think type should be `object`

SteveL-MSFT (changes requested):

reboot_pending/reboot_pending.dsc.resource.json
anmenaga: With `object` I get 
`ERROR Error: Schema: null is not of type "object"`

michaeltlombardi: The existing schema we apply to embedded commands _requires_ that `type` always be object ([docs][01], [schema source][02]).

[01]: https://learn.microsoft.com/powershell/dsc/reference/schemas/resource/manifest/schema/embedded?view=dsc-3.0&preserveView=true#type
[02]: https://github.com/PowerShell/DSC/blob/6483333b9ef689e795761d17428d2ff3ff2768a3/schemas/src/resource/manifest.schema.yaml#L114-L125

We could change this to allow users to specify the valid options as `object`, `null`, or `[object, null]` but I think it probably easier on resource authors to hand over an empty object when `properties` doesn't define any key-value pairs.

michaeltlombardi: Actually, the type can't be just `null`, it needs to be either `object` or `[object, null]` - otherwise defining the properties keyword in the schema means nothing, and the schema effectively can't be used to validate the output.

anmenaga: @michaeltlombardi so what should be the `type` in this bug fix?

SteveL-MSFT: It should be `[object, null]`

anmenaga: > It should be `[object, null]`

This fails with:
`Error: Schema: "[object, null]" is not valid under any of the schemas listed in the 'anyOf' keyword`

michaeltlombardi: We may have to also update the embedded schema definition for this to work. That would mean editing this line:

https://github.com/PowerShell/DSC/blob/6483333b9ef689e795761d17428d2ff3ff2768a3/schemas/src/resource/manifest.schema.yaml#L125

to:

```yaml
enum:
- object
- [object, 'null']
```

I _think_. I would need to do some quick local testing.

I'm still not confident we _should_ support passing `null` to the instance instead of an empty object - right now, an empty object would still be valid, because there are no required properties.

A schema with no required properties can receive an empty object and still be valid.

I can spin up a local change of the schema and submit a PR updating it if we definitely want to support both `object` and `[object, 'null']` for the valid types.

reboot_pending/reboot_pending.dsc.resource.json
```diff
@@ -1,16 +1,27 @@
...
+      "properties": {
+        "RebootPending": {
+            "type": "boolean"
```
michaeltlombardi, line 22: Since this property isn't configurable, it should be marked read-only - it is returned from the resource but not sent to it.

```suggestion
            "type": "boolean",
            "readOnly": true
```

reboot_pending/reboot_pending.dsc.resource.json
```diff
@@ -1,16 +1,27 @@
...
+      "-Command",
+      "reboot_pending.resource.ps1"
+    ]
+  },
+  "schema": {
+    "embedded": {
+      "$schema": "https://json-schema.org/draft/2020-12/schema",
+      "type": "null",
+      "properties": {
+        "RebootPending": {
```
michaeltlombardi, line 21: For consistency with other resources, should we use camelCasing for this resource?

SteveL-MSFT: Yes, for any new resource, we should use camelCasing

reboot_pending/reboot_pending.dsc.resource.json
SteveL-MSFT: I'm ok with an empty object vs null as it implies to me reading a config something deliberate vs accidental

anmenaga: ok, so what should the schema look like in this case?

reboot_pending/reboot_pending.dsc.resource.json
anmenaga: Updated.

reboot_pending/reboot_pending.dsc.resource.json
anmenaga: Updated the property to be `rebootPending` throughout this resource code.

reboot_pending/reboot_pending.dsc.resource.json
michaeltlombardi: The `type` should be `object` - you mentioned earlier that you got the error `ERROR Error: Schema: null is not of type "object"` - was that from a configuration command or direct resource invocation?

anmenaga: config command:
`dsc config get --path C:\temp\reboot_pending.yml`

SteveL-MSFT: @anmenaga can you add an example yaml for rebootPending?

SteveL-MSFT (changes requested):

reboot_pending/reboot_pending.dsc.resource.json
anmenaga: Added  example yaml and a test that uses it.

SteveL-MSFT (approved):

PR #499, opened 2024-07-24
Add DSC home directory to PATH to find included resources

### LINKED ISSUE #494: `dsc` should add it's home directory to `PATH` to find included resources
### Prerequisites

### Summary

`dsc` will look it it's home directory for resource manifests, but then fails to find the executable since it's not actually in `PATH`

### Steps to reproduce

1. build.ps1
2. ./bin/debug/dsc -l trace config test -p ./dsc/examples/osinfo_parameters.dsc.yaml

### Expected behavior

```console
This should succeed
```

### Actual behavior

```console
2024-07-18T20:23:02.440983Z ERROR dsc::subcommand: 137: Error: IO: No such file or directory (os error 2)
```

This is because although the `osinfo` resource is found, the `osinfo` executable isn't found.

`dsc` should add it's home directory to `PATH` (if not already there, probably at the end) so that the included resource exes get found on startup.

Discovery should be updated to not look in home directory and just use the new `PATH`

### Error details

_No response_

### Environment data

```PowerShell
PSVersion                      7.5.0-preview.3
GitCommitId                    7.5.0-preview.3
OS                             Darwin 23.5.0 Darwin Kernel Version 23.5.0: Wed May  1 20:16:51 PDT 2024; root:xnu-10063.121.3~5/RELEASE_ARM64_T8103
Platform                       Unix
```

### Version

dsc 3.0.0-preview.8

### Visuals

_No response_

anmenaga:
# PR Summary

Fix #494 

As described in the issue, after this fix DSC will update its process PATH env var to have path to dsc.exe (unless it is already there), so that included resources manifests and executables are found.
A side effect of this change is that resource processes can assume that PATH env var that they are started with contains DSC home directory.

Note: the above behavior will not take place if `DSC_RESOURCE_PATH` env var is defined.

### DIFF
--- a/dsc/tests/dsc_osinfo.tests.ps1
+++ b/dsc/tests/dsc_osinfo.tests.ps1
@@ -18,4 +18,20 @@ Describe 'Tests for osinfo examples' {
         $LASTEXITCODE | Should -Be 0
         $out.results[0].result.inDesiredState | Should -Be $IsMacOS
     }
+
+    It 'Verify dsc home directory is added to PATH to find included resources' {
+        $oldPath = $env:PATH
+        try {
+            $exe_path = Resolve-Path "$PSScriptRoot/../../bin/debug"
+            # Remove unwanted elements
+            $new_path = ($oldPath.Split([System.IO.Path]::PathSeparator) | Where-Object { $_ -ne $exe_path }) -join [System.IO.Path]::PathSeparator
+            $env:PATH = $new_path
+
+            $null = & "$PSScriptRoot/../../bin/debug/dsc" config test -p "$PSScriptRoot/../examples/osinfo_parameters.dsc.yaml"
+            $LASTEXITCODE | Should -Be 0
+        }
+        finally {
+            $env:PATH = $oldPath
+        }
+    }
 }
--- a/dsc/tests/dsc_osinfo.tests.ps1
+++ b/dsc/tests/dsc_osinfo.tests.ps1
@@ -19,19 +19,22 @@ Describe 'Tests for osinfo examples' {
         $out.results[0].result.inDesiredState | Should -Be $IsMacOS
     }
 
-    It 'Verify dsc home directory is added to PATH to find included resources' {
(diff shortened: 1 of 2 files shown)


dsc/tests/dsc_osinfo.tests.ps1
```diff
@@ -18,4 +18,20 @@ Describe 'Tests for osinfo examples' {
         $LASTEXITCODE | Should -Be 0
         $out.results[0].result.inDesiredState | Should -Be $IsMacOS
     }
+
+    It 'Verify dsc home directory is added to PATH to find included resources' {
+        $oldPath = $env:PATH
+        try {
+            $exe_path = Resolve-Path "$PSScriptRoot/../../bin/debug"
```
SteveL-MSFT, line 25: I don't think you should rely on it being in `debug` folder.  Would be better to get the path to `dsc` and `set-location` to the parent folder then resolve relative path to `dsc` to execute it

dsc_lib/src/discovery/command_discovery.rs
```diff
@@ -58,18 +58,27 @@ impl CommandDiscovery {
         };
 
         let mut paths = env::split_paths(&path_env).collect::<Vec<_>>();
+        // remove duplicate entries
+        let mut uniques = HashSet::new();
+        paths.retain(|e|uniques.insert((*e).clone()));
```
SteveL-MSFT, line 63: shouldn't we remove dupes after any modifications in case PATH already includes dsc home?

SteveL-MSFT (changes requested):

dsc_lib/src/discovery/command_discovery.rs
anmenaga: If it is already in the path then `if paths.contains(&exe_home_pb)` will be `true` and no modifications will be made.

dsc/tests/dsc_osinfo.tests.ps1
anmenaga: Good point; updated.

SteveL-MSFT (approved):

PR #507, opened 2024-07-31
[WIP] Microsoft.PowerShell.DSC module

Gijsreyn:
# PR Summary

As mentioned in the previous PR at: #498, the name might be too ambitious, but I thought to give it a shot.

The goal of the module is to help PowerShell users write DSC V3 configuration documents and resources in a PowerShell-ish way. This helps users adopt DSC V3 concepts by relating _known_ commands on the shell to DSCV1/V2 concepts.

Features to include:

- Argument completers for PSDSC DSC resources. It should be able to output PowerShell objects that in the end are sent through the pipeline to `dsc.exe` as JSON.
- Provide easy-to-use examples from the get-go.
- Able to execute both `.json`, `.yaml`, or `.ps1` files.
- Leverage functionality from adapters e.g. PowerShell/WindowsPowerShell
- ...

From this perspective in PS module format, it should be able to be checked for multiple PowerShell versions thanks to @gaelcolas and other contributors to the Sampler project. It can also be published easily in PowerShellGallery.


Gijsreyn:
@SteveL-MSFT Sorry was messing around with the branches, that's why I opened a new PR. Can you close #498?

@michaeltlombardi Can you skip the PR review on the previous one and when you have time, check this one out, please?

SteveL-MSFT:
@Gijsreyn I appreciate your efforts and enthusiasm on this.  However, before you spend more time writing code, I think it would be best if we started a bit slower and agree on the goals and requirements, then the design, and then the code should follow.  Note that prototype code for the purpose of evaluating feasibility and usability makes sense with the expectation that prototypes don't get turned into production code and prototype code is only intended for learning.

So perhaps we need to take a step back and just have a discussion starting with the intended goals of this project?

Gijsreyn:
@SteveL-MSFT I agree. My enthusiasm took over too hard as my idea train was spinning through me. I forgot to align more on the goals and requirements and only took my own learnings into the idea when starting in the DSC V3 era. What you said about prototype code, I totally get it. As I'm quite new to the community, would it be great to pitch such prototype code at the PowerShell / PowerShell DSC community calls to gather traction and get more ideas flowing? 

I'll take a step back for now and love to have further discussions on it. Best to do it here or on the discussions tab of this project? To keep things straightforward, feel free to point out any mismatches in vision or ideas in this project context if the module does not fit.

Thanks again for the quick response Steve.

SteveL-MSFT:
@Gijsreyn it would be great if you want to use the next Community Call to give a short demo of your prototype and point them to a discussion to solicit more feedback/ideas

Gijsreyn:
Cool Steve, I will try to prepare a small demo for the community call on the 15th. I cannot attend the 21st of August because I will be going on holiday. Do I have to get in touch with Steven or Sydney?

SteveL-MSFT:
@Gijsreyn just add yourself to the agenda here: https://github.com/PowerShell/PowerShell/discussions/24123

Issue #510, opened 2024-08-03
Duplicates allowed for parameters and values (or any HashMap)


SteveL-MSFT:
### Prerequisites

- [X] Write a descriptive title.
- [X] Make sure you are able to repro it on the latest version
- [X] Search the existing issues.

### Summary

Due to default `serde` behavior, duplicates in deserializing a `HashMap` is not an error and last one wins.  There is another crate `serde_with` that has a [macro](https://docs.rs/serde_with/latest/serde_with/rust/maps_duplicate_key_is_error/index.html) that prevents dupes, but it doesn't work if the property is an `Option<>`.  Tried a few ways to resolve this, but couldn't find a solutiont hat doesn't include implementing a custom deserializer.  Deferring this work.

### Steps to reproduce

```powershell
$configYaml = @'
$schema: https://raw.githubusercontent.com/PowerShell/DSC/main/schemas/2024/04/config/document.json
variables:
  myVariable: foo
  myVariable: bar
resources:
- name: test
  type: Test/Echo
  properties:
    output: "[variables('myVariable')]"
'@

dsc config get -d $configYaml
```

### Expected behavior

```console
Should error that `myVariable` is defined more than once
```

### Actual behavior

```console
`myVariable` takes the last set value
```

### Environment data

```PowerShell
Name                           Value
----                           -----
PSVersion                      7.5.0-preview.3
PSEdition                      Core
GitCommitId                    7.5.0-preview.3
OS                             Darwin 23.5.0 Darwin Kernel Version 23.5.0: Wed May  1 20:16:51 PDT 2024; root:xnu-10063.121.3~5/RELEASE_ARM64_T8103
Platform                       Unix
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0…}
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
WSManStackVersion              3.0
```

### Version

dsc 3.0.0-preview.8

michaeltlombardi:
This could be resolved okay with a note in the documentation - fwiw, users who are defining their configuration documents in VS Code with the YAML extension will get author-time feedback that an object can't define the same property twice, so even though DSC isn't currently raising an error, the user has _some_ way to know that defining the same key twice in an object isn't supported.

SteveL-MSFT:
[...]

Issue #509, opened 2024-08-03
Expressions is missing array index support


SteveL-MSFT:
### Summary of the new feature / enhancement

This should work where `variables('foo')` returns an array:

```
[variables('foo')[1]]
```

Also need to support multi-dimensional arrays.  This was found missing while developing `variables()` support. 

### Proposed technical implementation details (optional)

The tree-sitter grammar needs to be updated along with the parser.

PR #511, opened 2024-08-03
Add variables support

SteveL-MSFT:
# PR Summary

Add ability to have `variables` section in config and `variables()` function implementation.  Note that due to how serde works by default, duplicates don't error and last one wins (created https://github.com/PowerShell/DSC/issues/510)

Also added some additional debug tracing to parser to understand why array indexing didn't work.  Turns out I missed implementing it https://github.com/PowerShell/DSC/issues/509

## PR Context

Part of https://github.com/PowerShell/DSC/issues/57

### DIFF
--- a/dsc/examples/variables.dsc.yaml
+++ b/dsc/examples/variables.dsc.yaml
@@ -1,0 +1,15 @@
+$schema: https://raw.githubusercontent.com/PowerShell/DSC/main/schemas/2024/04/config/document.json
+parameters:
+  myParameter:
+    type: string
+    # the use of `concat()` here is just an example of using an expression for a defaultValue
+    defaultValue: "[concat('world','!')]"
+variables:
+  myOutput: "[concat('Hello ', parameters('myParameter'))]"
+  myObject:
+    test: baz
+resources:
+- name: test
+  type: Test/Echo
+  properties:
+    output: "[concat('myOutput is: ', variables('myOutput'), ', myObject is: ', variables('myObject').test)]"
--- a/dsc/tests/dsc_variables.tests.ps1
+++ b/dsc/tests/dsc_variables.tests.ps1
@@ -1,0 +1,46 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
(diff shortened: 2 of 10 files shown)


dsc/examples/variables.dsc.yaml
```diff
@@ -0,0 +1,15 @@
+$schema: https://raw.githubusercontent.com/PowerShell/DSC/main/schemas/2024/04/config/document.json
+parameters:
+  myParameter:
+    type: string
+    # the use of `concat()` here is just an example of using an expression for a defaultValue
+    defaultValue: "[concat('world','!')]"
```
anmenaga, line 6: There is a difference between parameters and variables:
parameters have 2 subfields: `type` and `defaultValue`,
but looks like variables have just implied `defaultValue` and type is implicitly derived.
Should this be consistent?

dsc/tests/dsc_variables.tests.ps1
```diff
@@ -0,0 +1,46 @@
...
+  properties:
+    output: "[variables('myVariable')]"
+'@
+        $out = dsc config get -d $configYaml | ConvertFrom-Json
+        Write-Verbose -Verbose $out
+        $LASTEXITCODE | Should -Be 0
+        $out.results[0].result.actualState.output | Should -Be 'bar'
+    }
+
+    It 'Missing variable returns error' {
```
anmenaga, line 30: The essence of a `variable` in any language is that it's value can be changed; I don't see such example or tests.

anmenaga (approved):

dsc/examples/variables.dsc.yaml
SteveL-MSFT: I took this directly from ARM.  `variables` are intended to just be used to keep values that get reused multiple times within the config and typically constructed from params based on examples I saw.

dsc/tests/dsc_variables.tests.ps1
SteveL-MSFT: You are correct, but that's not how it's used in an ARM template

PR #514, opened 2024-08-04
Add array index support to expression grammar

SteveL-MSFT:
# PR Summary

Add array index syntax support to expressions.

- tree-sitter grammar updated which includes removing `bracketInStringLiteral` as that is no longer valid as array index results in a bracket within an expression
- as part of tree-sitter-grammar change, collapse `member` and `index` as accessors that follow a function
- added more tests to tree-sitter corpus which also needed to be moved to under a `test` folder to be discovered correctly now
- the DSC parser updated to use the new tree-sitter output and also added handling for errors following a valid statement

Fix https://github.com/PowerShell/DSC/issues/509


dsc/tests/dsc_expressions.tests.ps1
```diff
@@ -0,0 +1,61 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+Describe 'Expressions tests' {
+    It 'Accessors work: <text>' -TestCases @(
+        @{ text = "[parameters('test').hello]"; expected = '@{world=there}' }
+        @{ text = "[parameters('test').hello.world]"; expected = 'there' }
```
tgauth, line 7: I see this covers the member accessor case - is the index accessor case also covered?

dsc_lib/src/parser/expressions.rs
```diff
@@ -32,64 +38,120 @@ impl Expression {
...
-                if let Some(object) = value.as_object() {
-                    if !object.contains_key(member) {
-                        return Err(DscError::Parser(format!("Member '{member}' not found")));
-                    }
-
-                    value = object[member].clone();
+            for accessor in &self.accessors {
+                match accessor {
+                    Accessor::Member(member) => {
+                        if !value.is_object() {
```
tgauth, line 123: Would it be less redundant to handle the non-object case as part the `as_object()` cast?

dsc_lib/src/parser/expressions.rs
```diff
@@ -32,64 +38,120 @@ impl Expression {
...
+                        }
+                        if let Some(object) = value.as_object() {
+                            if !object.contains_key(member) {
+                                return Err(DscError::Parser(format!("Member '{member}' not found")));
+                            }
+                            value = object[member].clone();
+                        }
+                    },
+                    Accessor::Index(index) => {
+                        if !value.is_array() {
```
tgauth, line 134: same comment as `as_object()` above

dsc/tests/dsc_expressions.tests.ps1
SteveL-MSFT: good catch, looks like I setup the parameters for it, but forgot to add in the test cases

dsc_lib/src/parser/expressions.rs
SteveL-MSFT: yes, I think it can just be an `else` statement

dsc_lib/src/parser/expressions.rs
```diff
@@ -32,64 +38,118 @@ impl Expression {
...
+                                return Err(DscError::Parser(format!("Member '{member}' not found")));
+                            }
+                            value = object[member].clone();
+                        } else {
+                            return Err(DscError::Parser("Member access on non-object value".to_string()));
+                        }
+                    },
+                    Accessor::Index(index) => {
+                        if let Some(array) = value.as_array() {
+                            if !index.is_number() {
```
tgauth, line 134: another extra check here with `is_number()`

tgauth (approved):

PR #517, opened 2024-08-07
Add error message for PowerShell adapter without input

Gijsreyn:
# PR Summary

The PowerShell adapter requires a check whenever input is empty, it should return a proper error message. In this case, the default @{} was given as a parameter without converting it to JSON. Now it checks the JSON input and if empty, converts it still to an empty JSON string.


Gijsreyn:
@anmenaga Can you review this PR?

powershell-adapter/psDscAdapter/powershell.resource.ps1
```diff
@@ -66,6 +66,11 @@ if ($jsonInput) {
     if ($jsonInput -ne '@{}') {
         $inputobj_pscustomobj = $jsonInput | ConvertFrom-Json
     }
+    else 
+    {
+        $jsonInput = $jsonInput | ConvertTo-Json
```
anmenaga, line 71: Can you please give examples of `expected behavior` and `actual behavior`?

anmenaga (changes requested):

powershell-adapter/psDscAdapter/powershell.resource.ps1
Gijsreyn: Whenever a user tries to do a get operation on a resource without specifying any input, it should give a proper error message to the user. That's it actually. If it can be captured higher level, that would also do the trick.

anmenaga:
A generic fix for this issue is in #525 .

PR #530, opened 2024-08-24
Adapter lookup table for non-list operations on adapted resources

anmenaga:
# PR Summary

This is a performance optimization that reduces time when searching for the adapter that implements a requested resource when non-list operations on this resource are invoked.

When a non-list operation on an adapted resource is invoked (e.g. `dsc resource get -r Test/TestCase`), discovery code first is doing `List` operation sequentially on all adapters until it finds which adapter implements requested resource.
If the adapter of interest happens to be in the end of line, then a huge perf cost is accumulated.
This change adds an adapter lookup table that is basically a mapping `<resource type>` -> `<was last seen in this adapter>`.
This lookup table is then used for sorting adapter list before doing `List` operations on them, this way the adapter of interest is searched first, reducing overall time of operation.

Adapter lookup file location:

- on Windows:  `$env:LocalAppData+"dsc\AdaptedResourcesLookupTable.json"`
- on Linux and Mac:  `$env:HOME+".dsc/AdaptedResourcesLookupTable.json"`

## PR Context
Before change:
```
PS C:\DSCv3> # 'Test/TestCase' resource is implemented in 'Test/TestAdapter'
PS C:\DSCv3> # Search order ('List' operation) without lookup table: `Microsoft.DSC/PowerShell`, `Microsoft.Windows/WindowsPowerShell`, `Microsoft.Windows/WMI`, `Test/TestAdapter`(needed resource found here)
PS C:\DSCv3> (Measure-Command {dsc resource get -r Test/TestCase}).TotalSeconds
41.8350422
```
After change:
```
PS C:\DSCv3> # 'Test/TestCase' resource is implemented in 'Test/TestAdapter'
PS C:\DSCv3> # Search order ('List' operation) based on lookup table: `Test/TestAdapter`(needed resource found here)
PS C:\DSCv3> (Measure-Command {dsc resource get -r Test/TestCase}).TotalSeconds
2.0769161
```


dsc_lib/src/discovery/command_discovery.rs
```diff
@@ -496,3 +507,96 @@ fn load_manifest(path: &Path) -> Result<DscResource, DscError> {
...
+
+fn sort_adapters_based_on_lookup_table(unsorted_adapters: &BTreeMap<String, Vec<DscResource>>, needed_resource_types: &Vec<String>) -> LinkedHashMap<String, Vec<DscResource>>
+{
+    let mut result:LinkedHashMap<String, Vec<DscResource>> = LinkedHashMap::new();
+    let lookup_table:HashMap<String, String> = load_adapted_resources_lookup_table();
+    // first add adapters (for needed types) that can be found in the lookup table
+    for needed_resource in needed_resource_types {
+        if let Some(adapter_name) = lookup_table.get(needed_resource) {
+            if let Some(resource_vec) = unsorted_adapters.get(adapter_name) {
+                trace!("Lookup table found resource '{}' in adapter '{}'", needed_resource, adapter_name);
```
SteveL-MSFT, line 519: We should use `debug!` more for traces that don't contain private data like this line

dsc_lib/src/discovery/command_discovery.rs
```diff
@@ -496,3 +507,96 @@ fn load_manifest(path: &Path) -> Result<DscResource, DscError> {
...
+        let adapter_name = &res_vec[0].require_adapter.as_ref().unwrap();
+        lookup_table.insert(resource_name.to_string().to_lowercase(), (*adapter_name).to_string());
+    };
+
+    save_adapted_resources_lookup_table(&lookup_table);
+}
+
+fn save_adapted_resources_lookup_table(lookup_table: &HashMap<String, String>)
+{
+    if let Ok(lookup_table_json) = serde_json::to_string_pretty(&lookup_table) {
```
SteveL-MSFT, line 549: We should not save this as pretty since it probably has a minor perf impact to write and read.  When debugging, we can use other tools to pretty print it.

SteveL-MSFT (changes requested):

dsc/tests/dsc_discovery.tests.ps1
```diff
@@ -96,4 +96,56 @@ Describe 'tests for resource discovery' {
             $env:DSC_RESOURCE_PATH = $oldPath
         }
     }
+
+    It 'Ensure List operation populates adapter lookup table' {
+        $lookupTableFilePath = if ($IsWindows) {
+            Join-Path $env:LocalAppData "dsc\AdaptedResourcesLookupTable.json"
+        } else {
+            Join-Path $env:HOME ".dsc" "AdaptedResourcesLookupTable.json"
+        }
```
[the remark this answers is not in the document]
anmenaga: Fixed.

dsc/tests/dsc_discovery.tests.ps1
```diff
@@ -96,4 +96,56 @@ Describe 'tests for resource discovery' {
...
+        Test-Path $lookupTableFilePath -PathType Leaf | Should -BeFalse
+
+        # perform Get on an adapter - this should create adapter lookup table file
+        $oldPSModulePath = $env:PSModulePath
+        $TestClassResourcePath = Resolve-Path "$PSScriptRoot/../../powershell-adapter/Tests"
+        $env:DSC_RESOURCE_PATH = $null
+        $env:PSModulePath += [System.IO.Path]::PathSeparator + $TestClassResourcePath
+        "{'Name':'TestClassResource1'}" | dsc resource get -r 'TestClassResource/TestClassResource' | Out-Null
+
+        Test-Path $lookupTableFilePath -PathType Leaf | Should -BeTrue
```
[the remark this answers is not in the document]
anmenaga: Added a rudimentary check.

dsc_lib/src/discovery/command_discovery.rs
anmenaga: Change `to_string_pretty` -> `to_string`.

dsc_lib/src/discovery/command_discovery.rs
```diff
@@ -496,3 +507,96 @@ fn load_manifest(path: &Path) -> Result<DscResource, DscError> {
...
+    result
+}
+
+fn add_resources_to_lookup_table(adapted_resources: &BTreeMap<String, Vec<DscResource>>)
+{
+    let mut lookup_table:HashMap<String, String> = load_adapted_resources_lookup_table();
+
+    for (resource_name, res_vec) in adapted_resources {
+        let adapter_name = &res_vec[0].require_adapter.as_ref().unwrap();
+        lookup_table.insert(resource_name.to_string().to_lowercase(), (*adapter_name).to_string());
```
SteveL-MSFT, line 541: Seems odd to get a ref only to de-ref it and copy to a string.  Would it be better on line 540 to just have `adapter_name` be a string?  Also is there a risk the `unwrap()` can fault?

dsc_lib/src/discovery/command_discovery.rs
```diff
@@ -496,3 +507,96 @@ fn load_manifest(path: &Path) -> Result<DscResource, DscError> {
...
+    } else {
+        debug!("Unable to serialize lookup_table to json");
+    }
+}
+
+fn load_adapted_resources_lookup_table() -> HashMap<String, String>
+{
+    let file_path = get_lookup_table_file_path();
+    
+    let lookup_table: HashMap<String, String> = match fs::read(file_path.clone()){
```
SteveL-MSFT, line 574: looks like `fs::read()` takes a ref, so you don't need to clone?

SteveL-MSFT (changes requested):

dsc_lib/src/discovery/command_discovery.rs
```diff
@@ -496,3 +507,96 @@ fn load_manifest(path: &Path) -> Result<DscResource, DscError> {
 
     Ok(resource)
 }
+
+fn sort_adapters_based_on_lookup_table(unsorted_adapters: &BTreeMap<String, Vec<DscResource>>, needed_resource_types: &Vec<String>) -> LinkedHashMap<String, Vec<DscResource>>
+{
+    let mut result:LinkedHashMap<String, Vec<DscResource>> = LinkedHashMap::new();
```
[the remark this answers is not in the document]
anmenaga: updated.

dsc_lib/src/discovery/command_discovery.rs
```diff
@@ -496,3 +507,96 @@ fn load_manifest(path: &Path) -> Result<DscResource, DscError> {
...
+    let lookup_table:HashMap<String, String> = load_adapted_resources_lookup_table();
```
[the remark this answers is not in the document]
anmenaga: updated.

dsc_lib/src/discovery/command_discovery.rs
```diff
@@ -496,3 +507,96 @@ fn load_manifest(path: &Path) -> Result<DscResource, DscError> {
...
+            result.insert(adapter_name.to_string(), adapters.clone());
+        }
+    }
+
+    result
+}
+
+fn add_resources_to_lookup_table(adapted_resources: &BTreeMap<String, Vec<DscResource>>)
+{
+    let mut lookup_table:HashMap<String, String> = load_adapted_resources_lookup_table();
```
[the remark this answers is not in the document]
anmenaga: updated.

dsc_lib/src/discovery/command_discovery.rs
anmenaga: Compiler says that borrowing is required:
`error[E0507]: cannot move out of index of std::vec::Vec<DscResource>.`

I rewrote it to handle unlikely `unwrap` fault.

dsc_lib/src/discovery/command_discovery.rs
anmenaga: Without `clone` the debug statement later on (line 586) throws `value borrowed here after move`.

dsc_lib/src/discovery/command_discovery.rs
```diff
@@ -496,3 +507,95 @@ fn load_manifest(path: &Path) -> Result<DscResource, DscError> {
...
+                    debug!("Unable to write lookup_table file {file_path:?}");
+                }
+            } else {
+                debug!("Unable to create parent directories of the lookup_table file {file_path:?}");
+            }
+        } else {
+            debug!("Unable to get directory of the lookup_table file {file_path:?}");
+        }
+    } else {
+        debug!("Unable to serialize lookup_table to json");
```
SteveL-MSFT, line 569: I think these should probably be `info!` instead of `debug!`.  Basically if it's something that can affect the end user, but not specifically a developer, we should use `info!`

SteveL-MSFT (changes requested):

dsc_lib/src/discovery/command_discovery.rs
anmenaga: Updated.

SteveL-MSFT (approved):

PR #533, opened 2024-08-28
Fix DateTime comparison in PSAdapter cache code

anmenaga:
# PR Summary
Previous code was using fractional seconds that resulted in unnecessary cache rebuilds.
This also adds tests that verify that there are no unnecessary cache rebuilds for both PSAdapter and WinPSAdapter.


powershell-adapter/Tests/powershellgroup.resource.tests.ps1
```diff
@@ -280,4 +280,31 @@ Describe 'PowerShell adapter resource tests' {
 
         $env:TestClassResourceResultCount = $null
     }
+
+    It 'Verify that there are no cache rebuids for several sequential executions' {
```
SteveL-MSFT, line 284:
```suggestion
    It 'Verify that there are no cache rebuilds for several sequential executions' {
```

powershell-adapter/Tests/win_powershellgroup.tests.ps1
```diff
@@ -70,4 +70,21 @@ Describe 'WindowsPowerShell adapter resource tests' {
       $res = $r | ConvertFrom-Json
       $res.results[0].result.actualState.result[0].properties.DestinationPath | Should -Be "$testFile"
     }
+
+    It 'Verify that there are no cache rebuids for several sequential executions' -Skip:(!$IsWindows) {
+
+        # remove cache file
+        $cacheFilePath = Join-Path $env:LocalAppData "dsc\WindowsPSAdapterCache.json"
+        Remove-Item -Force -Path $cacheFilePath -ErrorAction SilentlyContinue
```
SteveL-MSFT, line 78:
```suggestion
        Remove-Item -Force -Path $cacheFilePath -ErrorAction Ignore
```

SteveL-MSFT (changes requested):
I don't quite understand.  Even if using fractional seconds, there would have to be a change for the `LastWriteTime` to differ, right?

SteveL-MSFT (approved):

Issue #512, opened 2024-08-04
Need to support getting traces from child `dsc` processes


SteveL-MSFT:
### Summary of the new feature / enhancement

Need to propagate the current tracing level to child resources and accept JSON tracing from child `dsc` resources.

DSC currently uses the `tracing` crate JSON format which differs from the simplified format we support for other types of resources.

This will greatly improve diagnosability when things fail using things like the `group` resource since currently those traces are hidden.

### Proposed technical implementation details (optional)

Once https://github.com/PowerShell/DSC/pull/511 is merged, we can set the tracing level as part of the `context` and define a new argKind that passes the tracing level as an argument which the `dsc` based resources can accept.

We can then have `dsc` accept either the `tracing` JSON format in addition to the simplified format, or add an option to `dsc` to emit JSON traces using the simplified format.  The benefit of accepting the `tracing` format is that any resources written in Rust can just emit that.

michaeltlombardi:
The alternatives that occur to me for passing the trace level to the resource are:

1. To reuse the `DSC_TRACE_LEVEL` environment variable.

   I think this is the simplest option, but there's no way to have a resource indicate whether it supports tracing unless we add another property to the manifest.
1. To reuse the metadata pattern for passing information.

   This would, I think, be the more complex route, but worth considering for enabling resource authors and integrating tools to pass and retrieve metadata for resources.
1. To define a new well-known property.

   I prefer this option as the more explicit and less complicated way to communicate with the resource than metadata and would enable per-resource tracing (sometimes when tracing a config operation, I don't want to have to sift through a dozen resource outputs when running down a specific small behavior).
   
   The primary downside I see to this is that we can't (currently) cascade values to resources in a configuration document. Configuration authors would either need to use a group resource that cascades the value or to explicitly define/reference it with each resource. I suppose the default value of the property could be the value of `DSC_TRACE_LEVEL`, which might be enough?

I don't think defining a new `argKind` would be a bad idea, either.

SteveL-MSFT:
I did consider the env var route as it seemed the most obvious, but didn't want to introduce a new way to pass info to resources (although `DSC_CONFIG_ROOT` env var I guess already does that...)

SteveL-MSFT:
[...]

PR #541, opened 2024-09-07
Add `pass-through` trace format to `dsc` so that child-processes get traced appropriately by top process

SteveL-MSFT:
# PR Summary

Previously traces were rendered by child `dsc` processes which polluted the output and trace level didn't match what the user requested as they would be written at `trace` level only.

Change is to add a `pass-through` format which also makes child `dsc` not trace the output but instead write the JSON to stderr.

The top most `dsc` process will deserialize the JSON and write the traces with the appropriate trace level and also include the child process id to get a better idea of the process tree.

Also update the tree-sitter build script to not do a clean install of node modules unless running in ADO as it's way too slow.

Example with just INFO level:

image

## Context

Fix https://github.com/PowerShell/DSC/issues/512

### DIFF
--- a/dsc_lib/src/dscresources/command_resource.rs
+++ b/dsc_lib/src/dscresources/command_resource.rs
@@ -1,6 +1,7 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
+use clap::ValueEnum;
 use jsonschema::JSONSchema;
 use serde::Deserialize;
 use serde_json::Value;
@@ -827,15 +828,15 @@ pub fn log_stderr_line<'a>(process_id: &u32, trace_line: &'a str) -> &'a str
                 0
             };
             let trace_message = if include_target {
-                format!("Process {process_id}: {target}: {line_number}: {}", trace_object.fields.message)
+                format!("PID {process_id}: {target}: {line_number}: {}", trace_object.fields.message)
             } else {
-                format!("Process {process_id}: {}", trace_object.fields.message)
+                format!("PID {process_id}: {}", trace_object.fields.message)
             };
             match trace_object.level {
                 TraceLevel::Error => {
                     error!(trace_message);
                 },
-                TraceLevel::Warning => {
+                TraceLevel::Warn => {
                     warn!(trace_message);
                 },
                 TraceLevel::Info => {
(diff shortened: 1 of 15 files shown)


tgauth (approved):

dsc_lib/src/dscresources/command_resource.rs
```diff
@@ -815,26 +816,94 @@ pub fn log_stderr_line<'a>(process_id: &u32, trace_line: &'a str) -> &'a str
...
+            let target = if let Some(t) = trace_object.target.as_deref() { t } else {
+                include_target = false;
+                ""
+            };
+            let line_number = if let Some(l) = trace_object.line_number { l } else {
+                include_target = false;
+                0
+            };
+            let trace_message = if include_target {
+                format!("Process {process_id}: {target}: {line_number}: {}", trace_object.fields.message)
```
anmenaga, line 830: I feel like we should get rid of the `Process ` word in all traces - it takes too much real estate in the output.
Maybe just have `process_id` in square brackets instead.

dsc_lib/src/dscresources/command_resource.rs
```diff
@@ -815,26 +816,94 @@ pub fn log_stderr_line<'a>(process_id: &u32, trace_line: &'a str) -> &'a str
...
+            trace!("Process {process_id}: {}", trace_line);
             return trace_line;
         }
     };
 
     ""
 }
+
+#[derive(PartialEq, Eq, Deserialize)]
+enum TraceLevel {
```
anmenaga, line 879: We already have `pub enum TraceLevel` in `dsc\src\args.rs`
Can we have just one of these?

anmenaga (approved):

dsc_lib/src/dscresources/command_resource.rs
SteveL-MSFT: I'll shorten it to `pid`

dsc_lib/src/dscresources/command_resource.rs
SteveL-MSFT: Since `dsc` already reaches deep into `dsc_lib` for some things, we can have `dsc` use the one in `dsc_lib`, the only thing that's not ideal is that `dsc_lib` now will have to depend on `clap` crate

Issue #475, opened 2024-06-22
WMI Adapter updates


SteveL-MSFT:
### Summary of the new feature / enhancement

Support:

- set
- export
- delete
- method invocation

### Proposed technical implementation details (optional)

Many WMI classes support methods.  We should maybe just have a `WMIMethodInvocation` metaclass that takes the namespace, class, method name, and a property bag (key/value pairs) that would be the inputs to the method.  The only problem with this approach is that the user needs to know all that info.

Alternatively, but would be noisy when listing, is to have dynamic instances returned that represent each method that can be called.

SteveL-MSFT:
WMI Adapter should use a WMI query if there are properties specified:

```yaml
type: root.cimv2/Win32_OperatingSyste
properties:
  caption:
  version:
  osarchitecture:
  oslanguage:
```

This should be turned in to `select caption, version, osarchitecture, oslanguage from win32_operatingsystem`

SteveL-MSFT:
When we have support for resources to return metadata, the WMI adapter should return the namespace and class used as metadata

PR #548, opened 2024-09-14
Update WMI adapter to perform query based on provided properties

SteveL-MSFT:
# PR Summary

Previously, all the properties for a WMI class gets returned.  With this change, the specified properties get turned into a query.  This can actually improve performance if the WMI provider supports it, but otherwise it also means the output won't have information that the configuration did not request.

If a property has a value, then that value is used as part of the query.

Example output using the updated wmi inventory example:

```yaml
results:
- metadata:
    Microsoft.DSC:
      duration: PT1.0536124S
  name: WMI
  type: Microsoft.Windows/WMI
  result:
    actualState:
    - Model: Virtual Machine
      Domain: WORKGROUP
      Name: CPC-slee-8MW0NA
      Manufacturer: Microsoft Corporation
      TotalPhysicalMemory: 68665831424
    - Caption: Microsoft Windows 11 Enterprise
      OSArchitecture: 64-bit
      OSLanguage: 1033
      Version: 10.0.22631
    - SerialNumber: 9040-0535-4797-8220-8672-1612-79
      Manufacturer: Microsoft Corporation
      Model: null
    - SerialNumber: 0000-0011-2510-4748-5887-4988-97
      Manufacturer: Microsoft Corporation
      Version: VRTUAL - 1
    - MACAddress: 00:22:48:C1:70:B2
      ServiceName: netvsc
      NetConnectionID: Ethernet 6
      NetConnectionStatus: 2
      Name: 'Microsoft Hyper-V Network Adapter #5'
      AdapterType: Ethernet 802.3
```

Also fixed how the adapter writes traces so they show up correctly.

## PR Context

Address a part of https://github.com/PowerShell/DSC/issues/475


wmi-adapter/wmi.resource.ps1
```diff
@@ -9,10 +9,28 @@ param(
     $stdinput
 )
 
+trap {
```
anmenaga, line 12: Would be good to have a comment about what this `trap` does.

dsc/examples/wmi_inventory.dsc.yaml
```diff
@@ -0,0 +1,42 @@
...
+          serialnumber:
+      - name: network adapter
+        type: root.cimv2/Win32_NetworkAdapter
+        properties:
+          name:
+          macaddress:
+          adaptertype:
+          netconnectionid:
+          serviceName:
+          netconnectionstatus: 2
```
anmenaga, line 42: Is my understanding correct that this is the only instance in this file that will use `WHERE` clause?

anmenaga (approved):

dsc/examples/wmi_inventory.dsc.yaml
SteveL-MSFT: Yes

wmi-adapter/wmi.resource.ps1
SteveL-MSFT: ok, will add comment

PR #551, opened 2024-09-18
Update crates and dsc version

SteveL-MSFT:
# PR Summary

Update version to preview.11
Update crates to latest versions
Fix code due to crate changes

### DIFF
--- a/dsc/src/main.rs
+++ b/dsc/src/main.rs
@@ -124,14 +124,14 @@ fn ctrlc_handler() {
 }
 
 fn terminate_subprocesses(sys: &System, process: &Process) {
-    info!("Terminating subprocesses of process {} {}", process.name(), process.pid());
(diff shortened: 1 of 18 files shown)


dsc/src/main.rs
```diff
@@ -187,7 +187,7 @@ Press any key to close this window
     };
 
     // MS Store runs app using `sihost.exe`
-    if parent_process.name().to_lowercase() == "sihost.exe" || parent_process.name().to_lowercase() == "explorer.exe"{
+    if parent_process.name().to_ascii_lowercase() == "sihost.exe" || parent_process.name().to_ascii_lowercase() == "explorer.exe"{
```
anmenaga, line 190: What is the reason for this change?

anmenaga (approved):

dsc/src/main.rs
SteveL-MSFT: The crate version changed and the old method was renamed

Issue #537, opened 2024-09-04
Publish echo resource as `Microsoft.DSC.Debug/Echo`


michaeltlombardi:
### Summary of the new feature / enhancement

> As a user, I want to be able to use a non-acting resource to see how configuration functions and group resources behave.

> As a documentarian, I want to write executable examples that show users how configuration functions and group resources behave.

> As a user, I want to be able to debug complex cases in my configuration by echoing data without having to write custom resources or rely on logging in the resources I consume.

### Proposed technical implementation details (optional)

We have the `Test/Echo` resource implemented, but it's not published in the build and would be useful if published fully as a debug resource. Examples in the reference documentation depend on the echo resource to show behavior, but users who don't build from source can't execute the examples.

PR #553, opened 2024-09-18
Change `Echo` test resource to `Microsoft.DSC.Debug/Echo` resource

SteveL-MSFT:
# PR Summary

Pull the functionality out of the `dsctest` resource into a separate `Microsoft.DSC.Debug/Echo` resource

## PR Context

Fix https://github.com/PowerShell/DSC/issues/537


dscecho/src/args.rs
```diff
@@ -0,0 +1,11 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+use clap::Parser;
+
+#[derive(Debug, Parser)]
+#[clap(name = "dscecho", version = "1.0.0", about = "Debugging helper resource that echos the input.", long_about = None)]
+pub struct Args {
+    #[clap(short, long, help = "The input to the echo command as JSON.  If no input is provided the JSON schema for the input is returned.")]
```
anmenaga, line 9: For a generic 'echo functionality' I don't think we should limit the input to be a JSON. It probably should be any string.

anmenaga (approved):

dscecho/src/args.rs
SteveL-MSFT: What is echo'd doesn't have to be JSON, but the input has to be a valid JSON object that matches the resource schema

PR #563, opened 2024-09-30
Implement schema option for PsDscAdapter

Gijsreyn:
# PR Summary

This pull request adresses partially the option to call `dsc resource schema -r <resourceName>` for PowerShell adapters. If I'm not wrong, there is also an overarching issue but I couldn't find it. Pardon me on it.

There are other questions popping up with this pull request:

- How to define properties with types and description (proposal at [508](https://github.com/PowerShell/DSC/discussions/508))?
- Should definition metadata be included?
- How does the engine handle adapter resources to call schema executable through resource manifest?


powershell-adapter/psDscAdapter/powershell.resource.ps1
```diff
@@ -3,13 +3,14 @@
...
 param(
     [Parameter(Mandatory = $true, Position = 0, HelpMessage = 'Operation to perform. Choose from List, Get, Set, Test, Export, Validate.')]
-    [ValidateSet('List', 'Get', 'Set', 'Test', 'Export', 'Validate', 'ClearCache')]
+    [ValidateSet('List', 'Get', 'Set', 'Test', 'Export', 'Validate', 'Schema', 'ClearCache')]
     [string]$Operation,
     [Parameter(Mandatory = $false, Position = 1, ValueFromPipeline = $true, HelpMessage = 'Configuration or resource input in JSON format.')]
     [string]$jsonInput = '@{}'
 )
 
-function Write-DscTrace {
```
SteveL-MSFT, line 12: Because Pester requires the opening brace on the same line (as it's a parameter to a PowerShell function), so convention has been to keep the opening brace on the same line so it's consistent with Pester tests

SteveL-MSFT (changes requested):
Please add some tests

Gijsreyn:
> Please add some tests

@SteveL-MSFT I added one test now with a hardcoded resource manifest. I'm struggling to find out the logic how `dsc.exe` handles adapter resources types. For now to not break the tests, i have hardcoded the type.

Do you mind helping me out and point me in the right direction? I would have expected the engine to call the `dsc schema` command on adapter resources directly.

powershell-adapter/psDscAdapter/powershell.resource.ps1
Gijsreyn: Should be resolved with the formatting setting in `settings.json`.

ThomasNieto:
@Gijsreyn to confirm this enhancement works for both Windows PowerShell and PowerShell adapters correct?

Gijsreyn:
> @Gijsreyn to confirm this enhancement works for both Windows PowerShell and PowerShell adapters correct?

It's been a while, but yes, I have explicitly added it to the `powershell.resource.ps1`.

SteveL-MSFT:
@Gijsreyn can you resolve the merge conflict?

Gijsreyn:
@SteveL-MSFT The merge conflicts are resolved. I have hardcoded a resource manifest now to let the test pass, but I still need help with the earlier question. Can you guide me in the right direction, or can we pull it in for now?

SteveL-MSFT:
> > Please add some tests
> 
> @SteveL-MSFT I added one test now with a hardcoded resource manifest. I'm struggling to find out the logic how `dsc.exe` handles adapter resources types. For now to not break the tests, i have hardcoded the type.
> 
> Do you mind helping me out and point me in the right direction? I would have expected the engine to call the `dsc schema` command on adapter resources directly.

For adapters, `dsc` engine simply expects the adapter to handle schema validation and reject (as in error) if it doesn't match.  However, this design pre-dates the current work in Bicep and we now have a need to get JSONSchema for all resources (adapted or otherwise) so that intellisense can work.  I created https://github.com/PowerShell/DSC/issues/872 to discuss.

For now, I think we should hold this PR until we resolve how it should work for any type of adapter.

PR #568, opened 2024-10-08
Perf optimization in adapter lookup table

anmenaga:
# PR Summary

Previously adapter lookup table was serialized and saved every time;
With over 1000 entries for adapters shipped with `dsc.exe` this is performance expensive.

This change makes saving only when necessary - when the lookup table is actually updated.

### DIFF
--- a/dsc/tests/dsc_discovery.tests.ps1
+++ b/dsc/tests/dsc_discovery.tests.ps1
@@ -144,10 +144,25 @@ Describe 'tests for resource discovery' {
         $TestClassResourcePath = Resolve-Path "$PSScriptRoot/../../powershell-adapter/Tests"
         $env:DSC_RESOURCE_PATH = $null
         $env:PSModulePath += [System.IO.Path]::PathSeparator + $TestClassResourcePath
-        dsc resource list -a Microsoft.DSC/PowerShell | Out-Null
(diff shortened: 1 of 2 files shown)


dsc/tests/dsc_discovery.tests.ps1
```diff
@@ -144,10 +144,25 @@ Describe 'tests for resource discovery' {
         $TestClassResourcePath = Resolve-Path "$PSScriptRoot/../../powershell-adapter/Tests"
         $env:DSC_RESOURCE_PATH = $null
         $env:PSModulePath += [System.IO.Path]::PathSeparator + $TestClassResourcePath
-        dsc resource list -a Microsoft.DSC/PowerShell | Out-Null
-        "{'Name':'TestClassResource1'}" | dsc -l trace resource get -r 'TestClassResource/TestClassResource' 2> $TestDrive/tracing.txt
 
+        # remove adapter lookup table file
+        Remove-Item -Force -Path $script:lookupTableFilePath -ErrorAction SilentlyContinue
```
SteveL-MSFT, line 149:
```suggestion
        Remove-Item -Force -Path $script:lookupTableFilePath -ErrorAction Stop
```

Shouldn't we have the test fail if the file can't be deleted?

SteveL-MSFT (changes requested):

SteveL-MSFT (approved):

PR #574, opened 2024-10-16
DSC meta configuration

### LINKED ISSUE #282: DSC meta configuration

> As a user, I want to define default behaviors/options for `dsc` so that I don't need to pass the options to every invocation or define a function to do so for me.

Right now, there's no persistent way for a user to control how `dsc` behaves except for the `DSC_RESOURCE_PATH` environmental variable. Users may want to set the default output format for DSC to pretty-json, or only use manifests that have been signed, or disable specific providers (see #274).

As the options for how `dsc` should behave expand, these needs will compound - consider the `WhatIf` scenario, whether to pre-check for permissions, etc.

I haven't done much research on the available options for rust applications, but in an ideal world, we could borrow (a subset of) the functionality that [viper][01] has, supporting layered overrides, where each item in the following list takes precedence over the item below it:

- Explicitly defined flag option, like `--format`
- Environment variable value, like `DSC_FORMAT`
- Configuration file key value, like `format`
- Lookup from a key-value store, like consul or etcd (I think this is out of scope, included for completeness)
- Defined defaults in the code

Even if we only supported flags and environment variables, I think we'd have a more manageable UX. We could implement the configuration file handling later, if ever. With support for using environment variables as default options, users could use `.env` files as lightweight configuration, or set the variables in their CI jobs, or whatever makes sense for their context.

[01]: https://github.com/spf13/viper#why-viper

anmenaga:
# PR Summary

Fix #282 

This PR adds settings/policy JSON files support.
A particular dsc setting value (e.g. dsc resource search path) can be specified in any of the locations below.
Override order:
1) `policy` file. Writable only by Admins, but readable by everyone:
  - on WIndows: `$env:ProgramData+"\dsc\dsc.settings.json"`
  - on Linux: `"/etc/dsc/dsc.settings.json"`
2) command-line parameters. e.g.  `--trace-level ` or `--trace-format`
3) main active settings file: `dsc.settings.json` in `dsc` executable directory
4) default settings file: `dsc_default.settings.json` in `dsc` executable directory. Should not be modified by user and intended to be used in dsc upgrade scenarios as well as `reset settings to defaults` scenario. Top-level nodes in the file are versions of settings schema to support possible future breaking changes in settings schema.

Format and defaults for policy/settings currently look like this and [described in detail in the spec](https://github.com/PowerShell/DSC/issues/282#issuecomment-2338515588):
```
{
  "resourcePath": {
    "allowEnvOverride": true,
    "appendEnvPath": true,
    "directories": []
  },
  "tracing": {
    "level": "WARN",
    "format": "Default",
    "allowOverride": true
  }
}
```


dsc/src/util.rs
```diff
@@ -39,7 +41,7 @@ use syntect::{
     parsing::SyntaxSet,
     util::{as_24_bit_terminal_escaped, LinesWithEndings}
 };
-use tracing::{Level, debug, error, warn, trace};
+use tracing::{Level, info, debug, error, warn, trace};
```
SteveL-MSFT, line 44: Can you fix this so it's in alpha order?

dsc/src/util.rs
```diff
@@ -55,6 +57,23 @@ pub const EXIT_DSC_RESOURCE_NOT_FOUND: i32 = 7;
 pub const DSC_CONFIG_ROOT: &str = "DSC_CONFIG_ROOT";
 pub const DSC_TRACE_LEVEL: &str = "DSC_TRACE_LEVEL";
 
+#[derive(Deserialize)]
+pub struct TracingSetting {
+    level:  TraceLevel,
```
SteveL-MSFT, line 62: Use `///` comments to describe the members

dsc/src/util.rs
```diff
@@ -268,46 +287,78 @@ pub fn write_output(json: &str, format: &Option<OutputFormat>) {
...
-                        "WARN" => &TraceLevel::Warn,
-                        "INFO" => &TraceLevel::Info,
-                        "DEBUG" => &TraceLevel::Debug,
-                        "TRACE" => &TraceLevel::Trace,
-                        _ => {
-                            warn!("Invalid DSC_TRACE_LEVEL value '{level}', defaulting to 'warn'");
-                            &TraceLevel::Warn
-                        },
-                    }
+pub fn enable_tracing(trace_level_arg: Option<TraceLevel>, trace_format_arg: Option<TraceFormat>) {
```
SteveL-MSFT, line 290: Was there a specific reason to not pass by reference?  Also, I don't think we should have the `_arg` suffix as it starts looking more like hungarian

dsc/src/util.rs
```diff
@@ -268,46 +287,78 @@ pub fn write_output(json: &str, format: &Option<OutputFormat>) {
...
+
+    // read setting/policy from files
+    if let Ok(v) = get_setting("tracing") {
+        if v.policy != serde_json::Value::Null {
+            match serde_json::from_value::<TracingSetting>(v.policy) {
+                Ok(v) => {
+                    tracing_setting = v;
+                    policy_is_used = true;
+                },
+                Err(e) => { println!("{e}"); }
```
SteveL-MSFT, line 303: If there's an error, we should just go back to default and use `error!`

dsc/src/util.rs
```diff
@@ -268,46 +287,78 @@ pub fn write_output(json: &str, format: &Option<OutputFormat>) {
...
+            }
+        } else if v.setting != serde_json::Value::Null {
+            match serde_json::from_value::<TracingSetting>(v.setting) {
+                Ok(v) => {
+                    tracing_setting = v;
                 },
-                Err(_) => &TraceLevel::Warn,
+                Err(e) => { println!("{e}"); }
```
SteveL-MSFT, line 310: Same here

dsc/src/util.rs
```diff
@@ -268,46 +287,78 @@ pub fn write_output(json: &str, format: &Option<OutputFormat>) {
...
+            }
+        }
+    } else {
+        println!("Could not read 'tracing' setting");
```
SteveL-MSFT, line 314: Same here

dsc_lib/src/discovery/command_discovery.rs
```diff
@@ -11,23 +11,44 @@ use indicatif::ProgressStyle;
...
 pub struct CommandDiscovery {
     // use BTreeMap so that the results are sorted by the typename, the Vec is sorted by version
     resources: BTreeMap<String, Vec<DscResource>>,
     adapters: BTreeMap<String, Vec<DscResource>>,
     adapted_resources: BTreeMap<String, Vec<DscResource>>,
 }
 
+#[derive(Deserialize)]
+pub struct ResourcePathSetting {
+    allow_env_override: bool,
```
SteveL-MSFT, line 37: Use `///` comments to describe

dsc_lib/src/discovery/command_discovery.rs
```diff
@@ -37,31 +58,75 @@ impl CommandDiscovery {
         }
     }
 
+    fn get_resource_path_setting() -> Result<ResourcePathSetting, DscError>
+    {
+        if let Ok(v) = get_setting("resource_path") {
```
SteveL-MSFT, line 63: In Rust code, we want to use snake_casing, but in the JSON itself, we want to use camelCasing.

dsc_lib/src/discovery/command_discovery.rs
```diff
@@ -37,31 +58,75 @@ impl CommandDiscovery {
...
+                match serde_json::from_value::<ResourcePathSetting>(v.setting) {
+                    Ok(v) => {
+                        return Ok(v);
+                    },
+                    Err(e) => { return Err(DscError::Operation(format!("{e}"))); }
+                }
+            }
+        }
+
+        Err(DscError::Operation("Could not read 'resource_path' setting".to_string()))
```
SteveL-MSFT, line 82: Probably better to have a new `DscError::Setting()` enum value instead of overloading `Operation`

dsc_lib/src/discovery/command_discovery.rs
```diff
@@ -37,31 +58,75 @@ impl CommandDiscovery {
...
+            for p in resource_path_setting.directories {
+                if let Ok(v) = PathBuf::from_str(&p) {
+                    paths.push(v);
                 }
             }
-        };
 
-        let mut paths = env::split_paths(&path_env).collect::<Vec<_>>();
+            if resource_path_setting.append_env_path {
+                trace!("Appending PATH to resource_path");
```
SteveL-MSFT, line 115: This should probably be a debug message as no content is being emitted

dsc_lib/src/util.rs
```diff
@@ -3,6 +3,26 @@
...
 use serde_json::Value;
+use std::fs::File;
+use std::io::BufReader;
+use std::path::PathBuf;
+use std::path::Path;
+use std::env;
+use tracing::debug;
+
+pub struct DscSettingValue {
+    pub setting:  serde_json::Value,
```
SteveL-MSFT, line 14: Should be able to do this and make it more clean unless `Value` conflicts with a different type

```suggestion
    pub setting:  Value,
```

dsc/tests/dsc_settings.tests.ps1
```diff
@@ -0,0 +1,102 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+Describe 'tests for dsc settings' {
+    BeforeAll {
+
+        $script:policyFilePath = if ($IsWindows) {
+            Join-Path $env:ProgramData "dsc" "settings.dsc.json"
+        } else {
+            "/etc/.dsc/settings.dsc.json"
```
SteveL-MSFT, line 10: Is the app folder generally hidden as `.dsc` or should it just be `dsc`?

dsc/tests/dsc_settings.tests.ps1
```diff
@@ -0,0 +1,102 @@
...
+        
+        $script:dscDefaultv1SettingsJson.Replace('"level": "WARN"', '"level": "TRACE"') | Set-Content -Force -Path $script:dscSettingsFilePath
+
+        dsc resource list 2> $TestDrive/tracing.txt
+        "$TestDrive/tracing.txt" | Should -FileContentMatchExactly "Trace-level is Trace"
+    }
+
+    It 'ensure a new resource_path value in settings has effect' {
+        
+        $script:dscDefaultv1SettingsJson.Replace('"directories": []', '"directories": ["TestDir"]') | Set-Content -Force -Path $script:dscSettingsFilePath
```
SteveL-MSFT, line 56: Would be better to have at least two paths

dsc_lib/src/util.rs
```diff
@@ -37,3 +57,98 @@ pub fn parse_input_to_json(value: &str) -> Result<String, DscError> {
...
+
+    if (result.setting == serde_json::Value::Null) &&
+       (result.policy == serde_json::Value::Null) {
+        return Err(DscError::NotSupported(format!("Could not find '{value_name}' in settings").to_string()));
+    }
+
+    Ok(result)
+}
+
+fn load_value_from_json(path: &PathBuf, value_name: &str) -> Result<serde_json::Value, DscError> {
```
SteveL-MSFT, line 118: It looks like we read the file on every query to settings?  Ideally we should just read it once and deserialize into a struct that other things can use.  Perhaps store it in thread local storage?

SteveL-MSFT (changes requested):

build.ps1
```diff
@@ -42,7 +42,9 @@ $filesForWindowsPackage = @(
     'RunCommandOnSet.exe',
     'windowspowershell.dsc.resource.json',
     'wmi.dsc.resource.json',
-    'wmi.resource.ps1'
+    'wmi.resource.ps1',
+    'default_settings.v1.dsc.json',
```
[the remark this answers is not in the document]
anmenaga: I'm ok with changing the extension; but the version should stay in the filename otherwise that would complicate update scenario:

The idea was that any new version of dsc package would have default settings files of all previous versions.
E.g. `dsc v3.3` package/installation would have `default_settings.v1.dsc.json`, `default_settings.v2.dsc.json` and `default_settings.v3.dsc.json` (assuming by v3.3 there would be 2 breaking changes in settings schema).

If the filename would be the same across versions, then the proper settings update logic would have to be in the update tool/script (that we don't have yet), rather than in `dsc.exe`.  I though it's easier to have this logic to be in `dsc.exe`.

dsc/src/util.rs
anmenaga: fixed.

dsc/src/util.rs
anmenaga: updated.

dsc/src/util.rs
anmenaga: "Was there a specific reason to not pass by reference?"
negligible perf impact here - these are options of tiny enums, not much benefit of using references.

"I don't think we should have the _arg suffix as it starts looking more like hungarian"
The _arg suffix is supposed to mean "value from dsc command-line argument".
I'm open to other suggestions.

dsc/src/util.rs
anmenaga: tracing infrastructure is Not initialized at this point, so errors are just silently swallowed.
So I thought that printing to console in the next best thing.

dsc/src/util.rs
anmenaga: same

dsc_lib/src/discovery/command_discovery.rs
anmenaga: updated.

dsc_lib/src/discovery/command_discovery.rs
anmenaga: Updated.

dsc_lib/src/discovery/command_discovery.rs
anmenaga: updated.

dsc_lib/src/util.rs
anmenaga: Updated, even though I don't like it. :)

dsc/tests/dsc_settings.tests.ps1
anmenaga: Good point.
For settings, it probably should not be hidden.

Changed location to "/etc/dsc/settings.dsc.json" throughout the code and updated PR description.

dsc/tests/dsc_settings.tests.ps1
anmenaga: Updated

dsc_lib/src/util.rs
anmenaga: I don't think it is a problem at this point. We read it only 2 times at this point.
We have a much bigger perf problem with repeated resource manifest deserialization; I'd rather spend time on that.

dsc_lib/src/discovery/command_discovery.rs
anmenaga: Updated.

build.ps1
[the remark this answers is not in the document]
SteveL-MSFT: I see.  In that case, my preference would be to have a single json that is essentially a hashtable where the key is the version string.

dsc/src/util.rs
SteveL-MSFT: Agree the perf difference may be negligible, but future developers may be confused as we don't have a consistent pattern.  I think we should default to passing by ref unless we have a good reason not to.

For the param name, I think you should just drop the `_arg` suffix and use variable shadowing (same name) inside the function.

dsc/src/util.rs
SteveL-MSFT: In this case, we can still init to default and use tracing.

dsc_lib/src/util.rs
SteveL-MSFT: In that case, can you just add a comment in the code reflecting this?

anmenaga: Added a comment to the public `get_setting` function that should be used for reading settings.

dsc/src/util.rs
anmenaga: Updated.

michaeltlombardi:
To clarify, the scope of this initial PR is just to get DSC working with the settings/policy files, correct? The CLI / resource for getting and defining the values will follow in a future PR?

anmenaga:
> To clarify, the scope of this initial PR is just to get DSC working with the settings/policy files, correct? The CLI / resource for getting and defining the values will follow in a future PR?

Yes and Yes.

build.ps1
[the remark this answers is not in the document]
anmenaga: Updated to have a single json file with defaults for all versions.

dsc/src/util.rs
anmenaga: Updated.

build.ps1
```diff
@@ -48,6 +48,8 @@ $filesForWindowsPackage = @(
     'wmi.resource.ps1',
     'windows_baseline.dsc.yaml',
     'windows_inventory.dsc.yaml'
+    'dsc_default.settings.json',
+    'settings.dsc.json'
```
SteveL-MSFT, line 52: Should this be `dsc.settings.json`?

build.ps1
```diff
@@ -62,7 +64,9 @@ $filesForLinuxPackage = @(
     'powershell.dsc.resource.json',
     'psDscAdapter/',
     'RunCommandOnSet.dsc.resource.json',
-    'runcommandonset'
+    'runcommandonset',
+    'dsc_default.settings.json',
+    'settings.dsc.json'
```
SteveL-MSFT, line 69: same

build.ps1
```diff
@@ -77,7 +81,9 @@ $filesForMacPackage = @(
...
```
SteveL-MSFT, line 86: same

dsc/src/util.rs
```diff
@@ -268,46 +291,92 @@ pub fn write_output(json: &str, format: &Option<OutputFormat>) {
...
-                        },
-                    }
+#[allow(clippy::too_many_lines)]
+pub fn enable_tracing(trace_level_arg: &Option<TraceLevel>, trace_format_arg: &Option<TraceFormat>) {
+    
+    let mut policy_is_used = false;
+    let mut tracing_setting = TracingSetting::default();
+
+    let default_filter = EnvFilter::try_from_default_env()
+        .or_else(|_| EnvFilter::try_new("error"))
```
SteveL-MSFT, line 301: The default should be `warn`

dsc_lib/src/discovery/command_discovery.rs
```diff
@@ -37,31 +63,74 @@ impl CommandDiscovery {
...
-            match env::var_os("PATH") {
-                Some(value) => {
-                    trace!("Original PATH: {:?}", value.to_string_lossy());
-                    value
-                },
-                None => {
-                    return Err(DscError::Operation("Failed to get PATH environment variable".to_string()));
+            for p in resource_path_setting.directories {
+                let v = PathBuf::from_str(&p);
+                paths.push(v.unwrap_or_default());
```
SteveL-MSFT, line 115: What is default here?  It may be better to error if it can't unwrap.

dsc_lib/src/util.rs
```diff
@@ -37,3 +57,102 @@ pub fn parse_input_to_json(value: &str) -> Result<String, DscError> {
         }
     }
 }
+
+/// Will search setting files for the specified setting.
+/// Performance implication: Use this function economically as every call opens/reads several config files.
```
SteveL-MSFT, line 62:
```suggestion
/// Performance implication: Use this function economically as every call opens/reads several config files.
// TODO: cache the config
```

SteveL-MSFT (changes requested):

build.ps1
anmenaga: Good catch. Updated.

build.ps1
anmenaga: updated.

build.ps1
anmenaga: updated.

dsc/src/util.rs
anmenaga: updated

dsc_lib/src/discovery/command_discovery.rs
anmenaga: There was an error path originally;
but this line was giving me a headache in the style check, because, as it turns out, `from_str` can never fail despite its signature.

dsc_lib/src/util.rs
anmenaga: updated.

dsc/src/util.rs
```diff
@@ -268,46 +291,92 @@ pub fn write_output(json: &str, format: &Option<OutputFormat>) {
...
-                        },
-                    }
+#[allow(clippy::too_many_lines)]
+pub fn enable_tracing(trace_level_arg: &Option<TraceLevel>, trace_format_arg: &Option<TraceFormat>) {
+    
+    let mut policy_is_used = false;
+    let mut tracing_setting = TracingSetting::default();
+
+    let default_filter = EnvFilter::try_from_default_env()
+        .or_else(|_| EnvFilter::try_new("warning"))
```
SteveL-MSFT, line 301: I think it's `warn` and not `warning`

SteveL-MSFT (changes requested):

dsc/src/util.rs
anmenaga: It was a copy-paste from existing code; 
Updated in both places.

SteveL-MSFT (approved):
LGTM

Issue #583, opened 2024-10-24
statically link glibc


SteveL-MSFT:
### Summary of the new feature / enhancement

Different versions of Linux use different versions of glibc, the current build doesn't work on Oracle LInux, for example.  We should statically link glibc to remove the external dependency.

SteveL-MSFT:
As the flag didn't produce a completely statically linked binary, current plan is to just build statically linked musl build which has no external dependencies and work on glibc based systems.

PR #584, opened 2024-10-24
set build to use static glibc

SteveL-MSFT:
# PR Summary

For gnu builds, set the rust c flag to use static linking

## PR Context

Fix https://github.com/PowerShell/DSC/issues/583


SteveL-MSFT:
The rust flags didn't produce a statically linked glibc, so abandoning this

PR #587, opened 2024-10-26
Only build linux musl and update CFS auth

SteveL-MSFT:
# PR Summary

- Since musl build is statically linked, it will work on non-musl based Linux distros and have no dynamic dependencies.
- Update release pipeline to only build musl
- Update build.ps1 tar.gz packaging to rename linux package to just `linux`
- Add suffix to CFS feed so it always authenticates
- With CFS change, we can update cargo.toml and cargo.lock to just specify major.minor versions

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -3,7 +3,7 @@
 registry-auth = true
 
 [registries]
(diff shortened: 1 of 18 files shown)


SteveL-MSFT:
Validated test release

Issue #534, opened 2024-08-29
Offline image configuration


SteveL-MSFT:
### Summary of the new feature / enhancement

Support offline image configuration by having a group resource handle the mounting/unmounting of an image and passing the mounted path to resources that can support it.

### Proposed technical implementation details (optional)

Introduce a tag `offline` that indicates resources support being given a mounted path to work against instead of the current live system.  This would include the `registry` resource and others like `file`, `windowsservice`, etc...

How to pass the mounted path to resources?  Some options:

- A well known property `_mountedPath`
- Metadata that the resource looks for `mountedPath`

citelao:
Besides figuring out how to pass the mount path:

* What are the limitations of an offline resource?
* Can services ([like sshd](https://learn.microsoft.com/en-us/windows-server/administration/openssh/openssh_install_firstuse?tabs=powershell&pivots=windows-server-2025#enable-openssh-for-windows-server-2025)) be started offline? Can group policies (registry/non-registry) be configured offline?
* Will all resources need to support online & offline configuration?

SteveL-MSFT:
I think resources that support `offline` would need to have such a tag in the manifest to make it easier to discover (and the group resource to be described below could validate the resources provided support it).  Services can't be started offline, so any configuration of them would need to be via file or registry (for Windows).

I think we would want a different model that requires starting the image, configure, and shutting it down and not mix it with actual offline configuration.

My initial thinking is that DSC would include a `OfflineGroup` resource that has the properties:

```yaml
type: Microsoft.DSC/OfflineGroup
properties:
  imagePath: <path to vhd/vhdx to mount>
  resources: <array of resources that support `Offline` mode>
```

This group resource would be responsible for:

- verifying every resource specified supports `Offline` mode, otherwise it's an error
- mount the image specified and pass that path to each resource as metadata under `Microsoft.DSC/mountedPath`
- resources that support `Offline` would use this metadata (if present) to read the filesystem (like the registry hive or sshd_config), and write changes to that mounted filesystem on `set`
- unmount the image (on error or after final resource is completed)

For Windows, `%windir%` doesn't have to be on the `C:` drive, so there might need to be a way to pass some image environment variables also as metadata that resources can use.

SteveL-MSFT:
Thinking about this further, one of the problems is that resource authors won't think in terms of online/offline, but I suspect many resources can work against a file path.  So a simpler proposal would be a `DiskImage` resource:

```yaml
[...]

PR #589, opened 2024-10-29
Add `--system-root` parameter, `systemRoot()` and `path()` functions  to dsc

SteveL-MSFT:
# PR Summary

To enable offline configuration management, add `--system-root` parameter and `systemRoot()` function (takes no arguments) to DSC.  The provided path must exist.  Supported resources would need to accept the path to use rather than the current running OS.
If `--system-root` is not specified, then `systemRoot()` will default to `$env:SYSTEMDRIVE` on Windows and `/` on Linux/macOS.
Also added `path()` function that takes any number of strings and concatenates it with the OS directory separator char.

## PR Context

Fix https://github.com/PowerShell/DSC/issues/534

### DIFF
--- a/dsc_lib/src/configure/context.rs
+++ b/dsc_lib/src/configure/context.rs
@@ -25,7 +25,7 @@ impl Context {
         Self {
             execution_type: ExecutionKind::Actual,
             outputs: HashMap::new(),
-            target_path: PathBuf::new(),
+            target_path: get_default_os_target_path(),
             parameters: HashMap::new(),
             security_context: match get_security_context() {
                 SecurityContext::Admin => SecurityContextKind::Elevated,
@@ -41,4 +41,16 @@ impl Default for Context {
     fn default() -> Self {
         Self::new()
     }
+}
+
+#[cfg(target_os = "windows")]
+fn get_default_os_target_path() -> PathBuf {
+    // use SYSTEMDRIVE env var to get the default target path
+    let system_drive = std::env::var("SYSTEMDRIVE").unwrap_or_else(|_| "C:".to_string());
+    PathBuf::from(system_drive).join("\\")
+}
+
+#[cfg(not(target_os = "windows"))]
+fn get_default_os_target_path() -> PathBuf {
+    PathBuf::from("/")
 }
--- a/dsc_lib/src/configure/context.rs
+++ b/dsc_lib/src/configure/context.rs
@@ -12,6 +12,7 @@ use super::config_doc::{DataType, SecurityContextKind};
 pub struct Context {
     pub execution_type: ExecutionKind,
     pub outputs: HashMap<String, Value>, // this is used by the `reference()` function to retrieve output
+    pub mounted_path: String,
     pub parameters: HashMap<String, (Value, DataType)>,
     pub security_context: SecurityContextKind,
     pub variables: HashMap<String, Value>,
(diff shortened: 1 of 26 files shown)


dsc_lib/src/functions/mounted_path.rs
```diff
@@ -0,0 +1,66 @@
...
+use crate::DscError;
+use crate::configure::context::Context;
+use crate::functions::{AcceptedArgKind, Function};
+use serde_json::Value;
+use tracing::debug;
+
+#[derive(Debug, Default)]
+pub struct MountedPath {}
+
+/// Implements the 'mountedpath' function.
```
michaeltlombardi, line 13: Should this use camel casing, like `createArray` and `resourceId`?

dsc_lib/src/functions/mod.rs
```diff
@@ -74,6 +75,7 @@ impl FunctionDispatcher {
         functions.insert("max".to_string(), Box::new(max::Max{}));
         functions.insert("min".to_string(), Box::new(min::Min{}));
         functions.insert("mod".to_string(), Box::new(mod_function::Mod{}));
+        functions.insert("mountedpath".to_string(), Box::new(mounted_path::MountedPath{}));
```
michaeltlombardi, line 78: Should this be camel cased, like `createArray` and `resourceId`?

dsc_lib/src/configure/context.rs
```diff
@@ -12,6 +12,7 @@ use super::config_doc::{DataType, SecurityContextKind};
 pub struct Context {
     pub execution_type: ExecutionKind,
     pub outputs: HashMap<String, Value>, // this is used by the `reference()` function to retrieve output
+    pub mounted_path: String,
```
tgauth, line 15: How did you decide on using a `String`? Curious if `PathBuf` would be useful here.

SteveL-MSFT: Should probably be PathBuf as appending takes care of dir separator automatically.

dsc_lib/src/functions/target_path.rs
```diff
@@ -0,0 +1,60 @@
...
+use crate::DscError;
+use crate::configure::context::Context;
+use crate::functions::{AcceptedArgKind, Function};
+use serde_json::Value;
+use tracing::debug;
+
+#[derive(Debug, Default)]
+pub struct TargetPath {}
+
+/// Implements the 'mountedpath' function.
```
tgauth, line 13:
```suggestion
/// Implements the 'TargetPath' function.
```
there's a few other references to mounted path in this file and others

tgauth:
Would it be helpful to have some sort of `info` logging (maybe in the parser?) in the case of `--target-path` being provided without `TargetPath()` or is it obvious that `--target-path` would be ignored?

SteveL-MSFT:
> Would it be helpful to have some sort of `info` logging (maybe in the parser?) in the case of `--target-path` being provided without `TargetPath()` or is it obvious that `--target-path` would be ignored?

That could be useful as a warning message if the user provided `--target-path` but the configuration doesn't use it.  The challenge is that expressions are parsed at runtime, so the warning would only happen at the end after configuration was run.  I think this is ok.

SteveL-MSFT:
Decided to rename to `systemRoot` instead of `targetPath` as it more accurately describes what it does

tgauth (approved):

Issue #165, opened 2023-08-23
`DSC/AssertionGroup` reports as out of state for `test` method even when all instances are in the desired state


michaeltlombardi:
### Prerequisites

- [X] Write a descriptive title.
- [X] Make sure you are able to repro it on the latest version
- [X] Search the existing issues.

### Steps to reproduce

When using the `DSC/AssertionGroup` with the `dsc config test` command, I expect that the resource group is considered to be in the desired state when all instances in the group are in the desired state.

However, the resource group returns `inDesiredState` as `false` and lists the `resources` property in the `differingProperties` result field.

Define the following configuration on a 64-bit system as `repro.dsc.config.yaml`:

```yaml
$schema: https://raw.githubusercontent.com/PowerShell/DSC/main/schemas/2023/08/config/document.json
resources:
  - name: Operating System Assertion
    type: DSC/AssertionGroup
    properties:
      $schema: https://raw.githubusercontent.com/PowerShell/DSC/main/schemas/2023/08/config/document.json
      resources:
        - name: Is64BitOS
          type: Microsoft/OSInfo
          properties:
            bitness: '64'
```

### Expected behavior

```sh
cat ./repro.config.dsc.yaml | dsc config test
```

```yaml
results:
- name: Operating System Assertion
  type: DSC/AssertionGroup
  result:
    desiredState:
      $schema: https://raw.githubusercontent.com/PowerShell/DSC/main/schemas/2023/08/config/document.json
      resources:
      - name: Is64BitOS
        type: Microsoft/OSInfo
        properties:
          bitness: '64'
    actualState:
      results:
      - name: Is64BitOS
        type: Microsoft/OSInfo
        result:
          desiredState:
            bitness: '64'
          actualState:
            $id: https://developer.microsoft.com/json-schemas/dsc/os_info/20230303/Microsoft.Dsc.OS_Info.schema.json
            family: Windows
            version: 10.0.22621
            edition: Windows 11 Enterprise
            bitness: '64'
          inDesiredState: true
          differingProperties: []
      messages: []
      hadErrors: false
    inDesiredState: true
    differingProperties: []
messages: []
hadErrors: false
```

### Actual behavior

```sh
cat ./repro.config.dsc.yaml | dsc config test
```

```yaml
results:
- name: Operating System Assertion
  type: DSC/AssertionGroup
  result:
    desiredState:
      $schema: https://raw.githubusercontent.com/PowerShell/DSC/main/schemas/2023/08/config/document.json
[...]

PR #595, opened 2024-11-08
fix output of Assertion resource for test to resemble config and add support for array comparison

SteveL-MSFT:
# PR Summary

Two problems addressed by this PR:

- The output of `Assertion` simply returned the actual result to look like a `get` result, however, this causes the synthetic test doing a diff to fail as they don't resemble each other.  To address this, added a new hidden switch `--as--config` to inform `dsc` that this is a special case and the output needs to be transformed to a Configuration so that the synthetic test can work.

Here's example of the new output:

```yaml
metadata:
  Microsoft.DSC:
    version: 3.0.0
    operation: Test
    executionType: Actual
    startDatetime: 2024-11-07T16:45:48.237829-08:00
    endDatetime: 2024-11-07T16:45:48.789203-08:00
    duration: PT0.551374S
    securityContext: Restricted
results:
- metadata:
    Microsoft.DSC:
      duration: PT0.441752S
  name: Operating System Assertion
  type: Microsoft.DSC/Assertion
  result:
    desiredState:
      resources:
      - name: Is64BitOS
        type: Microsoft/OSInfo
        properties:
          bitness: '64'
      - name: 64bit test 2
        type: Microsoft/OSInfo
        properties:
          family: Windows
    actualState:
      $schema: https://raw.githubusercontent.com/PowerShell/DSC/main/schemas/2024/04/config/document.json
      resources:
      - type: Microsoft/OSInfo
        name: Is64BitOS
        properties:
          $id: https://developer.microsoft.com/json-schemas/dsc/os_info/20230303/Microsoft.Dsc.OS_Info.schema.json
          family: macOS
          version: 15.1.0
          bitness: '64'
          architecture: arm64
      - type: Microsoft/OSInfo
        name: 64bit test 2
        properties:
          $id: https://developer.microsoft.com/json-schemas/dsc/os_info/20230303/Microsoft.Dsc.OS_Info.schema.json
          family: macOS
          version: 15.1.0
          bitness: '64'
          architecture: arm64
    inDesiredState: false
    differingProperties:
    - resources
messages: []
hadErrors: false
```

- The second issue is that the synthetic test didn't handle arrays at all and `resources` is always an array.  Arrays need to be the same per the current diff rules:
  - same size
  - each element needs to exist
  - objects in actual can be superset of the expected

## PR Context

Fix https://github.com/PowerShell/DSC/issues/165

### DIFF
--- a/dsc/src/args.rs
+++ b/dsc/src/args.rs
@@ -102,8 +102,10 @@ pub enum ConfigSubCommand {
         path: Option<String>,
         #[clap(short = 'f', long, help = "The output format to use")]
         format: Option<OutputFormat>,
+        // Used by Assertion resource to return `test` result as a `get` result
         #[clap(long, hide = true)]
         as_get: bool,
+        // Used by Assertion resource to return `test` result as a configuration `test` result
         #[clap(long, hide = true)]
         as_config: bool,
     },
--- a/dsc/src/args.rs
+++ b/dsc/src/args.rs
@@ -103,6 +103,8 @@ pub enum ConfigSubCommand {
         #[clap(short = 'f', long, help = "The output format to use")]
         format: Option<OutputFormat>,
         #[clap(long, hide = true)]
+        as_get: bool,
+        #[clap(long, hide = true)]
         as_test: bool,
     },
     #[clap(name = "validate", about = "Validate the current configuration", hide = true)]
--- a/dsc/src/args.rs
+++ b/dsc/src/args.rs
@@ -103,7 +103,7 @@ pub enum ConfigSubCommand {
         #[clap(short = 'f', long, help = "The output format to use")]
         format: Option<OutputFormat>,
         #[clap(long, hide = true)]
-        as_get: bool,
+        as_test: bool,
     },
     #[clap(name = "validate", about = "Validate the current configuration", hide = true)]
     Validate {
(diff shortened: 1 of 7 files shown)


dsc/src/args.rs
```diff
@@ -102,6 +102,8 @@ pub enum ConfigSubCommand {
         format: Option<OutputFormat>,
         #[clap(long, hide = true)]
         as_get: bool,
+        #[clap(long, hide = true)]
```
SteveL-MSFT, line 105: Add description for hidden switches

dsc_lib/src/dscresources/dscresource.rs
```diff
@@ -363,28 +373,176 @@ pub fn get_diff(expected: &Value, actual: &Value) -> Vec<String> {
...
+            return false;
+        }
+    }
+
+    true
+}
+
+fn array_contains(array: &Vec<Value>, find: &Value) -> bool {
+    for item in array {
+        if find.is_boolean() && item.is_boolean() && find.as_bool().unwrap() == item.as_bool().unwrap() {
```
tgauth, line 435: would it be more efficient to nest the checks as an if/else? 
a worst case example - `find=false` and `item=true`

anmenaga (approved):

dsc_lib/src/dscresources/dscresource.rs
SteveL-MSFT: Since the comparison is left to right, if any fail, then no additional processing is needed.  For example, if `find` is NOT a boolean, the latter checks aren't executed.  If we have a nested `if` statement, then it'll be slightly more expensive as it starts a new execution of an `if` statement rather than just completing the previous one.

I actually started with nested `if` and clippy complained :)

PR #600, opened 2024-11-19
Allow module without version

Gijsreyn:
# PR Summary

Allow modules that don't have a version directory to be added in `$env:PSModulePath`.

## PR Context

The `Get-DscResourceModules` function scans the `$env:PSModulePath` to find relevant modules to be included in the cache. When a module doesn't have a version directory, it does not find the module. This change allows the ability to include modules that don't contain the directory.

I added this change because it can be useful for testing purposes to see the participation of DSC resources when developing them. For example, take the `winget-dsc` repository. Simply adding the resource to the environment variable adds it for participation.


SteveL-MSFT (changes requested):
Please add a test

powershell-adapter/psDscAdapter/psDscAdapter.psm1
```diff
@@ -33,14 +33,15 @@ function Get-DSCResourceModules
...
 
-        foreach($moduleFolder in Get-ChildItem $folder -Directory)
-        {
-            $addModule = $false
-            foreach($psd1 in Get-ChildItem -Recurse -Filter "$($moduleFolder.Name).psd1" -Path $moduleFolder.fullname -Depth 2)
-            {
-                $containsDSCResource = select-string -LiteralPath $psd1 -pattern '^[^#]*\bDscResourcesToExport\b.*'
-                if($null -ne $containsDSCResource)
-                {
+        $moduleFolders = Get-ChildItem $folder -Directory
```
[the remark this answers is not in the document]
anmenaga: 1. This is copy-paste of the original DSCv2 `Get-DscResource`;
2. `Get-Module -ListAvailable` has a bug in it where it incorrectly reporting exported DSC resources in some situations.
3. `Get-Module -ListAvailable` is slow compared to current scanning of psd1's found in PSModulePath.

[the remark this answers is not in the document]
Gijsreyn: I experienced the same, especially when you have a large set of modules across multiple PSModulePaths.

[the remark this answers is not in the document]
SteveL-MSFT: Can we add a comment in the code for future reference?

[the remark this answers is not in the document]
Gijsreyn: Got it covered Steve. Thanks Andrew for the explanation. Would you mind pulling it in if it is good?

powershell-adapter/Tests/TestClassNoVersion/TestClassNoVersion.psd1
```diff
@@ -0,0 +1,52 @@
...
+    Author               = 'Microsoft'
+
+    # Company or vendor of this module
+    CompanyName          = 'Microsoft Corporation'
+
+    # Copyright statement for this module
+    Copyright            = '(c) Microsoft. All rights reserved.'
+
+    # Functions to export from this module, for best performance, do not use wildcards and do not delete the entry, use an empty array if there are no functions to export.
+    FunctionsToExport    = '*'
```
SteveL-MSFT, line 28: Please remove the `*` or anything not actually used as part of this test module, it's simply bad practice to have the wildcard and don't want others to use this as an example

powershell-adapter/Tests/powershellgroup.resource.tests.ps1
```diff
@@ -4,10 +4,11 @@
 Describe 'PowerShell adapter resource tests' {
 
     BeforeAll {
-        $OldPSModulePath  = $env:PSModulePath
+        $OldPSModulePath = $env:PSModulePath
         $env:PSModulePath += [System.IO.Path]::PathSeparator + $PSScriptRoot
 
-        if ($IsLinux -or $IsMacOS) {
+        if ($IsLinux -or $IsMacOS)
+        {
```
SteveL-MSFT, line 11: Convention is to have the opening brace on the previously line

SteveL-MSFT (changes requested):

Gijsreyn:
Added the `OTBS` to the settings.json file. It might help other folks in the future when they are working with the PowerShell scripts. All remarks resolved @SteveL-MSFT

powershell-adapter/Tests/TestClassNoVersion/TestClassNoVersion.psd1
```diff
@@ -0,0 +1,52 @@
...
+    CompanyName          = 'Microsoft Corporation'
+
+    # Copyright statement for this module
+    Copyright            = '(c) Microsoft. All rights reserved.'
+
+    # Functions to export from this module, for best performance, do not use wildcards and do not delete the entry, use an empty array if there are no functions to export.
+    FunctionsToExport    = @()
+
+    # Cmdlets to export from this module, for best performance, do not use wildcards and do not delete the entry, use an empty array if there are no cmdlets to export.
+    CmdletsToExport      = '*'
```
SteveL-MSFT, line 31: There's no cmdlets being exported, right?

SteveL-MSFT (changes requested):

powershell-adapter/Tests/TestClassNoVersion/TestClassNoVersion.psd1
Gijsreyn: I took the same example as `TestClassResource.psd1`, but just leaving it empty now.

SteveL-MSFT (approved):
LGTM, waiting on Andrew to review

powershell-adapter/psDscAdapter/psDscAdapter.psm1
```diff
@@ -32,15 +32,17 @@ function Get-DSCResourceModules
         {
             continue
         }
+	# Partially re-used code from the original DSCv2 Get-DSCResource code
+	# Get-Module -ListAvailable is slow and has a bug causing incorrect reporting
+        $moduleFolders = Get-ChildItem $folder -Directory
+        if (-not $moduleFolders) {
+            $moduleFolders = Get-ChildItem (Split-Path $folder -Parent) | Where-Object { $_.Name -eq (Split-Path $folder -Leaf) }
```
anmenaga, line 39: The `-Depth 2` parameter in original code was supposed to handle both cases: modules with version subdirs and without.
Can you please give an example directory/module structure that was not handled correctly by old code?

anmenaga (changes requested):

powershell-adapter/psDscAdapter/psDscAdapter.psm1
Gijsreyn: When working in the winget-dsc repository, I often found myself doing something like:

```powershell
$env:PSModulePath += ";C:\source\winget-dsc\resources\Microsoft.DotNet.Dsc"

# Call function
Get-DscResourcesModules
# Old code did not list out the module
```

Hope that sheds some light on it.

anmenaga: Can you please post an output of 
`Get-ChildItem -Recurse -Path C:\source\winget-dsc\resources\Microsoft.DotNet.Dsc | %{$_.FullName}`
Thank you.

Gijsreyn: There you go:

image

I have added both. The left is the main branch, and the right is the current one in this pull request.

anmenaga: Thank you; in the right picture - what is the full path to `NpmDsc.psd1` that is returned by `Get-DSCResourceModules` ?

Gijsreyn: Added the following line to the script to display full path: `Write-Host "The .psd1 file: $($psd1.FullName)`

image

anmenaga: Ok, I think I got what's happening.
The module discovery in PowerShell (and consequently DSC PS adapter) requires that all module files should be in the module directory; and `$env:PSModulePath` should point to the parent of that module directory.
In your case, the line should be:
`$env:PSModulePath += ";C:\source\winget-dsc\resources"`
rather than
`$env:PSModulePath += ";C:\source\winget-dsc\resources\NpmDsc"`.
This should work with existing code. Please try it.

Gijsreyn: Thanks for the explanation, Andrew. I tried it in the past, and that works. But sometimes, I want only particular modules to be found, with the structure mentioned in this PR and in the test. What are your thoughts on those?

SteveL-MSFT: @Gijsreyn I think the problem is that this is contrary to how module discovery works with PS7 and it would be preferable to not deviate from that.  Is this only for development purposes or do you have a real scenario where you want to limit the discovery?

Gijsreyn: I can understand, and this was the use case. Feel free to abandon the pull request. I appreciated the time you guys looked into it.

Gijsreyn: Hi @SteveL-MSFT. After testing for a while, I got a scenario that breaks the current testing cycle.

Imagine the following use case found while developing in winget-dsc. To test out both cases by either using `Invoke-DscResource` or the class directly, we declare at the top of the test the `using module` statement.

When modules are being discovered in the lower levels, so say C:\source\winget-dsc\resources\NpmDsc, it correctly loads the `using module` statement, but it doesn't do it for PSDesiredStateConfiguration. If I now add it on higher level in the same call, I get an error stating:

PS C:\Users\User> $env:PSModulePath += ";C:\source\winget-dsc\resources\Microsoft.Windows.Setting.Language"
PS C:\Users\User> Get-DscResource
Exception: Exception setting "Module": "Cannot convert the "System.Object[]" value of type "System.Object[]" to type
"System.Management.Automation.PSModuleInfo"."

I know you've closed down the ticket, but perhaps we can still consider it. Thanks.

anmenaga: @Gijsreyn Please open a new issue and copy-paste your last post there.

Issue #588, opened 2024-10-27
DSC on Azure Pipelines Ubuntu


Gijsreyn:
### Prerequisites

- [X] Write a descriptive title.
- [X] Make sure you are able to repro it on the latest version
- [X] Search the existing issues.

### Summary

I am experiencing issues when installing `dsc.exe` on `ubuntu-latest` agent pool in Azure Pipelines. I'm running the following script:

```yaml
jobs: 
  - job: test
    displayName: 'test'
    pool: 
      vmImage: ubuntu-latest
    steps: 
      - checkout: self 
        persistCredentials: true 
      - task: PowerShell@2
        displayName: "Run"
        inputs:
          pwsh: true
          targetType: "inline"
          script: |
            $uri = 'https://github.com/PowerShell/DSC/releases/download/v3.0.0-preview.11/DSC-3.0.0-preview.11-x86_64-unknown-linux-gnu.tar.gz'
            $filePath = '/tmp/DSC-' + '3.0.0-preview.11' + '-x86_64-unknown-linux-gnu.tar.gz'

            curl -L -o $filePath $uri
            # Create the target folder where powershell will be placed
            sudo mkdir -p /opt/microsoft/dsc

            # Expand dsc to the target folder
            sudo tar zxf $filePath -C /opt/microsoft/dsc

            # Set execute permissions
            sudo chmod +x /opt/microsoft/dsc

            # Create the symbolic link that points to dsc
            sudo ln -s /opt/microsoft/dsc /usr/bin/dsc

            # Add to path
            $env:PATH += [System.IO.Path]::PathSeparator + "/usr/bin/dsc"

            # Call version
            dsc --version 

            # Call list of resources
            dsc -l trace resource list
          ignoreLASTEXITCODE: true

```

The `dsc --version` is properly called, but when doing the resource list, I can observe:

![image](https://github.com/user-attachments/assets/8bf2d488-97aa-4143-a438-228ba67d51ca)

This hangs for an infinity time. 

### Steps to reproduce

See description

### Expected behavior

```console
Able to run `dsc resource list` on ubuntu-latest agent
```

### Actual behavior

```console
Hangs
```

### Environment data

```PowerShell
Name                           Value
----                           -----
PSVersion                      7.4.5
PSEdition                      Core
GitCommitId                    7.4.5
OS                             Microsoft Windows 10.0.26100
Platform                       Win32NT
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0…}
[...]

PR #601, opened 2024-11-26
Change how to pass STDIN input via `--file`, rename other args

SteveL-MSFT:
# PR Summary

- Remove automatic reading of STDIN as it can block in environments that redirect STDIN but never send input like ADO Pipelines
- Use convention `--file -` to indicate to read from STDIN, updated CLI help
- Rename `--format` to `--output-format` to use `-f` to only mean `--file`, shorthand is now `-o`
- Rename `--document` to `--input` to have it consistent for config and resource invocation
- Rename `--trace-format` shorthand to `-t`
- Lots of test changes due to the parameters change

## PR Context

Fix https://github.com/PowerShell/DSC/issues/588

### DIFF
--- a/dsc/src/subcommand.rs
+++ b/dsc/src/subcommand.rs
@@ -220,6 +220,8 @@ pub fn config_export(configurator: &mut Configurator, format: &Option<OutputForm
 }
 
 fn initialize_config_root(path: &Option<String>) -> Option<String> {
+    // code that calls this pass in either None, Some("-"), or Some(path)
+    // in the case of `-` we treat it as None, but need to pass it back as subsequent processing needs to handle it
     let use_stdin = if let Some(specified_path) = path {
         if specified_path != "-" {
             return Some(set_dscconfigroot(specified_path));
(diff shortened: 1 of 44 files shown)


tgauth (approved):
LGTM

dsc/src/subcommand.rs
```diff
@@ -220,33 +220,45 @@ pub fn config_export(configurator: &mut Configurator, format: &Option<OutputForm
 }
 
 fn initialize_config_root(path: &Option<String>) -> Option<String> {
-    if path.is_some() {
-        let config_path = path.clone().unwrap_or_default();
-        Some(set_dscconfigroot(&config_path))
-    } else if std::env::var(DSC_CONFIG_ROOT).is_ok() {
+    let use_stdin = if let Some(specified_path) = path {
+        if specified_path != "-" {
```
anmenaga, line 224: I'm trying to understand why `set_dscconfigroot` and input from STDIN are mutually exclusive...
What are the things that we can read from STDIN ? Only configuration itself?

anmenaga (approved):

dsc/src/subcommand.rs
SteveL-MSFT: I'll add a comment here.  What happens here is that `--file` gets passed all the way to this function so `path` has either `None`, `-`, or an actual path (directory).  So we have to check if `-` is passed which means the path is not valid and we treat this function the same as if no path is provided.  However, the way this function is used, the path is passed back out of it and used later so we have to put back the `-` so subsequent code knows it's STDIN.

PR #603, opened 2024-12-03
Fix change in clippy rule from `&Option<Type>` to `Option<&Type>`

SteveL-MSFT:
# PR Summary

Clippy had a new rule where it's idiomatic to use `Option<&Type>` instead of `&Option<Type>` so that the contents of the Option are passed by reference instead of the Option wrapper.

The counter part of this is that you now need to use `.as_ref()` with an Option so that the contents are referenced.

Also fixed one missed arg from `--format` to `--output-format`

### DIFF
--- a/dsc/src/subcommand.rs
+++ b/dsc/src/subcommand.rs
@@ -34,7 +34,7 @@ use std::{
 };
 use tracing::{debug, error, trace};
 
-pub fn config_get(configurator: &mut Configurator, format: &Option<OutputFormat>, as_group: &bool)
+pub fn config_get(configurator: &mut Configurator, format: Option<&OutputFormat>, as_group: &bool)
 {
     match configurator.invoke_get() {
         Ok(result) => {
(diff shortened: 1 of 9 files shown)


dsc/src/subcommand.rs
```diff
@@ -606,7 +606,7 @@ fn list_resources(dsc: &mut DscManager, resource_name: &Option<String>, adapter_
 
             // if description is specified, skip if resource description does not contain it
             if description.is_some() &&
-                (manifest.description.is_none() | !manifest.description.unwrap_or_default().to_lowercase().contains(&description.as_ref().unwrap_or(&String::new()).to_lowercase())) {
+                (manifest.description.is_none() | !manifest.description.unwrap_or_default().to_lowercase().contains(&description.as_ref().unwrap_or(&&String::new()).to_lowercase())) {
```
tgauth, line 609:
```suggestion
                (manifest.description.is_none() | !manifest.description.unwrap_or_default().to_lowercase().contains(description.unwrap_or(&String::new()).to_lowercase())) {
```
Is the double reference still necessary for `description` now that it's a `&String` when unwrapped?

SteveL-MSFT: Good catch!  The double ref isn't needed, but `description` still needs to be a reference `.contains()` requires yet, but I was on auto-pilot making the changes and the `.as_ref()` which makes it a double ref isn't needed, so then it just needs to be `&String::new()`.  I'll make a fix.

tgauth (approved):

PR #604, opened 2024-12-06
(i18n) Provide translation strings for Registry

michaeltlombardi:
# PR Summary

This change replaces static strings in the registry app/lib with translation strings using `rust_i18n`.

### DIFF
--- a/registry/Cargo.toml
+++ b/registry/Cargo.toml
@@ -3,6 +3,11 @@ name = "registry"
 version = "1.0.0"
 edition = "2021"
 
+[package.metadata.i18n]
+available-locales = ["en-us"]
+default-locale = "en-us"
+load-path = "locales"
+
 [profile.release]
 strip = true
 # optimize for size
@@ -14,6 +19,7 @@ lto = true
 clap = { version = "4.5", features = ["derive"] }
 crossterm = "0.28"
 registry = "1.2"
+rust-i18n = { version = "3.0" }
 schemars = "0.8"
 serde = "1.0"
 serde_json = "1.0"
--- a/registry/src/args.rs
+++ b/registry/src/args.rs
@@ -2,9 +2,10 @@
 // Licensed under the MIT License.
 
 use clap::{Parser, Subcommand};
+use rust_i18n::t;
 
 #[derive(Parser)]
-#[clap(name = "registry", version = "0.0.1", about = "Manage state of Windows registry", long_about = None)]
+#[clap(name = "registry", version = "0.0.1", about = t!("args.about").to_string(), long_about = None)]
 pub struct Arguments {
 
     #[clap(subcommand)]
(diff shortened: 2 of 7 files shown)


registry/Cargo.toml
```diff
@@ -3,6 +3,11 @@ name = "registry"
 version = "1.0.0"
 edition = "2021"
 
+[package.metadata.i18n]
+available-locales = ["en"]
```
SteveL-MSFT, line 7: We should use `en-us` for the locale

registry/locales/cli.yml
```diff
@@ -0,0 +1,87 @@
+_version: 2
```
SteveL-MSFT, line 1: I believe Windows resources are typically one folder/file per locale, so we may want to stick with version 1

registry/src/args.rs
```diff
@@ -2,9 +2,10 @@
 // Licensed under the MIT License.
 
 use clap::{Parser, Subcommand};
+use rust_i18n::t;
 
 #[derive(Parser)]
-#[clap(name = "registry", version = "0.0.1", about = "Manage state of Windows registry", long_about = None)]
+#[clap(name = "registry", version = "0.0.1", about = t!("cli.about").to_string(), long_about = None)]
```
SteveL-MSFT, line 8: Cool, didn't think about using `to_string()` to make i18n work with clap.

registry/src/main.rs
```diff
@@ -6,6 +6,13 @@ use crossterm::event;
 #[cfg(debug_assertions)]
 use std::env;
 
+// #[macro_use]
+// extern crate rust_i18n;
+
+// Init translations using the `[package.metadata.i18n]` section in `Cargo.toml`
+use rust_i18n::t;
+rust_i18n::i18n!("locales");
```
SteveL-MSFT, line 14: Should specify `fallback = "en-us"`

registry/src/main.rs
```diff
@@ -135,19 +142,24 @@ pub fn enable_tracing() {
     let subscriber = tracing_subscriber::Registry::default().with(fmt).with(filter);
 
     if tracing::subscriber::set_global_default(subscriber).is_err() {
-        eprintln!("Unable to set global default tracing subscriber.  Tracing is diabled.");
+        let msg = t!("error.eprint.tracing_init");
```
SteveL-MSFT, line 145: These can just be:

```rust
eprintln!("{}", t!("error.eprint.tracing_init");
```

SteveL-MSFT (changes requested):

registry/src/main.rs
michaeltlombardi: I think that gets pulled from he package metadata, but can explicitly set it here.

registry/locales/cli.yml
michaeltlombardi: I'm not strictly opposed to that convention, especially if it's more familiar to contributors—I went with v2 because of how I was organizing the translations and believe that they'll be easier to maintain compared to replicating the structure across each file.

SteveL-MSFT: The other thing to consider is that I've been using `toml` and organizing it by source files and keeping it in order with the source file to make it a bit easier to map later

SteveL-MSFT:
@michaeltlombardi can you resolve the merge conflict

michaeltlombardi:
@SteveL-MSFT - not sure I understand how we're getting this error:

https://github.com/PowerShell/DSC/actions/runs/13246929993/job/37177774845#step:3:1628

Given that earlier in the build we're successfully downloading and compiling v1.2.13 for the treesitter:

https://github.com/PowerShell/DSC/actions/runs/13246929993/job/37177774845#step:3:47

And the failure is happening for `pal` which this PR doesn't seem to affect.

SteveL-MSFT (approved):

PR #614, opened 2024-12-12
Support writing progress as JSON

### LINKED ISSUE #592: Support writing progress as JSON

For integration tools that have GUIs, have an option to write progress as JSON to stderr instead of writing to console.

_No response_

anmenaga:
# PR Summary

Fix #592 

0) This adds a new argument to `dsc.exe`:
```
  -p, --progress-format <PROGRESS_FORMAT>
          Progress format to use [possible values: default, json]
```
1) For `default` progress is displayed in UI only;
  for `json` - progress messages are also written to STDERR stream in JSON form.
2) Fixed a bug with `Searching for adapted resources` progress which previously was always showing 0.

The structure of json progress message is following. `activity` and `percent_complete` are expected to always have value; the rest of the fields are optional.
Same structure will be expected from resources when they will want to provide progress status to `dsc` (this support will be  implemented soon).
```
pub struct DscProgressValue {
    pub activity:  String,
    pub status: String,
    pub percent_complete: u16,
    pub seconds_remaining: u64,
}
```

Dev notes:
1) Added `struct DscProgressBar` which is wrapping both UI progress bar and json serialization; this structure should be used for progress reporting gong forward.


dsc/src/subcommand.rs
```diff
@@ -282,11 +283,11 @@ pub fn config(subcommand: &ConfigSubCommand, parameters: &Option<String>, mounte
                     exit(EXIT_DSC_ERROR);
                 }
             };
-            (new_parameters, config_json)
+            (new_parameters, config_json, output_format)
         }
     };
 
-    let mut configurator = match Configurator::new(&json_string) {
+    let mut configurator = match Configurator::new(&json_string, output_format.as_ref()) {
```
SteveL-MSFT, line 290: I don't like having `output_format` as a parameter to the constructor.  Would prefer adding a `set_output_format()` function instead.  Alternatively since this is something used all over the place, maybe put into thread local storage and just retrieve from there instead of passing it around particularly since it shouldn't change during runtime.

dsc/tests/dsc_config_get.tests.ps1
```diff
@@ -55,4 +55,26 @@ Describe 'dsc config get tests' {
...
+            - name: Echo
+              type: Microsoft.DSC.Debug/Echo
+              properties:
+                output: hello
+"@
+        $config_yaml | dsc config get --output-format json -f - 2> $TestDrive/ErrorStream.txt
+        $LASTEXITCODE | Should -Be 0
+        $lines = Get-Content $TestDrive/ErrorStream.txt
+        $ProgressMessagesFound = $False
+        foreach ($line in $lines) {
```
SteveL-MSFT, line 72: instead of comparing text, since all output should be JSON, you should convert every line from json and check the object

dsc/tests/dsc_resource_list.tests.ps1
```diff
@@ -58,6 +58,20 @@ Describe 'Tests for listing resources' {
         }
     }
 
+    It 'json progress for resource subcommand' {
+        dsc resource list -o json -a '*' 2> $TestDrive/ErrorStream.txt
+        $LASTEXITCODE | Should -Be 0
+        $lines = Get-Content $TestDrive/ErrorStream.txt
+        $ProgressMessagesFound = $False
+        foreach ($line in $lines) {
+            if ($line.Contains("activity")) { # if line is a progress message
```
SteveL-MSFT, line 67: same here, use convertfrom-json

dsc_lib/src/configure/mod.rs
```diff
@@ -22,21 +24,21 @@ use security_context_lib::{SecurityContext, get_security_context};
...
 pub mod parameters;
 
-pub struct Configurator {
+pub struct Configurator<'a> {
     json: String,
     config: Configuration,
     pub context: Context,
     discovery: Discovery,
     statement_parser: Statement,
+    progress_format: Option<&'a OutputFormat>,
```
SteveL-MSFT, line 41: Instead of having this as an Option, I think it should just be `OutputFormat` and you have the constructor set a default.  This should also remove the need for lifetime.

dsc_lib/src/util.rs
```diff
@@ -24,6 +37,88 @@ impl Default for DscSettingValue {
     }
 }
 
+#[derive(Default, Debug, Clone, Serialize)]
+pub struct DscProgressValue {
```
SteveL-MSFT, line 41: I think the struct should just be called `Progress`

dsc_lib/src/util.rs
```diff
@@ -24,6 +37,88 @@ impl Default for DscSettingValue {
...
+    pub activity:  String,
+    pub status: String,
+    pub percent_complete: u16,
+    pub seconds_remaining: u64,
+}
+
+pub struct DscProgressBar {
```
SteveL-MSFT, line 48: This can just be `ProgressBar`

SteveL-MSFT (changes requested):

dsc_lib/src/util.rs
anmenaga: Updated.

dsc_lib/src/util.rs
anmenaga: Updated.

dsc_lib/src/configure/mod.rs
anmenaga: Updated.

dsc/src/subcommand.rs
anmenaga: Added `set_output_format()`

dsc/tests/dsc_config_get.tests.ps1
anmenaga: Updated.

dsc/tests/dsc_resource_list.tests.ps1
anmenaga: Updated.

dsc/tests/dsc_config_get.tests.ps1
```diff
@@ -56,6 +56,31 @@ Describe 'dsc config get tests' {
...
+            if ($line.StartsWith('{')) {
```
SteveL-MSFT, line 73: This check shouldn't be needed, just have the trace format be JSON and if any text comes through, that's a bug

dsc/tests/dsc_resource_list.tests.ps1
```diff
@@ -58,6 +58,23 @@ Describe 'Tests for listing resources' {
         }
     }
 
+    It 'json progress for resource subcommand' {
+        dsc resource list -o json -a '*' 2> $TestDrive/ErrorStream.txt
+        $LASTEXITCODE | Should -Be 0
+        $lines = Get-Content $TestDrive/ErrorStream.txt
+        $ProgressMessagesFound = $False
+        foreach ($line in $lines) {
+            if ($line.StartsWith('{')) {
```
SteveL-MSFT, line 67: same here

dsc_lib/src/util.rs
```diff
@@ -24,6 +39,95 @@ impl Default for DscSettingValue {
     }
 }
 
+/// Only `activity` and `percent_complete` fields are mandatory for Progress messages
+#[derive(Default, Debug, Clone, Serialize)]
+pub struct Progress {
+    pub activity:  String,
+    #[serde(default, skip_serializing_if = "is_default")]
+    pub status: String,
```
SteveL-MSFT, line 47: Seems like this should be an Option

dsc_lib/src/util.rs
```diff
@@ -24,6 +39,95 @@ impl Default for DscSettingValue {
...
+    pub percent_complete: u16,
+    #[serde(default, skip_serializing_if = "is_default")]
+    pub seconds_remaining: u64,
```
SteveL-MSFT, line 50: Make as Option.  Even `percent_complete` is probably an Option

dsc_lib/src/util.rs
```diff
@@ -24,6 +39,95 @@ impl Default for DscSettingValue {
...
+}
+
+pub struct ProgressBar {
+    progress_value:  Progress,
+    ui_bar: Span, //IndicatifSpanExt
+    length: u64,
+    position: u64,
+    emit_json: bool
```
SteveL-MSFT, line 58: perhaps this should be OutputFormat instead to make it more future proof?

SteveL-MSFT (changes requested):

dsc/tests/dsc_config_get.tests.ps1
anmenaga: updated

dsc/tests/dsc_resource_list.tests.ps1
anmenaga: updated

dsc_lib/src/util.rs
anmenaga: Updated.

dsc_lib/src/util.rs
anmenaga: Updated.

dsc_lib/src/util.rs
anmenaga: We'll worry about that when we get there. :)

SteveL-MSFT (changes requested):
One thing I just realized, the progress output format should not be tied to the result output format.  Instead, you should add a top level `--progress-format` like how we have `--trace-format` to specify it.

anmenaga:
> One thing I just realized, the progress output format should not be tied to the result output format. Instead, you should add a top level `--progress-format` like how we have `--trace-format` to specify it.

Updated code and PR description as suggested.

SteveL-MSFT (approved):

PR #616, opened 2024-12-16
Ensure `systemRoot()` always has trailing separator character

SteveL-MSFT:
# PR Summary

Change so that `systemRoot()` always includes a trailing separator.  This is due to how `PathBuf.push()` works on Windows vs non-Windows and `path()` adopts this behavior:

- On Windows, if you join a drive letter `c:` and `windows`, you get `c:windows` which is a relative path to the current `c:` drive location
- On Windows, if you join a drive letter in the middle, it's treated as absolute and replaces the previous path segments: `'a', 'c:', 'b'` will be `c:b`, you would use `c:\` to make sure it's absolute
- On non-Windows, the colon is valid directory character, so nothing special happens and is allowed

Added extra tests to `path()` that aren't needed for this PR, but good to have.

### DIFF
--- a/dsc_lib/src/functions/path.rs
+++ b/dsc_lib/src/functions/path.rs
@@ -54,7 +54,12 @@ mod tests {
     fn start_with_drive_letter() {
         let mut parser = Statement::new().unwrap();
         let result = parser.parse_and_execute("[path('C:\\','test')]", &Context::new()).unwrap();
+
+        #[cfg(target_os = "windows")]
         assert_eq!(result, format!("C:{SEPARATOR}test"));
+
+        #[cfg(not(target_os = "windows"))]
+        assert_eq!(result, format!("C:\\{SEPARATOR}test"));
     }
 
     #[test]
@@ -68,7 +73,7 @@ mod tests {
 
         // non-Windows, the colon is a valid character in a path
         #[cfg(not(target_os = "windows"))]
-        assert_eq!(result, format!("a{SEPARATOR}C:{SEPARATOR}test"));
+        assert_eq!(result, format!("a{SEPARATOR}C:\\{SEPARATOR}test"));
     }
 
     #[test]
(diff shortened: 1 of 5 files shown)


dsc_lib/src/functions/path.rs
```diff
@@ -31,8 +31,15 @@ impl Function for Path {
         debug!("Executing path function with args: {:?}", args);
 
         let mut path = PathBuf::new();
+        let mut first = true;
         for arg in args {
             if let Value::String(s) = arg {
+                // if first argument is a drive letter, add it with a separator suffix as PathBuf.push() doesn't add it
```
tgauth, line 37: Given the drive path in middle case, this isn't really a check for the first argument, just the first argument that is a drive letter. If it was checking for the first argument, `first` would need to be set to `false` regardless of the value of the `arg` after the first iteration.

Consider updating this comment or maybe renaming `first` for clarity?

dsc_lib/src/functions/path.rs
```diff
@@ -31,8 +31,15 @@ impl Function for Path {
...
+                if first && s.len() == 2 && s.chars().nth(1).unwrap() == ':' {
```
adityapatwardhan, line 38: On Unix `~/a:/b:/c:/d` is a valid path too. Will this change work with multiple `:` ?

Add a test for multiple `:` in path.

adityapatwardhan (changes requested):

dsc_lib/src/functions/path.rs
SteveL-MSFT: So what's happening with `PathBuf.push()` is that if you do: `'c:','windows'`, you'd expect `c:\windows`, but instead, you get `c:windows` on Windows.  So the check here is only if the first element is a drive letter.  However, based on their docs, if you do: `'a','c:','b'`, then you actually get `c:\b` as the driver letter in the middle is interpreted as absolute.  Not how I would have designed it, but this is the current behavior based on their docs.

dsc_lib/src/functions/path.rs
SteveL-MSFT: I'll add that test, but the behavior will be different between Windows and non-Windows

dsc_lib/src/functions/path.rs
SteveL-MSFT: Actually, thinking about this a bit, the alternative is to have `systemRoot()` return a trailing slash and just keep the `pathBuf` behavior and not have any special handling here.  I think I prefer this although it does mean `systemRoot()` doesn't match `%SYSTEMDRIVE%`, but I think it's ok.

tgauth (approved):

PR #620, opened 2024-12-28
Winget manifest uploader

Gijsreyn:
# PR Summary

This PR adds a function to the build script to upload the assets related to Windows machines to the winget-pkg repository.

## PR Context

The function grabs the relevant content types for Windows machines. One switch is introduced to submit the manifest, including a parameter to add the GitHub token. 

The only two things remaining is to have it added to the workflow after a release is created and not rebuild the enter project. I was thinking something of this:

```yaml
name: Winget
on:
  release:
    types: [published]

jobs:
  publish:
    runs-on: <agentName>
    env:
      WINGETCREATE_TOKEN: ${{ secrets.WINGETCREATE_TOKEN }}
    steps:
    - name: Create manifest and submit PR to winget-pkgs
      run: |
        ./build.ps1 -SubmitWinGetManifest -GitToken $env:WINGETCREATE_TOKEN
```

> [!NOTE]
> The `OTBS` setting formatted some code in the build.ps1 file.


SteveL-MSFT:
@Gijsreyn thanks for this, but I think the team needs to discuss if we want to separately publish to winget given that winget can install from the Store.  Just not sure if we want to support this different way of installing.

Gijsreyn:
@SteveL-MSFT Got it Steve. Personally with all the issues related to the store, having the ability to choose the `installerType` and the `source`, just throws in a better installation experience. When a user installs it from the `winget` source, the user in itself can choose for `.zip` or .`msixbundle`.

SteveL-MSFT:
@denelon do you think there's value/reason to publish to WinGet repo in addition to the Store?

denelon:
I think there are folks who would prefer WinGet as it can support making earlier versions available as opposed to the store just offering the latest version. If you add the packaged version here in addition to the other methods, users can choose which type they prefer, but WinGet would still default to the MSIX package for a new installation unless the user specifies via arguments or settings.

SteveL-MSFT:
@Gijsreyn can you resolve the merge conflict and I'll review and get this merged.  Thanks!

Gijsreyn:
@SteveL-MSFT Merge conflicts resolved. I had to rewrite a little part, because the PR was quite outdated. There's a new switch introduced that switches between `Microsoft.DSC` or `Microsoft.DSC.Preview` based on `-PreRelease`. Do note: the current WinGet repository requires 3 URLs. The `.msixbundle` is missing from the assets. See earlier talks at #638

denelon:
@Gijsreyn, I'd also suggest taking a look at how PowerToys publishes to WinGet using wingetcreate. @mdanish-kh has done a bunch of these at GitHub.

mdanish-kh:
While this script approach can definitely work, it involves a lot more "plumbing" than what could be simplified into a relatively straightforward post-publish GitHub Action. Examples:
- [microsoft/terminal](https://github.com/microsoft/terminal/blob/main/.github/workflows/winget.yml) <--- Handles both preview and stable releases
- [microsoft/edit](https://github.com/microsoft/edit/blob/main/.github/workflows/winget.yml)
- [microsoft/PowerToys](https://github.com/microsoft/PowerToys/blob/main/.github/workflows/package-submissions.yml)

Regardless, if it works it works :)

build.ps1
```diff
@@ -746,4 +732,46 @@ if ($packageType -eq 'msixbundle') {
...
+    $restParameters = @{
+        SslProtocol = 'Tls13'
+        Headers     = @{'X-GitHub-Api-Version' = '2022-11-28' }
+    }
+
+    $assets = ((Invoke-RestMethod -uri "https://api.github.com/repos/$Project/releases" -Headers $restParameters) | 
+        Where-Object -Property prerelease -EQ $IsPreRelease.IsPresent |
+         Select-Object -First 1 -Property assets).assets # ExpandProperty did not work
+    # Grab the download URLs for supported WinGet
+    $downloadUrls = $assets.Where({ $_.content_type -in @('application/zip', 'application/octet-stream') }).browser_download_url
```
mdanish-kh: This will break if a new zip based asset is added in the future. I suggest regex matching specific installers from the assets. One can refer to the sample workflows I shared above on how

build.ps1
```diff
@@ -746,4 +732,46 @@ if ($packageType -eq 'msixbundle') {
...
+
+    # TODO: Version number might change in the future
+    $url = $downloadUrls[0]
+    if ($url -match 'v(\d+\.\d+\.\d+)-preview\.(\d+)') {
+        $firstPart = $matches[1].Remove(3) # Remove the last digit
+        $lastPart = $matches[2] + ".0"
+        $version = "$firstPart.$lastPart"
+
+        $packageId = "$packageId.Preview"
+    }
```
mdanish-kh: Why not grab the version from the GitHub API response itself (there's a `tag_name` field in the response) instead of manually parsing it from the URL?

build.ps1
```diff
@@ -746,4 +732,46 @@ if ($packageType -eq 'msixbundle') {
...
+
+
+    & wingetcreate.exe update $packageId --version $version --urls $downloadUrls --submit --token $GitToken
```
mdanish-kh: With wingetcreate, you have to be explicit with how many URLs you're passing to the update command, because otherwise the update command will fail if the number of installers in the latest manifest at winget-pkgs don't match what's passed in the update command.

What this means is - if we're including MSIX in the latest manifest at winget-pkgs, this update command must also contain the MSIX URL otherwise the update will fail. I know the MSIX installer availability is delayed for any DSC releases so that may be an area of concern. If we're not including MSIX in the latest winget-pkgs manifests, then the workflow will work as intended

Gijsreyn: Hey @mdanish-kh, thanks for jumping in, I appreciate it! 

I have now created a `winget.yml` file. Do you mind having a glimpse one more time? If it's okay, I'll remove the other content from the `build.ps1`. 

Coming back to the MSIX, it depends if it will be added to the assets. @SteveL-MSFT ?

SteveL-MSFT: msixbundle will be added to the assets in the near future for 3.2

build.ps1
```diff
@@ -797,4 +783,46 @@ if ($packageType -eq 'msixbundle') {
     Write-Host -ForegroundColor Green "`ntar.gz file is created at $tarFile"
 }
 
-$env:RUST_BACKTRACE=1
+function Submit-DSCWinGetAssets {
+    param(
+        [string]$GitToken,
+        [switch] $IsPreRelease
```
SteveL-MSFT:
```suggestion
        [switch]$IsPreRelease
```

SteveL-MSFT (changes requested):
@Gijsreyn it doesn't seem like all of Muhammad's feedback has been addressed?  Also, do we only care about winget for msixbundle or do we expect people to want the zip?

.github/workflows/winget.yml
```diff
@@ -0,0 +1,27 @@
...
+  REGEX: 'DSC-(\d+\.\d+\.\d+(?:-preview\.\d+)?)-(?:x86_64|aarch64)-pc-windows-msvc\.zip'
+
+jobs:
+  publish:
+    runs-on: windows-latest # Action can only run on Windows
+    steps:
+      - name: Publish Microsoft.DSC ${{ github.event.release.prerelease && 'Preview' || 'Stable' }}
+        run: |
+          $assets = '${{ toJSON(github.event.release.assets) }}' | ConvertFrom-Json
+          $wingetRelevantAsset = $assets | Where-Object { $_.name -like '*.zip' -and $_.name -like '*.msixbundle' } | Select-Object -First 1
```
mdanish-kh: This code looks like it's either only gonna get the zip URL **OR** the MSIX URL, and we don't know which will it be?

mdanish-kh (changes requested):
I don't think we've addressed the underlying issue for us to be able to automate this properly, which is we don't know what release assets are gonna be available when the release is created / published. Is it gonna be both ZIP & MSIX? Or only ZIP then MSIX is published later? This matters because as I mentioned in https://github.com/PowerShell/DSC/pull/620#discussion_r2132675560, wingetcreate requires the number of URLs passed to update command to match the number of installers in the latest winget-pkgs manifest. That means, if winget-pkgs manifest contains ZIP and MSIX installers, then the update command needs to receive ZIP and MSIX URLs for the new release. If winget-pkgs only contains ZIP, then the update flow expects only a ZIP URL to be provided, otherwise the command will fail. We can debate if this is an area of improvement for wingetcreate (which it may be -> https://github.com/microsoft/winget-create/issues/351#issuecomment-1480294787), but the current flow requires same count of URLs.

---

Some hacky approaches I can think of:

### Retry with multiple commands

- Try first with wingetcreate update passing only the ZIP URL -> this will pass if release assets includes ZIP & winget-pkgs manifest for latest DSC version only includes ZIP
- If above fails, it probably means the winget-pkgs manifest contains MSIXBundle installer & update command needs the MSIX URL to succeed. 
    - If the MSIX is available in release asset, pass that to wingetcreate upgrade command along with ZIP URL
    - If MSIX is added at a later time to release assets, retry the workflow again when it's added or have the workflow be triggered automatically at a "release edited" event. (this may complicate things if release has to be manually edited for any other reason than adding the MSIX URL, invoking the workflow again for no reason & creating a PR)

```yaml
on:
  release:
    types: [published, edited]
```

SteveL-MSFT:
> I don't think we've addressed the underlying issue for us to be able to automate this properly, which is we don't know what release assets are gonna be available when the release is created / published. Is it gonna be both ZIP & MSIX? 

Would it make sense to only support MSIX for now?  Or do you think users would prefer ZIP since they can just install MSIX via Store (except for Server)?

mdanish-kh:
> Would it make sense to only support MSIX for now? Or do you think users would prefer ZIP since they can just install MSIX via Store (except for Server)?

I'm not particularly aware of the user preferences, since I haven't seen an issue at [winget-pkgs](https://github.com/microsoft/winget-pkgs) requesting a particular installer for DSC. Personally I would prefer the packaged MSIX version to be available via WinGet for convenience. Users can install it once & future updates can be handled automatically via the MSFT Store (or manually by WinGet as well). But I don't see why we would want to support "**only**" MSIX since ZIP is always available in the release assets. No reason to exclude that installation medium when it's always available in the release assets. I feel the question should be should we support MSIX or no, and to that I feel we should. WinGet will always default to MSIX installation if user doesn't specify, but having all installation mediums available gives users more choices.

SteveL-MSFT:
> > Would it make sense to only support MSIX for now? Or do you think users would prefer ZIP since they can just install MSIX via Store (except for Server)?
> 
> I'm not particularly aware of the user preferences, since I haven't seen an issue at [winget-pkgs](https://github.com/microsoft/winget-pkgs) requesting a particular installer for DSC. Personally I would prefer the packaged MSIX version to be available via WinGet for convenience. Users can install it once & future updates can be handled automatically via the MSFT Store (or manually by WinGet as well). But I don't see why we would want to support "**only**" MSIX since ZIP is always available in the release assets. No reason to exclude that installation medium when it's always available in the release assets. I feel the question should be should we support MSIX or no, and to that I feel we should. WinGet will always default to MSIX installation if user doesn't specify, but having all installation mediums available gives users more choices.

That makes sense.  In that case, we should support both MSIX and ZIP

.github/workflows/winget.yml
```diff
@@ -0,0 +1,27 @@
+name: Publish to Winget
+
+on:
+  release:
+    types: [published]
```
mdanish-kh:
```suggestion
on:
  workflow_dispatch:
  release:
    types: [published]
```

Let's add `workflow_dispatch` so that the job can be re-run manually from the Actions tab

.github/workflows/winget.yml
```diff
@@ -0,0 +1,27 @@
...
+jobs:
+  publish:
+    runs-on: windows-latest # Action can only run on Windows
+    steps:
+      - name: Publish Microsoft.DSC ${{ github.event.release.prerelease && 'Preview' || 'Stable' }}
+        run: |
+          $assets = '${{ toJSON(github.event.release.assets) }}' | ConvertFrom-Json
+          $wingetRelevantAsset = $assets | Where-Object { $_.name -like '*.zip' -and $_.name -like '*.msixbundle' } | Select-Object -First 1
+          $regex = [Regex]::New($env:REGEX)
+          $version = $regex.Match($wingetRelevantAsset.name).Groups[1].Value
```
mdanish-kh:
```suggestion
          $version = (${{ toJSON(github.event.release.tag_name) }}).Trim('v')
```

We don't need to perform regex matching. The version info is already available in the GitHub release event

.github/workflows/winget.yml
mdanish-kh:
```suggestion
          $x64ZIPInstallerUrl = $assets | Where-Object -Property name -like '*x86_64-pc-windows-msvc.zip' | Select-Object -ExpandProperty browser_download_url
          $arm64InstallerUrl = $assets | Where-Object -Property name -like '*aarch64-pc-windows-msvc.zip' | Select-Object -ExpandProperty browser_download_url
          $msixInstallerURL = $assets | Where-Object -Property name -like '*Win.msixbundle' | Select-Object -ExpandProperty browser_download_url
```

As I mentioned in a comment previously, the current code looks like it's either only gonna get the zip URL OR the MSIX URL, and we don't know which will it be. We should get URLs something like this instead

.github/workflows/winget.yml
```diff
@@ -0,0 +1,27 @@
...
+
+          $wingetPackage = "Microsoft.DSC${{ github.event.release.prerelease && '.Preview' || '' }}"
+
+          & curl.exe -JLO https://aka.ms/wingetcreate/latest
+          & .\wingetcreate.exe update $wingetPackage -s -v $version -u $wingetRelevantAsset.browser_download_url -t "${{ secrets.WINGET_TOKEN }}"
+
```
mdanish-kh:
```suggestion
          & .\wingetcreate.exe update $wingetPackage --version $version --urls $x64ZIPInstallerUrl $arm64InstallerUrl $msixInstallerURL --submit
```

Suggested changes to the command based on feedback above. Use of [`--token` param is deprecated in winget-create](https://github.com/microsoft/winget-create/blob/main/doc/token.md#usage). We should be using the `WINGET_CREATE_GITHUB_TOKEN` environment variable instead.

.github/workflows/winget.yml
```diff
@@ -0,0 +1,27 @@
+name: Publish to Winget
+
+on:
+  release:
+    types: [published]
+
+
+env:
+  REGEX: 'DSC-(\d+\.\d+\.\d+(?:-preview\.\d+)?)-(?:x86_64|aarch64)-pc-windows-msvc\.zip'
```
mdanish-kh:
```suggestion
env:
  WINGET_CREATE_GITHUB_TOKEN: ${{ secrets.WINGET_CREATE_GITHUB_TOKEN }}
```

removing REGEX env variable as not needed (assuming you agree with my suggestions above). Adding the `WINGET_CREATE_GITHUB_TOKEN` environment variable to be used for submission

.github/workflows/winget.yml
```diff
@@ -0,0 +1,27 @@
...
+    runs-on: windows-latest # Action can only run on Windows
+    steps:
+      - name: Publish Microsoft.DSC ${{ github.event.release.prerelease && 'Preview' || 'Stable' }}
+        run: |
+          $assets = '${{ toJSON(github.event.release.assets) }}' | ConvertFrom-Json
+          $wingetRelevantAsset = $assets | Where-Object { $_.name -like '*.zip' -and $_.name -like '*.msixbundle' } | Select-Object -First 1
+          $regex = [Regex]::New($env:REGEX)
+          $version = $regex.Match($wingetRelevantAsset.name).Groups[1].Value
+
+          $wingetPackage = "Microsoft.DSC${{ github.event.release.prerelease && '.Preview' || '' }}"
```
mdanish-kh:
```suggestion
          $wingetPackage = "Microsoft.DSC"
```

Personally I would suggest not handling Preview releases _for now_ to get this merged & handle preview in a future PR. The issue is that WinGet pipelines enforce that the PackageVersion should exactly match the MSIX version. For reference, the GitHub version for the latest preview is `3.2.0-preview.2`, but the MSIX version is actually `3.2.2`. Would need to add some logic to modify the version only for preview to get the MSIX version. You may do it here if you want, but I would suggest a separate PR since there are enough comments in this one already :D

mdanish-kh (changes requested):
@SteveL-MSFT, can you make sure that `WINGET_CREATE_GITHUB_TOKEN` is saved as a repository secret with the GitHub PAT value? Steps on generating the Personal access token needed for GitHub PR submission can be found [here in the WinGet-Create README](https://github.com/microsoft/winget-create#github-personal-access-token-classic-permissions).

Due to limitations in the GitHub API, we can't use fine-grained tokens. As such, I would advise not creating the GitHub PAT from your own account, but rather using a separate bot account. For reference, the PowerToys & Terminal team uses this bot account https://github.com/consvc for their submissions

Gijsreyn:
> @mdanish-kh can you do one more review of this? Thanks!
> 
> @Gijsreyn there's an empty `build.ps1` file?

Either I don't have my glasses on, or there is a little bug in GitHub @SteveL-MSFT, as you can see it's correctly in my branch: https://github.com/Gijsreyn/operation-methods/blob/winget-manifest-uploader/build.ps1

mdanish-kh:
> Either I don't have my glasses on, or there is a little bug in GitHub @SteveL-MSFT, as you can see it's correctly in my branch: [Gijsreyn/operation-methods@`winget-manifest-uploader`/build.ps1](https://github.com/Gijsreyn/operation-methods/blob/winget-manifest-uploader/build.ps1)


@Gijsreyn You've updated the file permissions somehow it seems & that's why it's being shown in the GitHub diffs

image


To reset the changes to the file, you can run `git checkout main -- .\build.ps1` in your branch and then commit & push the changes

.github/workflows/winget.yml
```diff
@@ -0,0 +1,29 @@
...
+    types: [published]
+
+
+env:  
+  WINGET_CREATE_GITHUB_TOKEN: ${{ secrets.WINGET_CREATE_GITHUB_TOKEN }}  
+
+jobs:
+  publish:
+    runs-on: windows-latest # Action can only run on Windows
+    steps:
```
mdanish-kh:
```suggestion
  publish:
    runs-on: windows-latest # Action can only run on Windows
    # Only submit stable releases
    if: ${{ !github.event.release.prerelease }}
    steps:
```

.github/workflows/winget.yml
```diff
@@ -0,0 +1,29 @@
...
+      - name: Publish Microsoft.DSC ${{ github.event.release.prerelease && 'Preview' || 'Stable' }}
```
mdanish-kh:
```suggestion
      - name: Publish DSC package
```

mdanish-kh (commented):
Just a couple of things and then this looks good to me. Since we are not handling Preview releases in this workflow for now, suggesting the following change to prevent workflow run for preview releases

Gijsreyn:
> > Either I don't have my glasses on, or there is a little bug in GitHub @SteveL-MSFT, as you can see it's correctly in my branch: [Gijsreyn/operation-methods@`winget-manifest-uploader`/build.ps1](https://github.com/Gijsreyn/operation-methods/blob/winget-manifest-uploader/build.ps1)
> 
> @Gijsreyn You've updated the file permissions somehow it seems & that's why it's being shown in the GitHub diffs
> 
> image
> To reset the changes to the file, you can run `git checkout main -- .\build.ps1` in your branch and then commit & push the changes

And you just spared me a couple of hours debugging. Thanks for both the comments. Should be resolved!

SteveL-MSFT (approved):

PR #621, opened 2025-01-13
Add support for `contentVersion` property

SteveL-MSFT:
# PR Summary

- if exists in input, the value is not validated and ignored
- if configuration is generated (like for `export`), it is set to `1.0.0` (which is semver compatible) even though ARM docs use the example `1.0.0.0` which likely comes from .NET Version class

## PR Context

Needed for Bicep integration which expects `contentVersion` for both input and output


tgauth (approved):

andyleejordan (commented):
Just got a chance to test this, worked great!

PR #626, opened 2025-01-16
Enable `Include` resource to take configuration and parameters as string content

SteveL-MSFT:
# PR Summary

The `Include` resource currently only takes the configuration to include and parameters as external files.  This change enables a choice of either a file path `configurationFile`, `parametersFile` or the content as a string `configurationContent`, `parametersContent` (can be mixed).  This is useful where another resource might generate a configuration that is used later within a larger configuration without needing to save it to a file first.

@michaeltlombardi for docs, it's important that the nested JSON/YAML is a text string and NOT a nested object this is why the tests have embedded YAML as a multi-line string and the JSON in quotes.

There is future work to allow a nested object for configuration and parameters that can be added to the enum but not part of this PR.

Seems like there is a new clippy rule for unnecessary `map` use so fixed that as well although unrelated to my changes.

### DIFF
--- a/dsc/src/resolve.rs
+++ b/dsc/src/resolve.rs
@@ -99,27 +99,6 @@ pub fn get_contents(input: &str) -> Result<(Option<String>, String), String> {
                     return Err(format!("{} '{include_path:?}': {err}", t!("resolve.invalidFile")));
                 }
             }
-            // // try to deserialize the Include content as YAML first
-            // let configuration: Configuration = match serde_yaml::from_str(&include_content) {
-            //     Ok(configuration) => configuration,
-            //     Err(_err) => {
-            //         // if that fails, try to deserialize it as JSON
-            //         match serde_json::from_str(&include_content) {
-            //             Ok(configuration) => configuration,
-            //             Err(err) => {
-            //                 return Err(format!("{} '{include_path:?}': {err}", t!("resolve.invalidFile")));
-            //             }
-            //         }
-            //     }
-            // };
-
-            // // serialize the Configuration as JSON
-            // match serde_json::to_string(&configuration) {
-            //     Ok(json) => json,
-            //     Err(err) => {
-            //         return Err(format!("JSON: {err}"));
-            //     }
-            // }
         },
         IncludeKind::Content(text) => {
             match parse_input_to_json(&text) {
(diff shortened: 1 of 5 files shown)


dsc/src/resolve.rs
```diff
@@ -14,14 +13,30 @@ use tracing::{debug, info};
...
+    #[serde(rename = "configurationContent")]
+    Content(String),
+}
+
+#[derive(Debug, Clone, PartialEq, Deserialize, Serialize, JsonSchema)]
+pub enum IncludeParametersKind {
     #[serde(rename = "parametersFile")]
-    pub parameters_file: Option<String>,
+    FilePath(String),
+    #[serde(rename = "parametersContent")]
```
anmenaga, line 30: These `serde renames` complicate debugging.
If a field name is dictated by good user-facing name in json, then it would be good to have it the same in the code.

dsc/src/resolve.rs
```diff
@@ -51,74 +66,104 @@ pub fn get_contents(input: &str) -> Result<(Option<String>, String), String> {
...
-                Ok(configuration) => configuration,
+            match parse_input_to_json(&include_content) {
+                Ok(json) => json,
                 Err(err) => {
                     return Err(format!("{} '{include_path:?}': {err}", t!("resolve.invalidFile")));
                 }
             }
-        }
-    };
+            // // try to deserialize the Include content as YAML first
```
anmenaga, line 102: here and below - why comment out this code instead of removing?

anmenaga (approved):

dsc/src/resolve.rs
SteveL-MSFT: The reason here is to conform to standards in JSON while also conforming to standards in Rust.  In general, it shouldn't be confusing as the only difference is camel casing vs snake casing.

dsc/src/resolve.rs
SteveL-MSFT: this was accidentally left over from refactoring this code, will remove

dsc/src/resolve.rs
SteveL-MSFT: Also, because this is an enum, Rust says to use Pascal casing

Issue #389, opened 2024-04-09
Unable to reuse `reference()` in set operation from get and test


michaeltlombardi:
### Prerequisites

- [X] Write a descriptive title.
- [X] Make sure you are able to repro it on the latest version
- [X] Search the existing issues.

### Steps to reproduce

When using the `reference()` function, you need to use dot-path notation to access the data from the result for a resource operation. For both `get` and `test` operations, you can use the `actualState` property of the return object to retrieve the state of the resource. However, for `set` operations, you need to use either the `beforeState` or `afterState` property.

This means that you can't define a `reference()` function once and reuse it across all operations - you need to define a separate configuration document.

You can reproduce this issue with this minimal configuration document:

```yaml
# repro.dsc.config.yaml
$schema: https://raw.githubusercontent.com/PowerShell/DSC/main/schemas/2023/08/config/document.json
resources:
- name: current user registry
  type: Microsoft.Windows/Registry
  properties:
    keyPath: HKLM\Software\Microsoft\Windows NT\CurrentVersion
    valueName: ProductName
    _exist: true
- name: Echo Reference
  type: Test/Echo
  properties:
    output: "[reference(resourceId('Microsoft.Windows/Registry', 'current user registry')).actualState.valueName]"
  dependsOn:
    - "[resourceId('Microsoft.Windows/Registry', 'current user registry')]"
```

Steps:

```powershell
$config = Get-Content -Raw -Path ./repro.dsc.config.yaml
$config | dsc config get
$config | dsc config test
$config | dsc config set
```

## Proposed fix

The following options occur to me:

1. Do the step-in for `actualState` and `afterState` on behalf of the users, instead of requiring them to manually step in with dot-notation - so instead of `reference(...).actualState.valueName`, the user would write `reference(...).valueName`. Are there cases where users will want to get at values _other than_ the canonical state of the resource with a reference? 
1. Rename the `actualState` and/or `afterState` properties to use the same name. I'm not sure I see a good option here, all of the new names I can think of are less descriptive of what's being returned.
1. Provide an error handler that specifically tries `afterState` for the `set` operation if it can't find the `actualState` key. This would quietly enable users to define one reference and reuse it across all operations, but it's a little bit non-obvious and adds a special code path.

### Expected behavior

```console
PS> $config | dsc config get
```
```yaml
results:
- name: current user registry
  type: Microsoft.Windows/Registry
  result:
    actualState:
[...]

PR #634, opened 2025-02-04
Change `reference()` to use the raw output of a resource

SteveL-MSFT:
# PR Summary

As discussed, `reference()` now refers to the raw output of a prior resource invocation and not the decorated one that comes from DSC engine.  This is because the decoration is different for `get`, `set`, `test`, and `export` such that use of this function means you can't use the same config for both `get` and `test`, for example.

Had to update version of tree-sitter to get past the failures in CI but not locally (also updated some other crates)

Also renamed an internal member to `references` instead of `outputs` so that the comment is no longer needed.

We'll defer the ability to access the decorated output as it's additive and not breaking until we have a real scenario to need it.

## PR Context

Fix https://github.com/PowerShell/DSC/issues/389


dsc/tests/dsc_reference.tests.ps1
```diff
@@ -2,8 +2,13 @@
 # Licensed under the MIT License.
 
 Describe 'Tests for config using reference function' {
-    It 'Reference works' {
-        $out = dsc config get -f $PSScriptRoot/../examples/reference.dsc.yaml | ConvertFrom-Json
+    It 'Reference works for <operation>' -TestCases @(
+        @{ operation = 'get' },
+        @{ operation = 'test' }
```
tgauth, line 7: are there any tests for reference with `set` and `export`?

tgauth (approved):

dsc/tests/dsc_reference.tests.ps1
SteveL-MSFT: Not at this time, will add later

PR #670, opened 2025-03-05
demonstrate file and log

mgreenegit:
# PR Summary


SteveL-MSFT (changes requested):
This isn't going to automatically get picked up as a test, did you mean to put this under `dsc/examples`?

Issue #674, opened 2025-03-06
`_inDesiredState` not being respected in `test` `state` output


JohnMcPMS:
### Prerequisites

- [x] Write a descriptive title.
- [x] Make sure you are able to repro it on the latest version
- [x] Search the existing issues.

### Summary

A resource that implements a `test` command and returns `state` with the `_inDesiredState` property [as documented](https://learn.microsoft.com/en-us/powershell/dsc/reference/schemas/resource/properties/indesiredstate?view=dsc-3.0#description) still has the `inDesiredState` top level property calculated by the property diff.

Related to #108, although my test case is actually the inversion of the described reason to add `_inDesiredState` (no property diff, but not in desired state).

Workarounds:
1. Fail if relevant properties are not provided to `test` rather than using defaults.
2. Make any relevant properties required.
3. [BEST] Return `stateAndDiff` so that the resource can control the property diff.

### Steps to reproduce

First, implement a resource with a `test` command that returns `state` with an `_inDesiredState` property.  Unfortunately, it looks like [only the archived registry resource](https://github.com/search?q=repo%3APowerShell%2FDSC+_inDesiredState+language%3ARust&type=code&l=Rust) uses it so there isn't an easy repro.

The fragment from my resource manifest:
```YAML
  "test" : 
  {
    "args" : 
    [
      "dscv3",
      "test-file",
      "--test"
    ],
    "executable" : "wingetdev",
    "input" : "stdin",
    "return" : "state"
  },
```

By allowing a default value for one of the resource properties, the desired state to actual state diff will be empty.  Since [the code only uses whether `diff_properties` is empty](https://github.com/PowerShell/DSC/blob/416c770e25b03c83c2c64e9508f08262924d8102/dsc_lib/src/dscresources/command_resource.rs#L290), and only values in the desired state that are different in the actual state are included, the diff is calculated as empty.

Sample output from calling my resource:
```
> @{ path = "E:\Temp\winget-dsc-test.txt" } | ConvertTo-Json | wingetdev dscv3 test-file --test
{"_exist":true,"_inDesiredState":false,"content":"Contents!","path":"E:\\Temp\\winget-dsc-test.txt"}
```

And output from calling `dsc resource test`:
```
> @{ path = "E:\Temp\winget-dsc-test.txt" } | ConvertTo-Json | dsc resource test --resource Microsoft.WinGet/TestFile -f -
desiredState:
  path: E:\Temp\winget-dsc-test.txt
actualState:
[...]

PR #676, opened 2025-03-06
Fix when resource test returns '_inDesiredState', that takes precedence

SteveL-MSFT:
# PR Summary

Even when a resource implement `test` returns `state` and contains `_inDesiredState` property, that value isn't being used and a synthetic comparison still occurs.

With this change, whether a resource returns `state` or `stateAndDiff` and contains `_inDesiredState`, then that value will be used regardless if there are any differing properties.  Diff of properties will still occur for information purposes even if `_inDesiredState = true`

Also fixed a resource string issue where the placeholder wasn't being replaced correctly.

## PR Context

Fix https://github.com/PowerShell/DSC/issues/674


tools/dsctest/src/main.rs
```diff
@@ -65,6 +67,17 @@ fn main() {
...
+        SubCommand::InDesiredState { input } => {
+            let mut in_desired_state = match serde_json::from_str::<in_desired_state::InDesiredState>(&input) {
+                Ok(in_desired_state) => in_desired_state,
+                Err(err) => {
+                    eprintln!("Error JSON does not match schema: {err}");
+                    std::process::exit(1);
+                }
+            };
+            in_desired_state.value = "SomethingElse".to_string();
+            serde_json::to_string(&in_desired_state).unwrap()
```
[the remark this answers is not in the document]
SteveL-MSFT: This is a test executable

tools/dsctest/src/main.rs
```diff
@@ -65,6 +67,17 @@ fn main() {
...
         },
+        SubCommand::InDesiredState { input } => {
+            let mut in_desired_state = match serde_json::from_str::<in_desired_state::InDesiredState>(&input) {
+                Ok(in_desired_state) => in_desired_state,
+                Err(err) => {
+                    eprintln!("Error JSON does not match schema: {err}");
+                    std::process::exit(1);
+                }
+            };
+            in_desired_state.value = "SomethingElse".to_string();
```
tgauth, line 78: For the sake of the test, when `_inDesiredState` is false, should the `in_desired_state.value` stay the same?

SteveL-MSFT: In this case, the intent is to just trust the resource if it includes `_inDesiredState` regardless of the values differ.

SteveL-MSFT:
@tgauth please continue the review

tgauth (approved):

Issue #683, opened 2025-03-08
registry.exe does not allow for empty valueName


Gijsreyn:
### Prerequisites

- [x] Write a descriptive title.
- [x] Make sure you are able to repro it on the latest version
- [x] Search the existing issues.

### Summary

When testing out the following DSC configuration document

```yaml
$schema: https://aka.ms/dsc/schemas/v3/bundled/config/document.json
resources:
- name: Key
  type: Microsoft.Windows/Registry
  properties:
    keyPath: HKCU\1\2\3
    valueName: Key
    _exist: true
```

I was expecting a key to be created in my registry. However, the result from the `dsc config get` didn't return it. That led me to the `registry.exe`, which does not allow for only passing in a `valueName`. 

### Steps to reproduce

1. Create the following configuration document

```yaml
# example.yaml
$schema: https://aka.ms/dsc/schemas/v3/bundled/config/document.json
resources:
- name: Key
  type: Microsoft.Windows/Registry
  properties:
    keyPath: HKCU\1\2\3
    valueName: Key
    _exist: true
```

2, Run `dsc config set --file example.yaml`

### Expected behavior

```console
Registry key created in the expected path.
```

### Actual behavior

```console
No registry key is created.
```

### Error details

```console

```

### Environment data

```PowerShell
Name                           Value
----                           -----
PSVersion                      7.4.7
PSEdition                      Core
GitCommitId                    7.4.7
OS                             Microsoft Windows 10.0.19045
Platform                       Win32NT
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0…}
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
WSManStackVersion              3.0
```

### Version

dsc 3.1.0-preview.1

SteveL-MSFT:
So the intent is to create a value called `Key` under that keyPath as REG_NONE?

Gijsreyn:
Perhaps I'm looking it through the wrong glasses, but I would expect the same behaviour somewhat as doing:

```powershell
New-ItemProperty -Path 'HKCU\1\2\3' -Name 'Key' 
```

In this case, the default value is of type `REG_SZ`. I don't mind if the default would be `REG_NONE`. The thing now is, whenever executing the above example through `dsc.exe`, test will keep returning `False`, which shouldn't be the case if the `registry.exe` has exited with 0.

PR #684, opened 2025-03-10
Fix `registry` resource to allow setting value with no data

SteveL-MSFT:
# PR Summary

- Allow setting registry value without data by using RZ_NONE type, however, the `valueData` member that is returned won't be present when RZ_NONE is used.
- Fix issue in build script that doesn't copy symbols to bin folder making it so you can't debug
- Fix small issues in i18n text so that it gets rendered correctly
- Fixed issue in helper function where if you have just the hive and one subkey (that doesn't exist yet), it incorrectly returns nothing for subkeys and thus doesn't get created
- Add tests

## PR Context

Fix https://github.com/PowerShell/DSC/issues/683

### DIFF
--- a/registry/src/registry_helper.rs
+++ b/registry/src/registry_helper.rs
@@ -72,7 +72,10 @@ impl RegistryHelper {
             Ok(Registry {
                 key_path: self.config.key_path.clone(),
                 value_name: Some(value_name.clone()),
-                value_data: Some(convert_reg_value(&value)?),
+                value_data: match value {
+                    Data::None => None,
+                    _ => Some(convert_reg_value(&value)?)
+                },
                 ..Default::default()
             })
         } else {
@@ -100,7 +103,7 @@ impl RegistryHelper {
                     };
 
                     if self.what_if {
-                        what_if_metadata.push(t!("registry_helper.whatIfCreateKey", "subkey" => subkey).to_string());
+                        what_if_metadata.push(t!("registry_helper.whatIfCreateKey", subkey = subkey).to_string());
                     }
                     else {
                         reg_key = reg_key.create(path, Security::CreateSubKey)?;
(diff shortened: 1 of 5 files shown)


registry/src/registry_helper.rs
```diff
@@ -72,7 +72,10 @@ impl RegistryHelper {
             Ok(Registry {
                 key_path: self.config.key_path.clone(),
                 value_name: Some(value_name.clone()),
-                value_data: Some(convert_reg_value(&value)?),
+                value_data: match value {
+                    Data::None => None,
```
tgauth, line 76: Is this `match` statement necessary? Looks like `convert_reg_value()` has been updated to handle the `Data::None` case.

registry/src/registry_helper.rs
```diff
@@ -116,8 +119,13 @@ impl RegistryHelper {
             Err(e) => return self.handle_error_or_what_if(e)
         };
 
-        if let Some(value_data) = &self.config.value_data {
-            let Ok(value_name) = U16CString::from_str(self.config.value_name.as_ref().unwrap()) else {
+        if let Some(value_name) = &self.config.value_name {
+            let value_data = match &self.config.value_data {
```
tgauth, line 123: same comment as above

registry/src/registry_helper.rs
```diff
@@ -153,12 +161,18 @@ impl RegistryHelper {
...
+                RegistryValueData::None => {
+                    Data::None
+                },
             };
 
             if self.what_if {
                 return Ok(Some(Registry {
                     key_path: self.config.key_path.clone(),
-                    value_data: Some(convert_reg_value(&data)?),
+                    value_data: match data {
```
tgauth, line 172: also same as above

registry/src/registry_helper.rs
SteveL-MSFT: Not exactly, but you bring up a good point, `convert_reg_value()` should return an `Option` instead and this match wouldn't be needed.  I'll change that.

tgauth (approved):

PR #688, opened 2025-03-13
Emit warning if executable in resource manifest can't be found

SteveL-MSFT:
# PR Summary

- Update build script to not copy WindowsPowerShell resource manifest on non-Windows, creates noise for testing
- During resource discovery when loading the manifest, validate that the executable specified for the operation can be found otherwise emit a warning message
- Remove outdated comment that `get` is always supported

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -380,7 +380,7 @@ if (!$SkipBuild) {
             if ($IsWindows) {
                 Copy-Item "*.dsc.resource.json" $target -Force -ErrorAction Ignore
             }
-            else { # don't copy WindowsPowerShell resource maisource manifest
+            else { # don't copy WindowsPowerShell resource manifest
                 Copy-Item "*.dsc.resource.json" $target -Exclude 'windowspowershell.dsc.resource.json' -Force -ErrorAction Ignore
             }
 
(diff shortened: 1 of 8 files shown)


build.ps1
```diff
@@ -377,7 +377,13 @@ if (!$SkipBuild) {
                 }
             }
 
-            Copy-Item "*.dsc.resource.json" $target -Force -ErrorAction Ignore
+            # skip copying WindowsPowerShell resource on non-Windows
+            if (!$IsWindows) {
```
tgauth, line 381: nit: could skip the negation and just use `if ($IsWindows)` by flipping the contents of the if/else statement

SteveL-MSFT: good point, will flip

build.ps1
```diff
@@ -377,7 +377,12 @@ if (!$SkipBuild) {
                 }
             }
 
-            Copy-Item "*.dsc.resource.json" $target -Force -ErrorAction Ignore
+            if ($IsWindows) {
+                Copy-Item "*.dsc.resource.json" $target -Force -ErrorAction Ignore
+            }
+            else { # don't copy WindowsPowerShell resource maisource manifest
```
tgauth, line 383:
```suggestion
            else { # don't copy WindowsPowerShell resource manifest
```

tgauth (approved):

PR #699, opened 2025-03-14
Add better tracing and fix propagation of inDesiredState for PSAdapter

SteveL-MSFT:
# PR Summary

- Tracing wasn't correct due to case-sensitivity of JSON and some traces were being written directly instead of through the helper
- Simple tracing should be lowercase and be `warn` instead of `warning`
- Add `_inDesiredState` property to the result object for `test` if any of the results contain `InDesiredState` and only true if all are true


SteveL-MSFT:
Need to make one more change to rename `InDesiredState` to `_inDesiredState` at the resource level to keep it consistent

tgauth (approved):

Issue #693, opened 2025-03-13
Not implemented: Custom resource not supported - Managing a Windows Service


apetitjean:
### Prerequisites

- [x] Write a descriptive title.
- [x] Make sure you are able to repro it on the latest version
- [x] Search the existing issues.

### Summary

I'm new to DSC v3 but not to DSC. I'm on Windows Server 2025, using PowerShell 7.5. I created this configuration file : 
```yaml
$schema: https://raw.githubusercontent.com/PowerShell/DSC/main/schemas/2023/08/config/document.json

resources:
  - name: Ensure Spooler Running
    type: PSDesiredStateConfiguration/Service
    adapter: Microsoft.Windows/WindowsPowerShell
    properties:
      Name: Spooler
      State: Running

```

I would like to configure a Windows Service to begin my learning journey of DSC v3 but I have to say, this is not a straightforward path.

Here's the command I type to test (or set) the configuration : 
`dsc config get -f C:\Scripts\DSCv3_configService.yaml`

and here's the error message I'm receiving : 
`2025-03-13T23:24:52.082168Z ERROR Not implemented: Custom resource not supported`

Am I doing something wrong or you have not yet implemented the Windows PowerShell DSC adapter?
Or maybe there's another way to do so? Could you please help me?

Thanks.

### Steps to reproduce

Here's my config :

```yaml
$schema: https://raw.githubusercontent.com/PowerShell/DSC/main/schemas/2023/08/config/document.json

resources:
  - name: Ensure Spooler Running
    type: PSDesiredStateConfiguration/Service
    adapter: Microsoft.Windows/WindowsPowerShell
    properties:
      Name: Spooler
      State: Running
```

Here's the command I type to test (or set) the configuration : 
`dsc config get -f C:\Scripts\DSCv3_configService.yaml`

Error : `2025-03-13T23:24:52.082168Z ERROR Not implemented: Custom resource not supported`

### Expected behavior

```console
I dunno
```

### Actual behavior

```console
`2025-03-13T23:24:52.082168Z ERROR Not implemented: Custom resource not supported`
```

### Error details

```console
2025-03-13T23:24:52.082168Z ERROR Not implemented: Custom resource not supported
```

### Environment data

```PowerShell
PS C:\> $PSVersionTable

Name                           Value
----                           -----
PSVersion                      7.5.0
PSEdition                      Core
GitCommitId                    7.5.0
OS                             Microsoft Windows 10.0.26100
Platform                       Win32NT
[...]

PR #720, opened 2025-03-21
Enable using adapter resources without the adapter wrapper

SteveL-MSFT:
# PR Summary

This is a breaking change, but should not affect anyone as no one has implemented an adapter yet, but we should doc it.

- The previous code had two problems:
  - the change was in `dsc` exe instead of the library so it was only written for direct resource invocation and not for configuration
  - it relied on passing a special property that the adapter would recognize

So all that code was removed.  The old code relied on modifying the input JSON and injecting the special property, so parameters for functions were changed to no longer need the input to be mutable.

Instead, this PR removes all that old code (the breaking change) and instead when a resource is specified whether in a config or direct resource invocation and requires an adapter, then internally it dynamically generates a configuration with that adapter along with the resource and executes that instead.  The resource part of the configuration invocation output is extracted to appear as though just the resource was executed.

In the PS Adapter, I removed the code that checked for the special property so this means adapters should ONLY expect to get configuration and doesn't need to handle whether the input it received was for a resource or a complete configuration.  Also, to improve error messages, added `-ErrorAction Stop` to use of `Invoke-DscResource` so that the correct error surfaces rather than a later error.

WMI Adapter updated to also only expect configuration.  TestAdapter fixed as export requires an array of instances of properties representing the resources.

## PR Context

Fix https://github.com/PowerShell/DSC/issues/693


SteveL-MSFT:
One problem is that if both WinPS and PS7 adapters support the same resource, the first one found is used.  This is generally not a problem, but when using `export`, it has to be the PS7 one since WinPS doesn't support `export`, but if the WinPS one is returned, then the operation fails.

michaeltlombardi:
> One problem is that if both WinPS and PS7 adapters support the same resource, the first one found is used. This is generally not a problem, but when using `export`, it has to be the PS7 one since WinPS doesn't support `export`, but if the WinPS one is returned, then the operation fails.

Maybe we can further filter the results by whether they support the current operation, or consider (eventually?) defining a default adapter or an adapter priority in the DSC settings?

For example, in the case of `export` where we filter results by operation:

1. Find all resources with the given name.
2. If there's more than one resource, filter out resources missing the appropriate capability.
3. Use the default selection for multiple resources with the given name.

That _could_ be implemented independently of any settings value that defines adapter priority.

SteveL-MSFT:
Rather than make this PR even bigger, I'm going to fix up the tests and defer the discovery to next PR.  For that one, I'll prepare for supporting specifying versions as discovery should take into account version and capability rather than the simple approach right now which is just finding the first type that matches.

powershell-adapter/psDscAdapter/psDscAdapter.psm1
```diff
@@ -380,14 +380,8 @@ function Get-DscResourceObject {
...
     else {
-        # mimic a config object with a single resource
-        $type = $inputObj.adapted_dsc_type
-        $inputObj.psobject.properties.Remove('adapted_dsc_type')
-        $desiredState += [dscResourceObject]@{
-            name       = $adapterName
-            type       = $type
-            properties = $inputObj
-        }
+        Write-DscTrace -Operation Error -Message 'Unexpected input format. The input must be a configuration object.'
```
SteveL-MSFT, line 383: Remove the `context` property within Microsoft.DSC metadata as it's no longer needed

powershell-adapter/Tests/TestAdapter/testadapter.resource.ps1
```diff
@@ -48,16 +48,19 @@ switch ($Operation) {
...
-    { @('Get','Set','Test','Export') -contains $_ } {
+    { @('Get','Set','Test') -contains $_ } {
+        "Operation: $Operation" | Write-DscTrace
 
-        # TestCase 1 = 'Verify adapted_dsc_type field'
-        if (($inputobj.TestCaseId -eq 1 ) -or ($_ -eq 'Export')){
-            $result = $inputobj.adapted_dsc_type -eq 'Test/TestCase'
-            $result = @{'TestCaseId'=1; 'Input'=''; result = $result } | ConvertTo-Json -Depth 10 -Compress
-            return $result
+        if (($inputobj.resources.properties.TestCaseId -eq 1 ) -or ($_ -eq 'Export')){
```
tgauth, line 54:
```suggestion
        if ($inputobj.resources.properties.TestCaseId -eq 1) {
```
Is the Export check still needed if that test case is handled below?

powershell-adapter/Tests/powershellgroup.resource.tests.ps1
```diff
@@ -70,10 +70,11 @@ Describe 'PowerShell adapter resource tests' {
         $r = "{'Name':'TestClassResource1','Prop1':'ValueForProp1'}" | dsc resource set -r 'TestClassResource/TestClassResource' -f -
         $LASTEXITCODE | Should -Be 0
         $res = $r | ConvertFrom-Json
-        $res.afterState.result | Should -Not -BeNull
+        $res.afterState.Prop1 | Should -BeExactly 'ValueForProp1'
+        $res.changedProperties | Should -BeNullOrEmpty
     }
 
-    It 'Export works on PS class-based resource' {
+    It 'Export works on PS class-based resource' -Pending {
```
tgauth, line 77: Is this `-Pending` because of the discovery work that's still needed? Can we add a comment / `TODO` here as well and for L.94?

SteveL-MSFT: Yes, it's pending as there will be a follow-up PR to address discovery for capabilities instead of first one.  I don't think TODO is needed as `Pending` means TODO and I plan to start on it after this is merged.

powershell-adapter/Tests/TestAdapter/testadapter.resource.ps1
SteveL-MSFT: Good catch, that check is no longer needed

PR #734, opened 2025-04-02
(SCHEMAS) Fix/update for v3.0.0 release

michaeltlombardi:
# PR Summary

This change set includes updates to the schema source files for various issues with the canonically published schemas, including:

- Mismatched terminology
- Erroneous `minItems` keyword for `resources` property of coniguration document
- Incorrect enumeration values for resource `kind`
- Incorrect description and default definition for `returnKind` for `set`, requiring updating shared definition, `set.return`, `whatIf.return`, `test.return`, and related documentation keywords.
- Incorrect URIs for `$schema` property in both configuration document and resource manifest schemas

This change also adds JSON schemas for the expected `stdout` JSON for various resource operations. This should help resource authors understand and validate their resource implementations.

TODO:

- [x] Regenerate schemas from source
- [ ] ~~Update and extend reference documentation~~ (deferred to separate PR)

## PR Context

Needed to address issues with the published schemas for the `v3.0.0` release. These changes and the technical review of them are required for completing the regeneration and documentation efforts.


michaeltlombardi:
@SteveL-MSFT - keeping this in draft, but it's ready for review - need to regenerate the schemas and add docs, but only want to do that after we're sure the actual schema implementations is accurate.

SteveL-MSFT (approved):
Nothing stands out to me as an issue

Issue #731, opened 2025-03-30
Assertion not breaking when dependsOn is specified


Gijsreyn:
### Prerequisites

- [x] Write a descriptive title.
- [x] Make sure you are able to repro it on the latest version
- [x] Search the existing issues.

### Summary

As per discussion #725, grabbing the example from: https://github.com/PowerShell/DSC/blob/main/dsc/examples/assertion.dsc.yaml doesn't work.

### Steps to reproduce

1. Create the configuration document from: https://github.com/PowerShell/DSC/blob/main/dsc/examples/assertion.dsc.yaml
2. Run `dsc config set --file assertion.dsc.yaml`

### Expected behavior

```console
Expect `current user registry` resource not to run
```

### Actual behavior

```console
`current user registry` runs
```

### Error details

```console
None
```

### Environment data

```PowerShell
Name                           Value
----                           -----
PSVersion                      7.4.7
PSEdition                      Core
GitCommitId                    7.4.7
OS                             Microsoft Windows 10.0.19045
Platform                       Win32NT
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0…}
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
WSManStackVersion              3.0
```

### Version

3.0.0

MattHyman:
I have also encountered this issue.

PR #736, opened 2025-04-04
Fix `Assertion` resource to fail when test fails

SteveL-MSFT:
# PR Summary

A previous PR that cleaned up the output of `Assertion` to fit with `get` and `set` caused the assertion to never actually fail, just return results in the desired format.

The change here is to add a new hidden switch `--as-assert` so that `dsc` knows it's running as `Assertion`.  Then when this is used it checks each test result and if any are NOT in desired state, it returns an error that the assertion failed.  Many of the changes is just to pass this parameter to the necessary functions.  A helper function is used to potentially recursively look at all the `inDesiredState` properties.

image

## PR Context

Fix https://github.com/PowerShell/DSC/issues/731

### DIFF
--- a/dsc/tests/dsc_assertion.tests.ps1
+++ b/dsc/tests/dsc_assertion.tests.ps1
@@ -1,7 +1,7 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT License.
 
-Describe 'Assertion resource tests' {
+Describe 'Assertion resource tests' -Skip:(!$IsWindows) {
     It 'Example works for <operation>' -TestCases @(
         @{ operation = 'get' }
         @{ operation = 'set' }
--- a/dsc/tests/dsc_assertion.tests.ps1
+++ b/dsc/tests/dsc_assertion.tests.ps1
@@ -1,0 +1,20 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+Describe 'Assertion resource tests' {
+    It 'Example works for <operation>' -TestCases @(
+        @{ operation = 'get' }
+        @{ operation = 'set' }
+        @{ operation = 'test' }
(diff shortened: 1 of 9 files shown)


dsc/tests/dsc_assertion.tests.ps1
```diff
@@ -0,0 +1,20 @@
...
+        @{ operation = 'get' }
+        @{ operation = 'set' }
+        @{ operation = 'test' }
+# TODO: Add export to test when https://github.com/PowerShell/DSC/issues/428 is fixed
+#        @{ operation = 'export' }
+    ) {
+        param($operation)
+        $jsonPath = Join-Path $PSScriptRoot '../examples/assertion.dsc.yaml'
+        $out = dsc config $operation -f $jsonPath 2> "$TestDrive/trace.log"
+        $LASTEXITCODE | Should -Be 2
```
tgauth, line 15: Why is it expected that the exit code will be 2 i.e. `EXIT_DSC_ERROR` instead of 8 i.e. `EXIT_DSC_ASSERTION_FAILED` here?

SteveL-MSFT: You're right, this should be 8

SteveL-MSFT: Actually, the 8 is coming from the child `dsc` process, but the parent one is returning `2` is which the DSC Error code.  So this is correct as we don't propagate the child exit code to the parent.

tgauth (approved):

dsc/tests/dsc_assertion.tests.ps1
SteveL-MSFT: Created https://github.com/PowerShell/DSC/issues/737 as it's not a simply change, but something we should eventually support

Issue #745, opened 2025-04-08
win_psDscAdapter fails to find resources during config run


haodeon:
### Prerequisites

- [x] Write a descriptive title.
- [x] Make sure you are able to repro it on the latest version
- [x] Search the existing issues.

### Summary

During `config set` if a resource modifies PS modules, for example installing IIS also installs the WebAdministration module into `C:\Windows\System32\WindowsPowerShell\v1.0\Modules`, this causes the `win_psDscAdapter` to refresh the cache during the run.

The cache refresh fails to keep certain modules in the cache. This causes a module not found error for later resources.

Subsequent `config set` runs continue to fail as it continues to keep using the incorrectly cached information.

Removing the cache manually and calling `dsc resource list --adapter Microsoft.Windows/WindowsPowerShell` correctly rebuilds the cache. If the `resource list` step is omitted and `config set` is run directly after removing the cache, it continues to rebuild the cache incorrectly.

In my testing it fails to find/cache the following modules

- PSDesiredStateConfiguration v1.1
- SChannelDsc
- WebAdministrationDsc

For some reason xPSDesiredStateConfiguration is not affected and will reliably remain in the cache.

### Steps to reproduce

1. **Ensure IIS and the WebAdministration module is not installed**
 
2. Create the following configuration document:
```yaml
$schema: https://raw.githubusercontent.com/PowerShell/DSC/main/schemas/v3.0.0/config/document.json
resources:
  - name: Install IIS
    type: Microsoft.Windows/WindowsPowerShell
    properties:
      resources:
        - name: Install IIS
          type: xPSDesiredStateConfiguration/xWindowsFeature
          properties:
            Name: Web-Server
            Ensure: Present
  - name: Web server installed
    type: Microsoft.DSC/Assertion
    properties:
      $schema: https://aka.ms/dsc/schemas/v3/bundled/config/document.json
      resources:
        - name: Use powershell adapter
          type: Microsoft.Windows/WindowsPowerShell
          properties:
            resources:
              - name: IIS present
                type: xPSDesiredStateConfiguration/xWindowsFeature
                properties:
                  Name: Web-Server
                  Ensure: Present
    dependsOn:
      - "[resourceId('Microsoft.Windows/WindowsPowerShell', 'Install IIS')]"
  - name: Harden IIS
    type: Microsoft.Windows/WindowsPowerShell
    properties:
      resources:
        - name: Disable directory browsing
          type: WebAdministrationDsc/WebConfigProperty
          properties:
[...]

PR #748, opened 2025-04-11
Fix resourceCache to return proper modules when PS modules are installed during run

Gijsreyn:
# PR Summary

This PR addresses the issue described in #745. When modules are installed during a document being applied, the subsequent modules that are checked in the cache are overwritten. Example:

1. The first run initiates the resource cache (dsc resource list) as no cache is found.
2. The first configuration is applied, installing new PowerShell modules (triggering the cache on module paths to be stale).
3. The assertion triggers the `Invoke-DscCacheRefresh -Module '<module>'`, flushing out the cache with only that particular module.
4. Modules underneath aren't found anymore in the same sequence.

It's a pretty rough edge case and quite interesting to debug. Nevertheless, there isn't any test added because it would either introduce an external dependency or require internal work to be implemented. Of course, I can be wrong on this and would appreciate some help to write one.


Gijsreyn:
@SteveL-MSFT, if you have some time, you might taking a glimpse?

SteveL-MSFT (changes requested):
Can you repro the original issue by having a test case that executes a test resource that adds a new test resource?

.vscode/launch.json
```diff
@@ -1,39 +1,39 @@
...
-            "name": "Debug sshdconfig",
-            "type": "cppvsdbg",
-            "request": "launch",
-            "program": "${workspaceFolder}/sshdconfig/target/debug/sshdconfig.exe",
-            "args": ["get"],
-            "cwd": "${workspaceFolder}"
-        }
-    ]
-}
+// {
```
SteveL-MSFT, line 1: Why was this included and commented out?

powershell-adapter/Tests/win_powershellgroup.tests.ps1
```diff
@@ -25,67 +25,154 @@ Describe 'WindowsPowerShell adapter resource tests - requires elevated permissio
...
-
-        $r = dsc resource list --adapter Microsoft.Windows/WindowsPowerShell
-        $LASTEXITCODE | Should -Be 0
-        $resources = $r | ConvertFrom-Json
-        ($resources | ? {$_.Type -eq 'PSDesiredStateConfiguration/File'}).Count | Should -Be 1
-    }
-
-    It 'Get works on Binary "File" resource' -Skip:(!$IsWindows){
-
+    # It 'Windows PowerShell adapter supports File resource' -Skip:(!$IsWindows){
```
SteveL-MSFT, line 28: Why are these tests commented out?

SteveL-MSFT (changes requested):

powershell-adapter/Tests/win_powershellgroup.tests.ps1
```diff
@@ -82,12 +81,100 @@ Describe 'WindowsPowerShell adapter resource tests - requires elevated permissio
...
+    properties: 
+      Name: NuGet
+    dependsOn:
+      - "[resourceId('Microsoft.Windows/WindowsPowerShell', 'File')]"
+      - "[resourceId('Microsoft.DSC/Assertion', 'File present')]"
+"@
+        # output to file for Windows PowerShell 5.1
+        $filePath = "$testdrive\test.assertion.dsc.resource.yaml"
+        $yaml | Set-Content -Path $filePath -Force
+        dsc config test -f $filePath 2> "$TestDrive/error.txt"
```
SteveL-MSFT, line 170: No change needed but more for your info, you can always pass in `$yaml` to `--input` instead of writing it as a file:

```powershell
dsc config test -i $yaml
```

SteveL-MSFT (approved):

powershell-adapter/Tests/win_powershellgroup.tests.ps1
Gijsreyn: I expected it also, yet, there was an issue: #752.

Issue #328, opened 2024-02-25
Support for Credentials


rdbartram:
### Summary of the new feature / enhancement

In order to be able to call certain types of resources, credential objects are required. I would be very useful to have an intuitive way to do this.

```yaml
# example.dsc.config.yaml
$schema: https://raw.githubusercontent.com/PowerShell/DSC/main/schemas/2023/10/config/document.json
resources:
  - name: Get info from classic DSC resources
    type: DSC/PowerShellGroup
    properties:
      resources:
      - name: Get Teams App Permission Policy
        type: Microsoft365DSC/MSFT_TeamsAppPermissionPolicy
        properties:
          Identity: Global
          Credential:
            userName: "admin@test.onmicrosoft.com"
            Password: "MyPassword"
```

Being able to define and then reference would be great but maybe some interp. from dsc exec itself to get values from environment or a secret manager would be great too

SteveL-MSFT:
Since credentials are sensitive, we wouldn't want the password to be in cleartext within the configuration itself.  Currently, you can use a `SecretString` parameter to pass it to the configuration and reference it via the `parameter()` function.  It would look something like:

```yaml
# example.dsc.config.yaml
$schema: https://raw.githubusercontent.com/PowerShell/DSC/main/schemas/2023/10/config/document.json
parameters:
  globalCredential:
    type: secureObject
resources:
  - name: Get info from classic DSC resources
    type: DSC/PowerShellGroup
    properties:
      resources:
      - name: Get Teams App Permission Policy
        type: Microsoft365DSC/MSFT_TeamsAppPermissionPolicy
        properties:
          Identity: Global
          Credential:
            userName: "[parameters('globalCredential').username)]"
            Password: "[parameters('globalCredential').password]"
```

Where `globalCredential` would look like:

```json
{
  "username": "admin@test.onmicrosoft.com",
  "password": "MyPassword"
}
```

So the parameters input or file would look like:

```yaml
parameters:
  globalCredential:
    username: admin@test.onmicrosoft.com
    password: MyPassword
```

And you would call dsc using:

```powershell
dsc config -f ./globalCredentials.parameters.yaml get -p ./example.dsc.config.yaml
```

where `-f` specifies the file path to the parameters to be used and `-p` to the configuration file.

Gijsreyn:
@SteveL-MSFT unfortunately, the adapter doesn't support it because it doesn't cast the `Credential` to a `PSCredential` object. For example, running the following configuration:

```yaml
[...]

PR #758, opened 2025-04-21
Add credential support on PowerShell adapters

Gijsreyn:
# PR Summary

This PR addresses #328. It adds a like check for credentials and searches the username and password fields as input.

## PR Context

PowerShell adapters currently don't support creating PSCredential objects. Most DSC community-based DSC resources have the $Credential as input. Some others may vary; that's why the `-Like` operator is used.


michaeltlombardi (commented):
This implementation seems to use the following snippet to determine whether to convert the input to a credential:

```powershell
$_.Name -like '*Credential*' -and
$_.Value.Username -and
$_.Value.Password
```

However, I think we should probably consider checking the _target property_ for whether it's expecting a credential or a secure string and _then_ trigger handling.

Gijsreyn:
If so, should we still check if the username/password is provided in such a way?

Gijsreyn:
Just rewrote the part, but I'm struggling a little with the following error message from the pipeline:

Exception calling "Invoke" with "4" argument(s): "The 'ConvertTo-SecureString' command was found in the module 
2025-04-21T14:56:29.911181Z TRACE dsc_lib::dscresources::command_resource: 890: PID 6992: 'Microsoft.PowerShell.Security', but the module could not be loaded. For more information, run 'Import-Module 
2025-04-21T14:56:29.911238Z TRACE dsc_lib::dscresources::command_resource: 890: PID 6992: Microsoft.PowerShell.Security'."

The test I wrote earlier for Windows PowerShell seems to fail on this part. I have commented it out for now. I have seen this issue before with other modules that run under the agent and wasn't able to find a workaround for it. Any suggestions are welcome.

Gijsreyn:
@SteveL-MSFT or @michaeltlombardi Do you perhaps have time to take another stab at this PR? The logic has changed to validate against the target property that is passed in.

powershell-adapter/psDscAdapter/psDscAdapter.psm1
```diff
@@ -416,18 +416,23 @@ function Invoke-DscOperation {
...
-                                $_.Value.psobject.properties | ForEach-Object -Begin {
-                                    $propertyHash = @{}
-                                } -Process {
-                                    $propertyHash[$_.Name] = $_.Value
-                                } -End {
-                                    $dscResourceInstance.$($_.Name) = $propertyHash
+                                $validateProperty = $cachedDscResourceInfo.Properties | Where-Object -Property Name -EQ $_.Name
+                                if ($validateProperty.PropertyType -eq 'PSCredential') {
+                                    if (-not $_.Value.Username -and -not $_.Value.Password) {
+                                        "Credential property '$($_.Name)' requires both username and password input object" | Write-DscTrace -Operation Error
```
SteveL-MSFT, line 429:
```suggestion
                                        "Credential object '$($_.Name)' requires both 'username' and 'password' properties" | Write-DscTrace -Operation Error
```

powershell-adapter/Tests/powershellgroup.config.tests.ps1
```diff
@@ -237,4 +237,22 @@ Describe 'PowerShell adapter resource tests' {
       }
     }
   }
+
+  It 'Config works with credential object' {
```
SteveL-MSFT, line 241: Can you have test cases that explicitly cover both ps adapter and winps adapter since both code was changed?  Also cover the case where either `username` and/or `password` is missing

SteveL-MSFT (changes requested):

powershell-adapter/Tests/powershellgroup.config.tests.ps1
Gijsreyn: Done! I have added a mock for Windows PowerShell. Please take a look at the comments for why. Could you please assist me in resolving the artifact retrieval issue for the Windows build?

SteveL-MSFT: Done

Gijsreyn:
Finally, sorry for the extensive commits @SteveL-MSFT , but it should be good to go now!

SteveL-MSFT (approved):

Issue #681, opened 2025-03-07
Discovery extension type resource


SteveL-MSFT:
### Summary of the new feature / enhancement

One of the challenges currently is that if an appx store app includes a DSC resource, the permissions on the folder that actually stores the app and its files don't allow discovery (enumeration) of the contents.  So this currently prevents `dsc` from finding DSC resources that are part of appx apps.

### Proposed technical implementation details (optional)

Instead of a resource, we introduce a new thing called an `extension` which would have a `.dsc.extension.json` manifest.

For this scenario, it would implement a `discover` method.  As part of `dsc` discovery, any extensions implementing this method would be called.  There would be a property in this method `return` which is either `path` or `manifest`.

If `path` then the extension returns this JSON:

```json
{
  "path": "c:\\program files\\foo\\bar.dsc.resource.json"
}
```

Which is an absolute path to the DSC resource manifest file.

Alternateively, `manifest` would return the contents of the manifest as a JSONLine.

In the appx case, however, since appx doesn't allow discovery (enumeration) in the app folder, we always need absolute paths including the exe that would be called.  So in the `path` case, we might need an additional boolean property `resolveCommandToPath` that informs `dsc` whether the command in the manifest should be resolved to the same directory as the manifest.  In this example, assume that `bar.dsc.resource.json` indicates the `command` is `bar`, so `dsc` would then be calling `c:\\program files\foo\\bar.exe`.

michaeltlombardi:
Is this intended to be compatible with the proposal in #461, or separate? Might require some consideration as the term `extension` gets overloaded.

There's also the proposal in #611 around _extended resources_

Making sure I understand this proposal, I might define a manifest named `appx.dsc.extension.json` with something like the following in it:

```jsonc
// C:\\some\\path\\appx.dsc.extension.json
{
  "discover": {
    "command": "app"
    "args": ["discover"]
  }
}
```

And DSC would find that manifest, then call `app discover` to process the emitted stdout JSON Lines like:
```jsonc
{ "path": "C:\\path\\to\\foo.dsc.resource.json" }
{ "manifest": { } // manifest elided
```

SteveL-MSFT:
@michaeltlombardi I think it would be a specific type of extension to #461, so we'd want to have the same model for any type of extension.  I'm using extension as meaning something that extends the capability of `dsc` and not used as part of configuration itself (like a resource).

Yes, your understanding of what's being proposed here is correct.

SteveL-MSFT:
[...]

PR #760, opened 2025-04-24
Add support for discovery extensions

SteveL-MSFT:
# PR Summary

- Add new subcommand to `dsc` to list extensions: `dsc extension list`
- Add new ExtensionManifest which is similar but differs from ResourceManifest.  Only `discover` capability currently supported.
- Enhance existing discovery code to handle both resources and extensions.  This includes pre-discoverying extensions when discovering resources since an extension may participate.  This included a new enum to handle returning either resources or extensions.
- Write extensions as table when used interactively
- Rename existing functions to generalize for both resource and extension use, this also required making some private functions public

The output from a discovery extension must match:

```json
{ "manifestPath": "<absolute path>" }
```

## PR Context

Fix https://github.com/PowerShell/DSC/issues/681


dsc_lib/src/extensions/discover.rs
```diff
@@ -0,0 +1,20 @@
...
+pub struct DiscoverMethod {
+    /// The command to run to get the state of the resource.
+    pub executable: String,
+    /// The arguments to pass to the command to perform a Get.
+    pub args: Option<Vec<ArgKind>>,
+}
+
+#[derive(Debug, Clone, PartialEq, Deserialize, Serialize, JsonSchema)]
+pub struct DiscoverResult {
+    #[serde(rename = "resourceManifestPath")]
```
SteveL-MSFT, line 18: add comment that this needs to be absolute

dsc/tests/dsc_extension_discover.ps1
```diff
@@ -0,0 +1,93 @@
+# Copyright (c) Microsoft Corporation.
```
SteveL-MSFT, line 1: this should be `dsc_extension_discover.tests.ps1`

michaeltlombardi:
Currently, the implementation requires returning the absolute path to a _resource manifest_ - but I
can already see use cases for discovering _extension_ manifests. For example, I might have
installed an extension from WinGet, and need to discover that, too.

Right now, it expects JSON like this:

```json
{"resourceManifestPath":"/foo/bar/baz.dsc.resource.json"}
```

But we could have it return this, and parse the last segment of the path to determine kind:

```json
{"absoluteManifestPath":"/foo/bar/baz.dsc.extension.json"}
```

Or require it return either `resourceManifestPath` or `extensionManifestPath`.

If we support extensions discovering extensions, we would need to recursively handle extension
discovery.

dsc_lib/src/discovery/command_discovery.rs
```diff
@@ -27,10 +27,20 @@ use which::which;
 use crate::util::get_setting;
 use crate::util::get_exe_path;
 
+const DSC_RESOURCE_EXTENSIONS: [&str; 3] = [".dsc.resource.json", ".dsc.resource.yaml", ".dsc.resource.yml"];
+const DSC_EXTENSION_EXTENSIONS: [&str; 3] = [".dsc.extension.json", ".dsc.extension.yaml", ".dsc.extension.yml"];
+
+#[derive(Clone)]
+pub enum ManifestResource {
```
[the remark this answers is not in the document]
SteveL-MSFT: I was having a hard time coming up with a name that was more general, but just `Manifest` is not correct, but `DscManifest` is ok.

SteveL-MSFT:
I think we can defer extension discovery til when we need it

SteveL-MSFT:
Renamed from `resourceManifestPath` to just `manifestPath` to reuse extensions in the future.  I don't think we need to prefix with `absolute`

michaeltlombardi (approved):

tgauth (approved):

PR #762, opened 2025-04-24
Experimental Appx resource discovery extension

SteveL-MSFT:
# PR Summary

Experimental discovery extension to find resources that are packaged as part of an Appx package
Also updated discovery to remove duplicates (matching resource type name and version)

### DIFF
--- a/dsc_lib/src/discovery/command_discovery.rs
+++ b/dsc_lib/src/discovery/command_discovery.rs
@@ -306,6 +306,11 @@ impl ResourceDiscovery for CommandDiscovery {
                     }
                 }
                 self.adapters = adapters;
+                // go through the resource and remove duplicates with same name and version
+                for (_resource_name, resource_vec) in &mut resources {
+                    let mut uniques: HashSet<String> = HashSet::new();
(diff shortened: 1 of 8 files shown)


dsc_lib/src/discovery/command_discovery.rs
```diff
@@ -306,6 +306,11 @@ impl ResourceDiscovery for CommandDiscovery {
                     }
                 }
                 self.adapters = adapters;
+                // go through the resource and remove duplicates with same name and version
```
tgauth, line 309: Is the intent of `insert_resource()` to take care of duplicates when `skip_duplicate_version` is true (as it is on L.303)?

I think https://github.com/PowerShell/DSC/blob/main/dsc_lib/src/discovery/command_discovery.rs#L570 
should be `if skip_duplicate_version` instead of  `if !skip_duplicate_version`

SteveL-MSFT: Good catch, let me check on this

tgauth (approved):

PR #764, opened 2025-04-25
Change order for winps-adapter to prefer Windows modules first

SteveL-MSFT:
# PR Summary

The WinPS adapter was previously appending the WinPS module path, but instead should be prepending so it gets used first


michaeltlombardi (approved):
LGTM!

PR #769, opened 2025-04-28
Add check in PS Adapter in case proeprty is `$null`

SteveL-MSFT:
# PR Summary

Issue was caught by static analysis.  Ignore the whitespace changes.

There might be a chance that `$validateProperty` is `$null` so need to check that before accessing the member


mgreenegit (approved):
Looks like a simple change. The rest appears to be autoformats.

PR #774, opened 2025-04-30
Fix adapter configuration set which returns `result`, but needs to be `resources`

SteveL-MSFT:
# PR Summary

The problem is that using PS Adapter implicitly means the engine dynamically creates a new configuration that wraps the resource with appropriate adapter (PS7 or WinPS5.1).  This gets run which works fine.  However, a part of `set` also performs a `get` to get the `before_state`.  A `get` returns the payload as `result`, however, the actual `before_state` as defined in the configuration is `resources`.  So the fix is to address this situation where `before_state` needs to be retrieved at runtime and make it look like a correct `set` response.


dsc_lib/src/configure/mod.rs
```diff
@@ -405,12 +405,18 @@ impl Configurator {
                         let GetResult::Resource(after_result) = after_result else {
                             return Err(DscError::NotSupported(t!("configure.mod.groupNotSupportedForDelete").to_string()))
                         };
-                        let before_value = serde_json::to_value(&before_response.actual_state)?;
-                        let after_value = serde_json::to_value(&after_result.actual_state)?;
+                        let diff = get_diff(&before_response.actual_state, &after_result.actual_state);
+                        let mut before: Map<String, Value> = serde_json::from_value(before_response.actual_state)?;
+                        // a `get` will return a `result` property, but an actual `set` will have that as `resources`
+                        if before.contains_key("result") && !before.contains_key("resources") {
+                            before.insert("resource".to_string() ,before["result"].clone());
```
tgauth, line 412:
```suggestion
                            before.insert("resources".to_string() ,before["result"].clone());
```

tgauth (approved):

Issue #707, opened 2025-03-17
WindowsPowerShell adapter should not discover resources unavailable in Windows PowerShell


michaeltlombardi:
### Prerequisites

- [x] Write a descriptive title.
- [x] Make sure you are able to repro it on the latest version
- [x] Search the existing issues.

### Summary

In #698, a user reported an error trying to use a PSDSC resource with the `Microsoft.Windows/WindowsPowerShell` adapter. In that case, the root issue was that even though the adapter discovered a resource, it couldn't be used in the adapter because the PSDSC resource module was installed to a PowerShell module path, not in the Windows PowerShell module path.

The Windows PowerShell adapter should _only_ discover and advertise resources usable in Windows PowerShell.

Checking my `WindowsPSAdapterCache.json` file showed PowerShell module folders alongside Windows PowerShell module folders, indicating the discovery process is somehow including modules that may have PSDSC resources but can't be discovered by the Windows PowerShell adapter and lead to confusing errors.

### Steps to reproduce

1. Invoke `dsc resource list --adapter Microsoft.Windows/WindowsPowerShell`
2. Check the `WindowsPSAdapterCache.json` file for modules not discoverable by Windows PowerShell.

### Expected behavior

```console
The cache only includes modules available in Windows PowerShell.
```

### Actual behavior

```console
The module includes PowerShell module paths as well as Windows PowerShell paths.
```

### Error details

```console

```

### Environment data

```PowerShell
Name                           Value
----                           -----
PSVersion                      7.5.0
PSEdition                      Core
GitCommitId                    7.5.0
OS                             Microsoft Windows 10.0.26100
Platform                       Win32NT
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0…}
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
WSManStackVersion              3.0
```

### Version

3.0.0

Gijsreyn:
I'm just gizzing in the dark here a bit. I was looking at the code, and I think the main responsible commands are:

```powershell
$DscResources = Get-DscResource
$Modules = Get-Module -ListAvailable
```

[...]

PR #777, opened 2025-05-01
For WinPS Adapter remove PS7 paths from PSModulePath

SteveL-MSFT:
# PR Summary

The WinPS adapter is inheriting PS7 paths when `dsc` is run from PS7, this causes WinPS to find incompatible modules.  The fix here is when WinPS adapter is used to explicitly remove PS7 paths from `$env:PSModulePath`.

The other change is something I noticed and I think it was copied from the PS7 adapter but WinPS adapter doesn't need to worry about non-Windows so simplified the code.

## PR Context

Fix https://github.com/PowerShell/DSC/issues/707


mgreenegit (approved):
Great catch

theJasonHelmick (approved):
LGTM

Issue #791, opened 2025-05-08
PS Adapter Export does not serialize enums correctly


ThomasNieto:
### Prerequisites

- [x] Write a descriptive title.
- [x] Make sure you are able to repro it on the latest version
- [x] Search the existing issues.

### Summary

I've added the `static Export()` method to my class-based resources and have `Ensure` as an enum. When running `dsc export` the enum gets converted into the numeric value instead of the string value.

```yaml
$schema: https://aka.ms/dsc/schemas/v3/bundled/config/document.json
contentVersion: 1.0.0
resources:
- type: AnyPackageDsc/Package
  name: AnyPackageDsc/Package-0
  properties:
    result:
    - Version: 3.12.2140.44225
      Prerelease: false
      Provider: AnyPackage.Msi\Msi
      Source: C:\WINDOWS\Installer\3fc3c86b.msi
      Latest: false
      Ensure: 1
      Name: Microsoft Visual Studio Setup WMI Provider
      Reasons: null
      AdditionalParameters: {}
```

### Steps to reproduce

Create DSC class based resource with an enum.

### Expected behavior

```console
Enums use the string value
```

### Actual behavior

```console
Enums use the numeric value
```

### Error details

```console

```

### Environment data

```PowerShell
7.6-preview 4
```

### Version

3.0.1

Gijsreyn:
@ThomasNieto I just submitted a pull request. If you're interested in what was changed, you can always check it out at the above link.

Gijsreyn:
@ThomasNieto I think this is fixed in preview.6.

PR #800, opened 2025-05-09
Add fix to return enum string values

Gijsreyn:
# PR Summary

This PR addresses #791. While it is a small redesign, casting the value to a string makes logical sense. When users are exporting or getting values, they can capture the results and store them as a configuration document instead of having them returned as integers. This would match the schematics, even though that is not implemented _fully_.


Gijsreyn:
@mgreenegit I'm bothering Steve too much, so I thought of giving you a heads-up. Could you take a quick look at this one? It shouldn't be that big of a change.

michaeltlombardi (commented):
Minor nits for readability and spacing characters.

Gijsreyn:
Thanks @michaeltlombardi for the check. I thought there was a remark once to do it in the style Pester tests are written (OTBS). I switched the blocks in `psDscAdapter.psm1` to match that style.

michaeltlombardi (approved):

theJasonHelmick (commented):
LGTM

PR #806, opened 2025-05-10
Add Bicep example

Gijsreyn:
# PR Summary

This PR adds a simple Bicep example

### DIFF
--- a/dsc/examples/file.bicep
+++ b/dsc/examples/file.bicep
@@ -1,0 +1,22 @@
+// This example demonstrates how to create a file using the Windows PowerShell DSC extension.
+// The file is created in the C:\DSC directory on the target machine.
+// You should at least have the Bicep CLI v0.34.34 installed to run this example with experimental feature desiredStateConfiguration turned on.
+
+targetScope = 'desiredStateConfiguration'
+
+resource powerShellAdapter 'Microsoft.Windows/WindowsPowerShell@2025-01-07' = {
+  name: 'Use Bicep to create file'
+  properties: {
+      resources: [
+          {
+              name: 'File'
+              type: 'PSDesiredStateConfiguration/File'
+              properties: {
+                  Ensure: 'Present'
+                  Type: 'Directory'
+                  DestinationPath: 'C:\\DSC\\'
+              }
+          }
+      ]
+  }
+}


SteveL-MSFT:
I'm going to hold off on adding Bicep examples until we make more progress on our Bicep extension and also a DSC extension for Bicep.

Gijsreyn:
With the Bicep extension merged and a request from Jason on #973, I made a slight rename, including a condition. Can you check once again @SteveL-MSFT?

SteveL-MSFT (changes requested):
Can you add a test that uses this file to ensure we don't accidentally break it in the future?

Gijsreyn:
Merged with the latest changes as requested @SteveL-MSFT. Should be good to go :)

dsc/examples/file_with_condition.dsc.bicep
```diff
@@ -0,0 +1,38 @@
...
+  name: 'Use Bicep to create file'
+  properties: {
+      resources: [
+          {
+              name: 'File'
+              type: 'PSDesiredStateConfiguration/File'
+              properties: {
+                  Ensure: 'Present'
+                  Type: 'File'
+                  DestinationPath: 'C:\\DSC\\config.txt'
```
SteveL-MSFT: Instead of hardcoding `C:`, you should use the `systemRoot()` function to get the drive

dsc/examples/file_with_condition.dsc.bicep
```diff
@@ -0,0 +1,38 @@
...
+// The file is created in the C:\DSC directory on the target machine.
+// You should at least have the Bicep CLI v0.34.34 installed to run this example with experimental feature desiredStateConfiguration turned on.
+// To run the second resource, you can add the --parameters '{"parameters":{"restartService":true}}' flag to the command line.
+
+targetScope = 'desiredStateConfiguration'
+
+@description('Set to true to ensure the service is running after the file creation.')
+param restartService bool = false
+
+resource powerShellAdapter 'Microsoft.Windows/WindowsPowerShell@2025-01-07' = {
```
SteveL-MSFT: You should be able to use the File resource implicitly instead of explicitly using the adapter which I think will be more common usage

SteveL-MSFT (changes requested):

theJasonHelmick (approved):
+ Steve's comments - LGTM

dsc/examples/file_with_condition.dsc.bicep
Gijsreyn: Only if that was possible. I have created an [issue](https://github.com/Azure/bicep/issues/17762) on Bicep repository. Do you mind pulling in the example for now?

PR #812, opened 2025-05-14
fix: Check Type instance has DscResource attribute

haodeon:
# PR Summary

PR #797 introduced a workaround to detect if DSC resources were `ClassBased`. This PR adds an additional validation by checking the instance has the `DscResource` attribute. 

## PR Context

The workaround incorrectly sets `ImplementationDetail` to `ClassBased` when there is another type with the same name, like `TimeZone` or `Environment`.

### DIFF
--- a/powershell-adapter/psDscAdapter/win_psDscAdapter.psm1
+++ b/powershell-adapter/psDscAdapter/win_psDscAdapter.psm1
@@ -225,7 +225,8 @@ function Invoke-DscCacheRefresh {
 
             # workaround: Use GetTypeInstanceFromModule to get the type instance from the module and validate if it is a class-based resource
             $classBased = GetTypeInstanceFromModule -modulename $moduleName -classname $dscResource.Name -ErrorAction Ignore
-            if (-not ([string]::IsNullOrEmpty($classBased))) {
+            if ($classBased -and ($classBased.CustomAttributes.AttributeType.Name -eq 'DscResourceAttribute')) {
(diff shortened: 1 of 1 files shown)


Gijsreyn:
Nice catch @haodeon!

powershell-adapter/psDscAdapter/win_psDscAdapter.psm1
```diff
@@ -225,7 +225,9 @@ function Invoke-DscCacheRefresh {
 
             # workaround: Use GetTypeInstanceFromModule to get the type instance from the module and validate if it is a class-based resource
             $classBased = GetTypeInstanceFromModule -modulename $moduleName -classname $dscResource.Name -ErrorAction Ignore
-            if (-not ([string]::IsNullOrEmpty($classBased))) {
+            if ($classBased -and ($instance.GetProperties() | Where-Object {
+                $_.CustomAttributes.AttributeType.Name -eq 'DscPropertyAttribute' } | Select-Object -First 1)) {
+                "Detected class-based resource: $($dscResource.Name) => Type: $($classBased.BaseType.FullName)" | Write-DscTrace
```
michaeltlombardi, line 230: Instead of checking the properties, you can check whether the class has the `[DscResource()]` attribute - any class with that attribute is checked for validity as a PSDSC resource at _parse time_ - those checks (that you have the correct method signatures, at least one key property, etc) are built directly into the parser.

haodeon: Code now checks for `DscResource` attribute instead of scanning properties for `DscProperty` attribute.

Issue #818, opened 2025-05-17
Error while testing the `winps_script.dsc.yaml` example


FlippingBinary:
### Prerequisites

- [x] Write a descriptive title.
- [x] Make sure you are able to repro it on the latest version
- [x] Search the existing issues.

### Summary

I am getting an error "You cannot call a method on a null-valued express" when trying the [`winps_script.dsc.yaml` example](https://github.com/PowerShell/DSC/blob/8fd7ef48f0a1d543a55cc08866b9ebdaf564dd95/dsc/examples/winps_script.dsc.yaml) with the stable release, or a `manifest description: Error` warning and error with the preview release. If I set the `Process` `ExecutionPolicy` to `Bypass`, the preview release gives the same error as the stable release.

I'm including trace outputs from both the stable and preview releases of the CLI to show the differences in impact the `Process` `ExecutionPolicy` has on the two different releases. I installed them with `winget install --id Microsoft.DSC -e` and `winget install --id Microsoft.DSC.Preview -e` (one at a time with the other uninstalled), so those are the names I use in the filenames of the attached trace output files. In each case, I had the `CurrentUser` and `LocalMachine` `ExecutionPolicy` set to `Bypass` (temporarily) and either set the `Process` `ExecutionPolicy` to `Undefined` or `Bypass` to show the impact that has on both releases.

Each trace output was generated with this command: `dsc -l trace config test --file .\winps_script.dsc.yaml`.

[Microsoft.DSC.Preview-Bypass.txt](https://github.com/user-attachments/files/20266177/Microsoft.DSC.Preview-Bypass.txt)
[Microsoft.DSC.Preview-Undefined.txt](https://github.com/user-attachments/files/20266179/Microsoft.DSC.Preview-Undefined.txt)
[Microsoft.DSC-Bypass.txt](https://github.com/user-attachments/files/20266176/Microsoft.DSC-Bypass.txt)
[Microsoft.DSC-Undefined.txt](https://github.com/user-attachments/files/20266178/Microsoft.DSC-Undefined.txt)

### Steps to reproduce

1. Save a local copy of the example script.
2. Call `test` or any of the other methods on it. e.g. `dsc config test --file .\winps_script.dsc.yaml`

### Expected behavior

```console
No errors
```

### Actual behavior

```console
An error: `You cannot call a method on a null-valued expression.`
```

### Error details

```console
[...]

PR #820, opened 2025-05-19
Add executionpolicy WMI

Gijsreyn:
# PR Summary

Set executionpolicy to bypass for WMI DSC resource. Relates to #715.


Gijsreyn:
@SteveL-MSFT , perhaps it was missed, but I think you implemented it only for the PowerShell adapters. Thought of hitting this one up as it was found in #818

SteveL-MSFT (approved):

Issue #814, opened 2025-05-15
`DSC_RESOURCE_PATH` should be prefixed to `PATH`


SteveL-MSFT:
### Summary of the new feature / enhancement

Currently, users can set env var `DSC_RESOURCE_PATH` to override searching `PATH` for resource manifests.  However, the executable found in the resource manifest still relies on `PATH` to find the exe.

### Proposed technical implementation details (optional)

Seems like we should just prefix `DSC_RESOURCE_PATH` to `PATH` when it's defined so that resource manifest is still constrained to the user defined env var and searching for the exe prefers the user defined location.

SteveL-MSFT:
While working on this, I think the intent of `DSC_RESOURCE_PATH` is to allow higher level tools to constrain the search path, but this should include search path for the executables and not just the resource manifests.  As such, this means `PATH` should be overwritten with `DSC_RESOURCE_PATH`.  However, a side effect of this is that any exe that then depends on another exe may not find it unless it's defined within `DSC_RESOURCE_PATH`.  I think we might just need to document this and tooling that wants the prefix behavior should do that within `DSC_RESOURCE_PATH`.  For example, in PowerShell:

```powershell
$env:DSC_RESOURCE_PATH = "foo;bar;$env:PATH"
```

PR #825, opened 2025-05-21
Fix `DSC_RESOURCE_PATH` env var to also limit exe search

SteveL-MSFT:
# PR Summary

The original intent of `DSC_RESOURCE_PATH` was to constrain searching not only manifests but also executables to the paths of that env var.  However, the current code only used it for finding resource manifests and relied on the existing `PATH` to find executables.

The change is to set the process `PATH` env var to same as `DSC_RESOURCE_PATH` such that the original intent is respected.  We should make this clear in the docs.  If the user wants to temporarily add to `PATH` using this env var, they should append `PATH` to it like:

```powershell
$env:DSC_RESOURCE_PATH = "foo;$env:PATH"
```

## PR Context

Fix https://github.com/PowerShell/DSC/issues/814


SteveL-MSFT:
Looks like a bunch of tests relied on the previous behavior and I'll need to fix them

PR #830, opened 2025-05-23
Fix schema image

Gijsreyn:
# PR Summary

PR added to fix the images to match the latest schema, including `document.vscode.json`. This was found from the discussion on #815.


Gijsreyn:
@michaeltlombardi , I tried to get your system settings, but they didn't fully match. I hope the images are good enough, as I kept the size in mind. Perhaps a remark on the example configuration documents is to include the `.vscode` when working in VSCode?

Issue #446, opened 2024-06-05
Add whatif support to PS Adapter


tgauth:
### Summary of the new feature / enhancement

As a PS Adapter user I want to utilize whatif so that I can preview the impact - continuation of #70

Gijsreyn:
@tgauth Hey Tess, I might be a bit eager on the work already on the `winget-dsc` repository, but I already took a shot to implement whatIf for Pip3Dsc. Knowing that this has not yet been developed in `dsc.exe`, I was curious what you think of the current branch implementation on: https://github.com/Gijsreyn/winget-dsc/blob/what-if-support/resources/PythonPip3Dsc/PythonPip3Dsc.psm1

You can see a method whatIf added. The executable that is being executed has the `--dry-run` option available. I'm capturing that output and add it to the `_metaData` property. 

Do you mind having a look at it if you have time? Thanks in advance!

PR #840, opened 2025-05-28
Implement WhatIf PS adapter

Gijsreyn:
# PR Summary

This PR addresses #446 by adding WhatIf capability in the PowerShell adapter


powershell-adapter/powershell.dsc.resource.json
```diff
@@ -1,28 +1,14 @@
...
-          "-Command",
-          "./psDscAdapter/powershell.resource.ps1 List"
-        ]
-        },
-      "config": "full"
-    },
-    "get": {
+  "$schema": "https://aka.ms/dsc/schemas/v3/bundled/resource/manifest.json",
+  "type": "Microsoft.DSC/PowerShell",
+  "version": "0.1.0",
```
SteveL-MSFT, line 4: Since we shipped, we should probably update this to 1.0.0 and increment as we make changes going forward, same for win-ps adapter

SteveL-MSFT (changes requested):
I have a fundamental concern about this although I appreciate the effort.  In PowerShell, scripts/cmdlets can use `ShouldProcess` as a way to get `-whatif` support.  In this PR, you're modeling `whatIf` the same way we support it for native DSC resources.  This ties into the conversation we had this morning about whether we should tell PS resource authors to have a class only for schema and implement functions instead where `ShouldProcess` would work.  Although I suppose we COULD still support allowing a `whatIf` method on a class, but I would rather not create something no one will use.

michaeltlombardi:
Related issue for defining method signatures to support for DSC:

- #860

powershell-adapter/powershell.dsc.resource.json
Gijsreyn: Understood. I haven't tried using the `ShouldProcess`. I wasn't sure if that could be captured into a variable and returned to the DSC engine.

SteveL-MSFT:
The WG is deferring closing on this in regards on how to expose WhatIf to PS class based resources until after we close on the model design for new resource devkit for DSCv3 PS based resources.

Issue #858, opened 2025-06-03
The `RebootPending` resource cannot be used in `Microsoft.DSC/Assertion`


Gijsreyn:
### Prerequisites

- [x] Write a descriptive title.
- [x] Make sure you are able to repro it on the latest version
- [x] Search the existing issues.

### Summary

When using the `RebootPending` DSC resource, it cannot be used in an assertion to check for a pending reboot.

### Steps to reproduce

1. Create the following configuration document:

```yaml
$schema: https://aka.ms/dsc/schemas/v3/bundled/config/document.json
resources:
- name: Managed key
  type: Microsoft.Windows/Registry
  properties:
    _exist: true
    keyPath: HKCU\DscExamples\ManagedKey
  dependsOn:
  - "[resourceId('Microsoft.DSC/Assertion','Assert pending reboot')]"
- name: Assert pending reboot
  type: Microsoft.DSC/Assertion
  properties:
    $schema: https://aka.ms/dsc/schemas/v3/bundled/config/document.json
    resources:
    - name: Check pending reboot
      type: Microsoft.Windows/RebootPending
```

2. Execute the document by running `dsc config test -f pendingReboot.config.dsc.yaml`

An error occurs.

### Expected behavior

```console
The assertion would succeeds.
```

### Actual behavior

```console
2025-06-03T05:26:04.602215Z ERROR trace_message="PID 31604: JSON: EOF while parsing a value at line 1 column 0"
2025-06-03T05:26:04.604962Z ERROR Command: Resource 'dsc' [exit code 2] manifest description: Resource error
```

### Error details

```console

```

### Environment data

```PowerShell
Name                           Value
----                           -----
PSVersion                      7.5.1
PSEdition                      Core
GitCommitId                    7.5.1
OS                             Microsoft Windows 10.0.19045
Platform                       Win32NT
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0…}
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
WSManStackVersion              3.0
```

### Version

dsc 3.1.0-preview.6

Gijsreyn:
Closes as resolved #859

PR #857, opened 2025-06-03
Add RebootPending and WindowsPowerShell reference documentation

Gijsreyn:
# PR Summary

The following PR adds the `Microsoft.Windows/RebootPending` and `Microsoft.Windows/WindowsPowerShell` reference documentation. Minor adjustments made in the `Microsoft.DSC/PowerShell` reference doc.

PR #859, opened 2025-06-03
Convert pending reboot to object

Gijsreyn:
# PR Summary

This PR addresses issue #858. This was found while writing the reference documentation in #857.


Gijsreyn:
@SteveL-MSFT I thought of simply making this an object and removing the read-only property. Perhaps I'm missing something in the schema, but this allows the property to be still optional.

reboot_pending/reboot_pending.dsc.resource.json
```diff
@@ -16,11 +16,10 @@
   "schema": {
     "embedded": {
       "$schema": "https://json-schema.org/draft/2020-12/schema",
-      "type": "null",
+      "type": "object",
       "properties": {
         "rebootPending": {
-            "type": "boolean",
-            "readOnly": true
```
michaeltlombardi, line 23: We don't want to remove the `readOnly` keyword here, I think - it indicates that the property can only be retrieved by the resource, not managed, which is true in this case.

reboot_pending/reboot_pending.dsc.resource.json
```diff
@@ -16,11 +16,10 @@
   "schema": {
     "embedded": {
       "$schema": "https://json-schema.org/draft/2020-12/schema",
-      "type": "null",
+      "type": "object",
```
michaeltlombardi:
```suggestion
      "type": ["object", "null"],
```

In this case, we want to support both not specifying any properties during input (`null`) and the actual object.

reboot_pending/reboot_pending.dsc.resource.json
Gijsreyn: Just tested it, still works!

reboot_pending/reboot_pending.dsc.resource.json
Gijsreyn: Added.

Gijsreyn:
Thanks for the review Mikey. I also added your remark from #857 on exit codes.

reboot_pending/reboot_pending.dsc.resource.json
```diff
@@ -13,16 +13,23 @@
...
-            "type": "boolean",
-            "readOnly": true
+          "type": "boolean",
+          "readOnly": true
         }
       }
     }
   }
-}
+}
```
michaeltlombardi, line 35: Nit:
```suggestion
}
```

michaeltlombardi (approved):

Issue #813, opened 2025-05-14
resources get --all should return a simple array of resources


Bpoe:
### Summary of the new feature / enhancement

The current output of this argument set places a higher burden on higher level tools. A simple array of resources would be more useful and easier to parse.

### Proposed technical implementation details (optional)

Consider this is what I have to do today:

```
// Execute the external command and capture output
var dscCommand = "c:\\bin\\dsc\\dsc.exe";
var dscArgs = $"resource get --all --resource Microsoft.GuestConfiguration/users";

var processInfo = new ProcessStartInfo(dscCommand, dscArgs)
{
    RedirectStandardOutput = true,
    UseShellExecute = false,
    CreateNoWindow = true
};

var process = Process.Start(processInfo);
var output = new List<string>();
while (!process.StandardOutput.EndOfStream)
{
    output.Add(process.StandardOutput.ReadLine());
}

process.WaitForExit();

// Deserialize the JSON output and extract the resources
var resources = new List<Resource>();
foreach (var line in output)
{
    var resource = JsonSerializer.Deserialize<DscResourceResult>(line);
    resources.Add(resource.ActualState);
}

public class Resource
{
    [JsonPropertyName("id")]
    public string Id { get; set; } = string.Empty;

    [JsonPropertyName("name")]
    public string Name { get; set; } = string.Empty;

    [JsonPropertyName("type")]
    public string Type { get; set; } = string.Empty;

    [JsonPropertyName("properties")]
    public object Properties { get; set; } = new object();
}

public class DscResourceResult
{
    [JsonPropertyName("actualState")]
    public ArmResource ActualState { get; set; }
}
```

It would be much easier to do this instead:
```
// Execute the external command and capture output
var dscCommand = "c:\\bin\\dsc\\dsc.exe";
var dscArgs = $"resource get --all --resource Microsoft.GuestConfiguration/users";

var processInfo = new ProcessStartInfo(dscCommand, dscArgs)
{
    RedirectStandardOutput = true,
    UseShellExecute = false,
    CreateNoWindow = true
};

var process = Process.Start(processInfo);
var output = new StringBuilder();
while (!process.StandardOutput.EndOfStream)
{
    output.AppendLine(process.StandardOutput.ReadLine());
}

process.WaitForExit();

// Deserialize the JSON output and extract the resources
var resources = JsonSerializer.Deserialize<List<Resource>>(output);

public class Resource
{
    [JsonPropertyName("id")]
[...]

PR #861, opened 2025-06-04
Add `json-array` as supported format to `get --all`

SteveL-MSFT:
# PR Summary

When using `dsc resource get --all`, `--format-output` now has a `json-array` option.  This is only valid when used with `--all` which returns the results as a single json array rather than multiple JSONLines.

## PR Context

Fix https://github.com/PowerShell/DSC/issues/813

### DIFF
--- a/dsc/locales/en-us.toml
+++ b/dsc/locales/en-us.toml
@@ -73,6 +73,7 @@ implementedAs = "implemented as"
 invalidOperationOnAdapter = "Can not perform this operation on the adapter itself"
 setInputEmpty = "Desired input is empty"
 testInputEmpty = "Expected input is required"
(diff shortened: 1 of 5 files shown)


dsc/locales/en-us.toml
```diff
@@ -73,6 +73,7 @@ implementedAs = "implemented as"
 invalidOperationOnAdapter = "Can not perform this operation on the adapter itself"
 setInputEmpty = "Desired input is empty"
 testInputEmpty = "Expected input is required"
+jsonError = "JSNO: %{err}"
```
tgauth:
```suggestion
jsonError = "JSON: %{err}"
```

PR #865, opened 2025-06-05
Separate registry_helper into lib

tgauth:
# PR Summary
- separates registry_helper.rs from registry resource into a registry_lib 

## PR Context
- allows other resources to utilize registry_lib


sshdconfig/tree-sitter-ssh-server-config/bindings/rust/build.rs
```diff
@@ -0,0 +1,21 @@
+fn main() {
```
SteveL-MSFT: We don't want to check in the bindings, but have them generated each time in case we update to newer versions of tree-sitter, see https://github.com/PowerShell/DSC/blob/main/tree-sitter-dscexpression/tree-sitter.json

SteveL-MSFT (changes requested):

sshdconfig/tree-sitter-ssh-server-config/bindings/rust/build.rs
tgauth: oops, didn't mean to commit those!

SteveL-MSFT (approved):

PR #873, opened 2025-06-08
Add pending reboot reason

Gijsreyn:
# PR Summary

Add a reason field property to indicate why there is a pending reboot. This is for #857


reboot_pending/reboot_pending.resource.ps1
```diff
@@ -1,13 +1,24 @@
...
-   $util = [wmiclass]"\\.\root\ccm\clientsdk:CCM_ClientUtilities"
-   $status = $util.DetermineIfRebootPending()
-   if(($status -ne $null) -and $status.RebootPending){
-     return @{ rebootPending = $true } | ConvertTo-Json -Compress
-   }
- }catch{}
- 
-return @{ rebootPending = $false } | ConvertTo-Json -Compress
\ No newline at end of file
+# Reg keys are documented here: https://learn.microsoft.com/en-us/mem/configmgr/core/servers/deploy/install/list-of-prerequisite-checks#pending-system-restart
```
SteveL-MSFT, line 4: Since you're already modifying this file, can you put the copyright header at the top?

reboot_pending/reboot_pending.resource.ps1
```diff
@@ -1,13 +1,24 @@
...
+  if(($null -ne $status) -and $status.RebootPending){
+    $reasons += "SCCM Client"
+  }
+}catch{}
+
+$result = @{
+  rebootPending = $reasons.Count -gt 0
+  reason = if ($reasons.Count -gt 0) { $reasons } else { $null }
+}
+return $result | ConvertTo-Json -Compress
```
SteveL-MSFT: Add newline at the end

reboot_pending/reboot_pending.resource.ps1
```diff
@@ -1,13 +1,24 @@
...
-   $status = $util.DetermineIfRebootPending()
-   if(($status -ne $null) -and $status.RebootPending){
-     return @{ rebootPending = $true } | ConvertTo-Json -Compress
-   }
- }catch{}
- 
-return @{ rebootPending = $false } | ConvertTo-Json -Compress
\ No newline at end of file
+# Reg keys are documented here: https://learn.microsoft.com/en-us/mem/configmgr/core/servers/deploy/install/list-of-prerequisite-checks#pending-system-restart
+$reasons = @()
```
SteveL-MSFT: It's minor, but better to use a generic list than an array

SteveL-MSFT (changes requested):
This is a great change that seems obvious now :)

reboot_pending/reboot_pending.resource.ps1
Gijsreyn: I thought the array was improved in PowerShell 7, that's why I used +=. Anyway, it is adjusted.

Gijsreyn:
Remarks resolved. Sometimes you've to see the obvious first.

reboot_pending/reboot_pending.resource.ps1
```diff
@@ -1,13 +1,28 @@
...
+    $reasons.Add("SCCM Client")
+  }
+}catch{}
+
+$result = @{
+  rebootPending = $reasons.Count -gt 0
+  reason = if ($reasons.Count -gt 0) { $reasons } else { $null }
+}
+
+return $result | ConvertTo-Json -Compress
```
SteveL-MSFT:
```suggestion
return $result | ConvertTo-Json -Compress

```

reboot_pending/tests/reboot_pending.tests.ps1
```diff
@@ -14,4 +14,21 @@ Describe 'reboot_pending resource tests' {
         $LASTEXITCODE | Should -Be 0
         $out.results.result.actualState.rebootPending | Should -Not -BeNullOrEmpty
     }
+
+    It 'reboot_pending should have a reason' -Skip:(!$IsWindows) {
+        BeforeAll {
```
SteveL-MSFT: Is using a `BeforeAll` within an `It` the recommended way over `try..finally`?

Gijsreyn: If it were a single test, I would say that, based on its purpose, it should be added fully at the top. But we have multiple, that's why it's best practice to scope it for the particular test.

SteveL-MSFT: I think you misunderstand my question.  Per https://pester.dev/docs/commands/beforeall/, the use of `BeforeAll` is intended for within a `Context` or `Describe` where you would have multiple tests (hence the `All`).  SInce thi s is a single test, it seems better to use the existing pattern of `try..finally`.

SteveL-MSFT (changes requested):

reboot_pending/tests/reboot_pending.tests.ps1
Gijsreyn: Keeping it sharp Steve! Thanks for the link. I have adjusted the code.

reboot_pending/tests/reboot_pending.tests.ps1
```diff
@@ -14,4 +14,20 @@ Describe 'reboot_pending resource tests' {
...
+        try {
+            if (-not (Get-ItemProperty "$keyPath\$keyName" -ErrorAction SilentlyContinue)) {
+                New-ItemProperty -Path $keyPath -Name $keyName -Value 1 -PropertyType DWord -Force | Out-Null
+            }
+
+            $out | dsc resource get -r Microsoft.Windows/RebootPending | ConvertFrom-Json
+            $LASTEXITCODE | Should -Be 0
+            $out.actualState.reason | Should -Not -BeNullOrEmpty
+        } finally {
+            Remove-ItemProperty -Path $keyPath -Name $keyName -ErrorAction SilentlyContinue
```
SteveL-MSFT:
```suggestion
            Remove-ItemProperty -Path $keyPath -Name $keyName -ErrorAction Ignore
```

SteveL-MSFT (approved):

Issue #853, opened 2025-05-29
Class-based resources with Export does not show capability


ThomasNieto:
### Prerequisites

- [x] Write a descriptive title.
- [x] Make sure you are able to repro it on the latest version
- [x] Search the existing issues.

### Summary

The PowerShell adapters should inspect class-based resources AST for a static `Export()` method and set the capability automatically.

### Steps to reproduce

* Create a class-based resource with `Export()` method.
* Run `dsc resource list --adapter Microsoft.Dsc/PowerShell`

### Expected behavior

```console
Export cap is listed
```

### Actual behavior

```console
Export cap is NOT listed
```

### Error details

```console

```

### Environment data

```PowerShell
7.6-preview6
```

### Version

3.1-preview6

SteveL-MSFT:
The current code requires the PS Module to advertise its DSC capabilities within the module manifest under `PrivateData.PSData.DscCapabilities`

We should probably add code such that if that field is empty or doesn't exist, then we perform analysis of the PSClass to see what capabilities it actually implements.

SteveL-MSFT:
@Gijsreyn is this something you'd want to take on?

Gijsreyn:
@SteveL-MSFT, I just submitted two separate PRs. Please take a look if you have some time.

PR #876, opened 2025-06-10
Fix discovery of capabilities PSAdapter

Gijsreyn:
# PR Summary

This PR addresses #853. The reason for the additional function is mainly due to the expected methods that the engine can expect.


powershell-adapter/Tests/powershellgroup.resource.tests.ps1
```diff
@@ -28,6 +28,7 @@ Describe 'PowerShell adapter resource tests' {
         $LASTEXITCODE | Should -Be 0
         $resources = $r | ConvertFrom-Json
         ($resources | ? { $_.Type -eq 'TestClassResource/TestClassResource' }).Count | Should -Be 1
+        ($resources | Where-Object -Property type -EQ 'TestClassResource/TestClassResource').capabilities | Should -BeIn @('get', 'set', 'test', 'export')
```
SteveL-MSFT, line 31: This test should be more specific.  Looking at the psd1, it should only be exactly `get` and `test`.  Maybe it makes sense to have another test resource that has `set` and `export`

powershell-adapter/psDscAdapter/psDscAdapter.psm1
```diff
@@ -529,6 +530,28 @@ function GetTypeInstanceFromModule {
     return $instance
 }
 
+function GetClassBasedCapabilities ($functionMemberAst) {
+    $capabilities = @()
```
SteveL-MSFT: Might as well use a list here

powershell-adapter/psDscAdapter/psDscAdapter.psm1
```diff
@@ -578,4 +601,5 @@ class DscResourceInfo {
     [string] $ImplementedAs
     [string] $CompanyName
     [System.Collections.Generic.List[DscResourcePropertyInfo]] $Properties
+    [string[]] $Methods
```
SteveL-MSFT: do we want to call this `Capabilities` instead?

SteveL-MSFT (changes requested):

powershell-adapter/Tests/powershellgroup.resource.tests.ps1
Gijsreyn: This is exactly why this change was made. If it cannot find it from the class, it looks at the .psd1 file. If the .psd1 file doesn't contain the DscCapabilities, it defaults back to get, set, and test.

SteveL-MSFT: I see.  Perhaps you can add a comment in the code indicating this is the behavior.

SteveL-MSFT (approved):

Issue #878, opened 2025-06-10
Class-based DSC resource properties alignment


Gijsreyn:
### Prerequisites

- [x] Write a descriptive title.
- [x] Make sure you are able to repro it on the latest version
- [x] Search the existing issues.

### Summary

When using the WinPS adapter, class-based properties don't align with what is happening in the PS adapter. This is because the WinPS adapter uses the `Get-DscResource` command from `PSDesiredStateConfiguration`, whereas the PS adapter leverages AST. See screenshot:

![Image](https://github.com/user-attachments/assets/67a1c1dc-6bf0-4525-a9c0-deb093b111d6)

### Steps to reproduce

1. Add the `dsc.exe` to the PATH variable
2. Run `dsc resource get --resource TestClassResource/TestClassResource --input (@{Name = 'test'} | ConvertTo-Json)`
3. Add the tests folder to PSModulePath
4. Run `dsc resource list --adapter Microsoft.Windows/WindowsPowerShell`
5. Run `dsc resource get --resource TestClassResource/TestClassResource --input (@{Name = 'test'} | ConvertTo-Json)` again

See the output:

![Image](https://github.com/user-attachments/assets/0dfc80d5-1b5a-4720-948f-bbadac7e309a)

### Expected behavior

```console
The properties should be the same
```

### Actual behavior

```console
The properties aren't the same
```

### Error details

```console

```

### Environment data

```PowerShell
Name                           Value
----                           -----
PSVersion                      7.5.1
PSEdition                      Core
GitCommitId                    7.5.1
OS                             Microsoft Windows 10.0.22631
Platform                       Win32NT
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0…}
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
WSManStackVersion              3.0
```

### Version

dsc 3.2.0-preview.1

michaeltlombardi:
The missing properties are automatically inserted to support PSDSC configuration behaviors. While the `DependsOn` property doesn't have a use in DSC, the `PsDscRunAsCredential` might be required in [Windows PowerShell / PSDSC v1.1](https://learn.microsoft.com/powershell/dsc/configurations/runasuser?view=dsc-1.1&preserve-view=true) to invoke the resource outside of the system context.

michaeltlombardi:
[...]

PR #877, opened 2025-06-10
Fix discovery of capabilities WinPSAdapter

Gijsreyn:
# PR Summary

This PR addresses issue #853. The WinPS adapter didn't have the logic that PS adapter had. Partially grabbed logic from it to determine the capabilities

PR #879, opened 2025-06-10
Fix property discovery on class-based resources WinPS adapter

Gijsreyn:
# PR Summary

This PR addresses issue #878. Note: there are no tests and might conflict with PR #877. Depending on these, I have to resolve the conflicts and add relevant tests.


Gijsreyn:
@SteveL-MSFT, could you please let me know which of the other PRs goes first? Then I can fix the merge conflict and rewrite some of the logic :)

powershell-adapter/psDscAdapter/win_psDscAdapter.psm1
```diff
@@ -239,10 +243,10 @@ function Invoke-DscCacheRefresh {
             }
 
             $dscResourceCacheEntries.Add([dscResourceCacheEntry]@{
-                Type            = "$moduleName/$($dscResource.Name)"
-                DscResourceInfo = $DscResourceInfo
-                LastWriteTimes  = $lastWriteTimes
-            })
+                    Type            = "$moduleName/$($dscResource.Name)"
```
SteveL-MSFT: This indentation doesn't look right?

Gijsreyn:
> Can you add a test?

Ready to go :)

powershell-adapter/psDscAdapter/win_psDscAdapter.psm1
SteveL-MSFT: Isn't this double indented?

powershell-adapter/psDscAdapter/win_psDscAdapter.psm1
```diff
@@ -584,6 +588,72 @@ function ValidateMethod {
...
+
+        [Parameter(Mandatory = $true)]
+        [string] $className
+    )
+
+    if (".psd1" -notcontains ([System.IO.Path]::GetExtension($filePath))) {
+        return @('get', 'set', 'test')
+    }
+
+    $module = $filePath.Replace('.psd1', '.psm1')
```
SteveL-MSFT: Although probably true in most cases, it seems we should be using `RootModule` in the psd1 instead of assuming the psm1 has same name

SteveL-MSFT (changes requested):

powershell-adapter/psDscAdapter/win_psDscAdapter.psm1
SteveL-MSFT: still there

SteveL-MSFT (changes requested):

Gijsreyn:
Somehow, the OTBS and auto-save do that double indentation. It should be good now.

powershell-adapter/psDscAdapter/win_psDscAdapter.psm1
```diff
@@ -584,6 +565,72 @@ function ValidateMethod {
...
+    $module = Get-ChildItem -Path (Split-Path $filePath -Parent) -Filter *.psm1 -File | Select-Object -First 1 -ExpandProperty FullName
+
+    $properties = [System.Collections.Generic.List[DscResourcePropertyInfo]]::new()
+
+    if (Test-Path $module -ErrorAction Ignore) {
+        [System.Management.Automation.Language.Token[]] $tokens = $null
+        [System.Management.Automation.Language.ParseError[]] $errors = $null
+        $ast = [System.Management.Automation.Language.Parser]::ParseFile($module, [ref]$tokens, [ref]$errors)
+        foreach ($e in $errors) {
+            $e | Out-String | Write-DscTrace -Operation Error
```
SteveL-MSFT: If there are any errors, should we just exit with an exit code?

powershell-adapter/psDscAdapter/win_psDscAdapter.psm1
```diff
@@ -584,6 +565,72 @@ function ValidateMethod {
...
+
+        [Parameter(Mandatory = $true)]
+        [string] $className
+    )
+
+    if (".psd1" -notcontains ([System.IO.Path]::GetExtension($filePath))) {
+        return @('get', 'set', 'test')
+    }
+
+    $module = Get-ChildItem -Path (Split-Path $filePath -Parent) -Filter *.psm1 -File | Select-Object -First 1 -ExpandProperty FullName
```
SteveL-MSFT: Hypotheically, a module author could have a different psm1 for WinPS vs PS7.  Would it work if we import the module and then use the ModuleInfo to find the relevant psm1 files?

SteveL-MSFT (changes requested):

SteveL-MSFT (approved):

PR #884, opened 2025-06-12
Surface _kind, _securityContext, and _name from resource during export

### LINKED ISSUE #598: Export should allow resource to supply name to generated instance

Currently, when `dsc` calls `export`, for each returned instance, it generates the `name` property as an incrementing value prefixed by the resource type.  This means the end user needs to manually modify each instance if they want it a bit more descriptive.

Resources should be able to return a `_name` type metadata as part of export that would be used instead.

tgauth:
# PR Summary
- update `export` to surface specific properties from resource at top-level, while also removing those properties from the lower-level object
- add corresponding pester test

## PR Context
- fix #598


dsc_lib/src/configure/config_doc.rs
```diff
@@ -134,9 +134,13 @@ pub struct Resource {
     #[schemars(regex(pattern = r"^\[resourceId\(\s*'[a-zA-Z0-9\.]+/[a-zA-Z0-9]+'\s*,\s*'[a-zA-Z0-9 ]+'\s*\)]$"))]
     pub depends_on: Option<Vec<String>>,
     #[serde(skip_serializing_if = "Option::is_none")]
+    pub kind: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
     pub properties: Option<Map<String, Value>>,
     #[serde(skip_serializing_if = "Option::is_none")]
     pub metadata: Option<Map<String, Value>>,
+    #[serde(rename = "securityContext", skip_serializing_if = "Option::is_none")]
+    pub security_context: Option<String>,
```
SteveL-MSFT: `securityContext` is part of metadata already

dsc_lib/src/configure/mod.rs
```diff
@@ -67,9 +67,21 @@ pub fn add_resource_export_results_to_configuration(resource: &DscResource, conf
     } else {
         for (i, instance) in export_result.actual_state.iter().enumerate() {
             let mut r = config_doc::Resource::new();
+            let mut props: Map<String, Value> = serde_json::from_value(instance.clone())?;
+            if let Some(kind) = props.remove("_kind") {
+                r.kind = kind.as_str().map(std::string::ToString::to_string);
```
SteveL-MSFT, line 76: This will fail if a resource incorrectly returns `_kind` as something other than a string

tools/dsctest/src/args.rs
```diff
@@ -10,6 +10,7 @@ pub enum Schemas {
     ExitCode,
     InDesiredState,
     Export,
```
SteveL-MSFT: You could just modify the `Export` test resource instead of adding a new one

SteveL-MSFT (changes requested):

dsc_lib/src/configure/config_doc.rs
tgauth: Is the goal to surface `securityContext` in the resource metadata or the resource's Microsoft.DSC metadata (or somewhere else entirely)?

SteveL-MSFT: It should be within the `Microsoft.DSC` metadata where we have `securityContext` already

ThomasNieto:
What is `_kind` and are there any restrictions on the values?

SteveL-MSFT:
> What is `_kind` and are there any restrictions on the values?

The intent is to align with ARM where we have scenarios where this info is needed https://learn.microsoft.com/en-us/azure/azure-resource-manager/templates/syntax#resources.  So short answer, is it's currently just a string with no validation.  It isn't information that will be used by DSC itself, but higher level tools so scenario specific.   I expect most DSC resources to not use this.

PR #886, opened 2025-06-15
Fix Getting Started doc adaptedResources filter

ryanspletzer:
# PR Summary

Fix a Where-Object filter in the getting started doc for showing adaptedResources.

Apologies for the trailing whitespace removal, too, if that's an issue I'll remove those changes.

## PR Context

Was perusing the docs and tried this out locally and noticed the problem.


ryanspletzer:
Looks like there's a more shorthand way to do this now, pushed another commit.

PR #903, opened 2025-06-23
DSC module example

anwather:
# PR Summary

Add a simple example to use an existing DSC resource not contained in the PSDesiredStateConfiguration module

## PR Context

I'm having some trouble getting started with DSC v3 - this will be helpful for others who have used DSC v2


Gijsreyn:
@anwather does the reference documentation on this [link](https://github.com/PowerShell/DSC/blob/main/docs/reference/resources/Microsoft/Windows/WindowsPowerShell/examples/manage-a-windows-service.md) help also? Perhaps you can also add it as example there.

michaeltlombardi (approved):

PR #907, opened 2025-06-24
Add default shell support to sshdconfig

tgauth:
# PR Summary
 - update sshdconfig resource to include get & set to configure Windows registry for Win32-OpenSSH's default shell options
 

## PR Context
- https://github.com/PowerShell/Win32-OpenSSH/wiki/DefaultShell


sshdconfig/src/main.rs
```diff
@@ -1,43 +1,44 @@
...
-        }
-        Err(e) => {
-            eprintln!("Error exporting SSHD config: {e:?}");
+    let args = Args::parse();
+
+    let result = match &args.command {
+        Command::Export => invoke_export(),
+        Command::Get => invoke_get(),
+        Command::Set { input } => invoke_set(input),
+        Command::Schema { as_global } => {
```
SteveL-MSFT: I think it would be better to have this as an enum given we're likely to add more resources for sshd in the future

sshdconfig/src/args.rs
```diff
@@ -0,0 +1,38 @@
...
+    pub command: Command,
+}
+
+#[derive(Subcommand)]
+pub enum Command {
+    /// Export `sshd_config`
+    Export,
+    /// Get default shell, eventually to be used for `sshd_config` and repeatable keywords
+    Get,
+    Schema {
```
SteveL-MSFT: Would prefer to have it in the order of Get, Set, Test, Export, Schema

sshdconfig/src/error.rs
```diff
@@ -7,6 +7,8 @@ use thiserror::Error;
 pub enum SshdConfigError {
     #[error("Command: {0}")]
     CommandError(String),
+    #[error("IO: {0}")]
```
SteveL-MSFT: Should probably do the i18n work now before it becomes more of a chore later

sshdconfig/src/get.rs
```diff
@@ -0,0 +1,85 @@
...
+
+use crate::error::SshdConfigError;
+
+/// Invoke the get command.
+///
+/// # Errors
+///
+/// This function will return an error if the desired settings cannot be retrieved.
+pub fn invoke_get() -> Result<(), SshdConfigError> {
+    // TODO: distinguish between get commands for default shell, repeatable keywords, and sshd_config
```
SteveL-MSFT: Probably make sense for `get` to take an enum and then use a `match` to determine which helper function to call

sshdconfig/src/get.rs
```diff
@@ -0,0 +1,85 @@
...
+#[cfg(windows)]
+fn get_default_shell() -> Result<(), SshdConfigError> {
+    let registry_helper = RegistryHelper::new("HKLM\\SOFTWARE\\OpenSSH", Some("DefaultShell".to_string()), None)?;
+    let default_shell: Registry = registry_helper.get()?;
+    let mut shell = None;
+    let mut shell_arguments = None;
+    if let Some(value) = default_shell.value_data {
+        match value {
+            RegistryValueData::String(s) => {
+                let parts: Vec<&str> = s.split_whitespace().collect();
```
SteveL-MSFT: I think it would help future engineers to have a comment indicating that the DefaultShell value is a command-line with the exe and args as a single string.

sshdconfig/src/get.rs
```diff
@@ -0,0 +1,85 @@
...
+                if parts.is_empty() {
+                    return Err(SshdConfigError::InvalidInput("DefaultShell cannot be empty".to_string()));
+                }
+                shell = Some(parts[0].to_string());
+                if parts.len() > 1 {
+                    shell_arguments = Some(parts[1..].iter().map(|&s| s.to_string()).collect());
```
SteveL-MSFT: In the case of args that contain whitespace, won't this not work correctly?

sshdconfig/src/get.rs
```diff
@@ -0,0 +1,85 @@
...
+
+    let registry_helper = RegistryHelper::new("HKLM\\SOFTWARE\\OpenSSH", Some("DefaultShellEscapeArguments".to_string()), None)?;
+    let escape_args: Registry = registry_helper.get()?;
+    let mut escape_arguments = None;
+    if let Some(value) = escape_args.value_data {
+        if let RegistryValueData::DWord(b) = value {
+            if b == 0 || b == 1 {
+                escape_arguments = if b == 1 { Some(true) } else { Some(false) };
+            } else {
+                return Err(SshdConfigError::InvalidInput("DefaultShellEscapeArguments must be a boolean".to_string()));
```
SteveL-MSFT: `boolean` seems technically incorrect here, shouldn't this say `0 or 1`?

sshdconfig/src/get.rs
```diff
@@ -0,0 +1,85 @@
...
+            }
+        } else {
+            return Err(SshdConfigError::InvalidInput("DefaultShellEscapeArguments must be a boolean".to_string()));
```
SteveL-MSFT: Here should say must be a DWord

sshdconfig/src/get.rs
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    let result = DefaultShell {
+        shell,
+        cmd_option,
+        escape_arguments,
+        shell_arguments
+    };
+
+    let output = serde_json::to_string_pretty(&result)?;
```
SteveL-MSFT: We should probably output as compressed and not pretty

sshdconfig/src/get.rs
```diff
@@ -0,0 +1,85 @@
...
+    println!("{output}");
+    Ok(())
+}
+
+#[cfg(not(windows))]
+pub fn get_default_shell() -> Result<(), SshdConfigError> {
+    Err(SshdConfigError::InvalidInput("Windows registry operations not applicable to this platform".to_string()))
```
SteveL-MSFT: For the user error, it should be "`DefaultShell` is only applicable to Windows"

sshdconfig/src/main.rs
```diff
@@ -1,43 +1,44 @@
...
+            let schema = if *as_global {
+                schema_for!(DefaultShell)
+            } else {
+                schema_for!(SshdConfigParser)
+            };
+            println!("{}", serde_json::to_string_pretty(&schema).unwrap());
```
SteveL-MSFT: Same here, shouldn't be pretty

sshdconfig/src/set.rs
```diff
@@ -0,0 +1,89 @@
...
+    } else {
+        remove_registry("DefaultShellEscapeArguments")?;
+    }
+
+    Ok(())
+}
+
+#[cfg(not(windows))]
+pub fn set_default_shell(_shell: Option<String>, _cmd_option: Option<String>, _escape_arguments: Option<bool>, _shell_arguments: Option<Vec<String>>) -> Result<(), SshdConfigError> {
+    Err(SshdConfigError::InvalidInput("Windows registry operations not applicable to this platform".to_string()))
```
SteveL-MSFT: Message should be: ``Setting `DefaultShell` is only applicable to Windows``

sshdconfig/src/set.rs
```diff
@@ -0,0 +1,89 @@
...
+        if shell_path.is_relative() && shell_path.components().any(|c| c == std::path::Component::ParentDir) {
+            return Err(SshdConfigError::InvalidInput("shell path must not be relative".to_string()));
+        }
+        if !shell_path.exists() {
+            return Err(SshdConfigError::InvalidInput(format!("shell path does not exist: {shell}")));
+        }
+
+        let mut shell_data = shell.clone();
+        if let Some(shell_args) = shell_arguments {
+            let args_str = shell_args.join(" ");
```
SteveL-MSFT: shell_args content might need to be enclosed in quotes if it has whitespace or have quote escaped.  See https://github.com/dotnet/runtime/blob/ee293c309a9a51f310f29e368e74edb23170ba70/src/libraries/System.Private.CoreLib/src/System/PasteArguments.cs#L10 for an example how to do this properly

sshdconfig/src/set.rs
```diff
@@ -0,0 +1,89 @@
...
+            shell_data = format!("{shell} {args_str}");
+        }
+
+        set_registry("DefaultShell", RegistryValueData::String(shell_data))?;
```
SteveL-MSFT: `DefaultShell` should be a const defined somewhere, same with the other reg value names

SteveL-MSFT (changes requested):

sshdconfig/sshd.dsc.resource.json
```diff
@@ -0,0 +1,34 @@
+{
+    "$schema": "https://aka.ms/dsc/schemas/v3/bundled/resource/manifest.json",
+    "type": "Microsoft.Windows.OpenSSH/SSHD",
```
SteveL-MSFT: I think we want to rename this to: `Microsoft.OpenSSH.SSHD/Windows` since it only applies to Windows and want to keep the `Microsoft.OpenSSH.` namespacing for intellisense consistency.

I'm open to some better names suggestions.  cc @michaeltlombardi, @mgreenegit

michaeltlombardi: I think `Microsoft.OpenSSH.SSHD/Windows` is okay as long as this resource will encapsulate every Windows-specific setting for SSHD and no other settings.

SteveL-MSFT: > I think `Microsoft.OpenSSH.SSHD/Windows` is okay as long as this resource will encapsulate every Windows-specific setting for SSHD and no other settings.

Yes, the intent here is to have it only cover Windows specific settings which currently is just `DefaultShell` (which has sub-arguments), but it leaves it open for any future ones we might add.

sshdconfig/src/set.rs
tgauth: Thinking about this more, DefaultShell does not take arguments, so this will be simpler (I mixed this up with 
setting `$env:SSH_ASKPASS` which does take arguments)

sshdconfig/src/args.rs
```diff
@@ -36,3 +39,9 @@ pub struct DefaultShell {
     pub escape_arguments: Option<bool>,
     pub shell_arguments: Option<Vec<String>>,
 }
+
+#[derive(Clone, Debug, Eq, PartialEq, ValueEnum)]
+pub enum Resource {
```
SteveL-MSFT: Instead of using `resource` which is already overused, perhaps call this `Setting`?

SteveL-MSFT (changes requested):

sshdconfig/sshd.dsc.resource.json
```diff
@@ -0,0 +1,37 @@
+{
```
SteveL-MSFT: Should rename the resource manifest `sshd-windows.dsc.resource.json` to make it more clear

build.ps1
```diff
@@ -109,7 +111,8 @@ $filesForMacPackage = @(
     'RunCommandOnSet.dsc.resource.json',
     'runcommandonset',
     'sshdconfig',
-    'sshdconfig.dsc.resource.json'
+    'sshd.dsc.resource.json',
```
SteveL-MSFT: isn't this only for Windows?

sshdconfig/src/args.rs
```diff
@@ -0,0 +1,47 @@
...
+        // Used to inform which schema to generate
+        #[clap(short = 's', long, hide = true)]
+        setting: Setting,
+    },
+}
+
+#[derive(Clone, Debug, Deserialize, Eq, JsonSchema, PartialEq, Serialize)]
+pub struct DefaultShell {
+    pub shell: Option<String>,
+    pub cmd_option: Option<String>,
```
SteveL-MSFT, line 41: use serde rename so that the fields are camelCased instead of snake_case

sshdconfig/src/get.rs
```diff
@@ -0,0 +1,81 @@
...
+
+    let output = serde_json::to_string(&result)?;
+    println!("{output}");
+    Ok(())
+}
+
+#[cfg(not(windows))]
+fn get_default_shell() -> Result<(), SshdConfigError> {
+    Err(SshdConfigError::InvalidInput(t!("get.windowsOnly").to_string()))
+}
```
SteveL-MSFT: add newline at end of file

SteveL-MSFT (approved):

Issue #685, opened 2025-03-11
An abstraction to handle using secrets


SteveL-MSFT:
### Summary of the new feature / enhancement

Current design requires passing secrets into a configuration as parameters as either `SecureString` or `SecureObject`.  However, it may be easier/better to have a way for a configuration to have sufficient information to pull secrets from a known secret store.  For example, at runtime having the secrets pulled from Azure KeyVault.

### Proposed technical implementation details (optional)

This will likely be used for enterprise which have a shared or common secret store while community configurations that require secrets may still make sense to be passed in as parameters.

What may make sense is to introduce a `getSecret('storeName', 'secretName')` function.  The `storeName` is associated with a specific DSC extension with that name and the `secretName` is passed to that extension.  The secret itself is retrieved from the extension as clear text, but treated within DSC as a `SecureString` or `SecureObject` (both should work and up to what the extension returns).

SteveL-MSFT:
Now that we have an extension mechanism, I think the way this will work is that an extension will implement the `secret` operation.  At this time, I don't think we support storing secrets, only retrieving them, so the function will simply be `secret(<nameOfSecret>)`.  `dsc` will call all extensions that implement the `secret` operation with the name of the secret as an argument:

- if one returns a value, that value is stored as a `SecureString` (I don't think we need to worry about `SecureObject` to simplify the contract)
- if none return a value, that is an error
- if more than one returns a value, that is an error

Extensions that don't contain that secret should still return a 0 success code.  A non-zero exit code should only indicate a problem retrieving the secret other than `not found`.

This model does not define how the user authenticates with the secret store, expectation is that the extension returns an appropriate error message informing the user how to unlock or login as needed.

SteveL-MSFT:
I think the way this will work is:

1. function `secret(<nameOfSecret>, [vaultName], [secureString|secureObject])`
2. only the first param is required which is the name of the secret, if no vault is specified, then all vaults are queried for the secret
    1. if more than one return a secret, it is an error
    2. if none return the secret, it is an error
    3. if the secret is an empty string, this is allowed
4. the `[vaultName]` must match the vault extension type name such as: `Microsoft.Azure/KeyVault`
5. vaults may return either a `secureString` or `secureObject` identified by the last optional parameter    1. if not specified, then DSC will try to deserialize the secret into a JSON object, if successful, it's a `secureObject` (and thus you can use dot notation to access members), if deserialization fails, it's a `secureString`
[...]

Issue #867, opened 2025-06-06
DSC should follow the parameters file schema from Azure


Bpoe:
### Summary of the new feature / enhancement

DSC uses the Azure Template schema for its configuration document. It should also follow the Azure schema for its parameters file:
'''
{
  "$schema": "https://schema.management.azure.com/schemas/2019-04-01/deploymentParameters.json#",
  "contentVersion": "1.0.0.0",
  "parameters": {
    "<first-parameter-name>": {
      "value": "<first-value>"
    },
    "<second-parameter-name>": {
      "value": "<second-value>"
    }
  }
}
'''

https://learn.microsoft.com/en-us/azure/azure-resource-manager/templates/parameter-files

Bpoe:
Also, please passing parameters from the PowerShell cmdlet:
https://learn.microsoft.com/en-us/azure/azure-resource-manager/templates/deploy-powershell#pass-parameter-values

When passing parameters via command line, parameters are simple kvp

SteveL-MSFT:
The second item should be a separate issue.  For the format of the file, I think we can check the `$schema` and use it against the ARM format otherwise expect the current format

SteveL-MSFT:
As part of this work, add support for `reference` in parameters file which would leverage `secret()`

Issue #673, opened 2025-03-06
Support encrypted secrets


SteveL-MSFT:
### Summary of the new feature / enhancement

There are scenarios where a resource might export secrets.  Using `SecureString` and `SecureObject` currently is insufficient as the contents are in clear text and the only contract is that the contents are redacted when traced.

### Proposed technical implementation details (optional)

We can adopt CMS (Cryptographic Message Syntax) using public key encryption.  This requires the user to provide a public key used for encryption.  In this scenario, a resource that exports secrets would export it as a SecureString or SecureObject and when DSC receives it, would encrypt it using the given public key.  When this config is applied, the user needs to provide the private key to decrypt the secrets which DSC would then pass to the appropriate resources.  In this case, resources don't handle encryption/decryption directly and only DSC itself handles it.

michaeltlombardi:
I think we should also document/encourage resource authors _not_ to return secrets from resources - I've been documenting this as a primary reason to define the `writeOnly` keyword for secret properties as `true` from a resource API perspective.

I can't think of a good reason to return the actual value for a secret passed to the resource except _maybe_ for resources that focus on managing secrets themselves.

Definitely agree that encrypting any values marked as secret is the correct way forward, regardless of guidance on resource authoring.

SteveL-MSFT:
@michaeltlombardi we might only say we support this for export

PR #908, opened 2025-06-25
Add support for `secret()` function and secret extensions

SteveL-MSFT:
# PR Summary

This PR adds support for `secret extensions` that get used in the config via the `secret(<name>, [vault])` function.
The `secret()` function has the first parameter as the name of the secret and the second optional parameter as the vault name.
- if the vault is not specified, then an empty string is passed to extensions and extensions are expected to globally search for that secret
- secret extensions should ALWAYS return a result, but the result JSON can have `secret` be `null` indicating that secret was not found
- secret extensions should ONLY return an error (non-zero exit code) for other reasons (like authn need to happen, network connectivity issues, etc...) with a user actionable error message
- if more than 1 secret is returned, that is considered an error
- if no secret is returned, that is considered an error

One of the challenges is plumbing the layers so that the discovered extensions get bubbled up so that the `secret()` function can call extensions to do their work.  Basically: command_discovery -> discovery -> configuration_context

Extension manifest extended for new `secret` operation with it's own custom args for passing the secret name and vault name.
Some fixes to i18n strings.

## PR Context

Fix https://github.com/PowerShell/DSC/issues/685


dsc_lib/src/functions/secret.rs
```diff
@@ -0,0 +1,85 @@
...
+                },
+                Err(err) => {
+                    warn!("{}", t!("functions.secret.extensionReturnedError", extension = extension.type_name.clone(), error = err));
+                }
+            }
+        }
+        if secret_returned {
+            Ok(Value::String(result))
+        } else {
+            Err(DscError::Function("secret".to_string(), t!("functions.secret.secretNotFound", name = secret_name).to_string()))
```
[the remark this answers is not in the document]
SteveL-MSFT: Extensions are expected to inform user of issues that prevent the extension from working correctly

ThomasNieto:
If I'm reading the spec right, DSC will look through all secret extensions in order to find the secret? If I have a few secret extensions can I specify which extension it should be searching? Like in DSC resources we specify the type name.

SteveL-MSFT:
> If I'm reading the spec right, DSC will look through all secret extensions in order to find the secret? If I have a few secret extensions can I specify which extension it should be searching? Like in DSC resources we specify the type name.

The intent is to follow `SecretManagement` module where the end user with a shared configuration can choose whatever extension they want to retrieve the secret so currently not supporting specifying a specific extension.  If you have a good use case, I can certainly add it as a 3rd optional parameter.

ThomasNieto:
The use case I'm thinking of is during vault migration. Lets say I have an existing vault extension using one technology and migrating to a new vault extension. The vaults have the same name in both extensions. A particular secret could exist in the original or new or both depending on the stage. The secret would also have the same name. Specifying the extension would be helpful to differentiate which vault should be used and also reduce time during secret discovery.

SteveL-MSFT:
> The use case I'm thinking of is during vault migration. Lets say I have an existing vault extension using one technology and migrating to a new vault extension. The vaults have the same name in both extensions. A particular secret could exist in the original or new or both depending on the stage. The secret would also have the same name. Specifying the extension would be helpful to differentiate which vault should be used and also reduce time during secret discovery.

Can you open an issue for vault migration?  Or perhaps add your scenario to https://github.com/PowerShell/DSC/issues/673 which is about exporting secrets.  In my mind, the capability you're asking for might be better suited for vault DSC resources that support export and set.  However, we would need to standardize the schema so it works across vaults.

ThomasNieto:
The use case I describe really isn't about having DSC handle vault migration but when in the process of migrating from one vault to another you will a period of time where you need to have multiple vault extensions present which may have duplicate secrets.

1. Device has both vault extensions
2. Secret exists in existing vault
3. Configuration pulls from existing vault
4. Secret is copied to new vault
5. Configuration is updated to use new vault
6. At some point later existing vault secret is removed

In my use case both vault extensions would be present on the device for a long time since each team and script would need to update their references from the existing vault extension to the new one on a secret by secret basis.

SteveL-MSFT:
> The use case I describe really isn't about having DSC handle vault migration but when in the process of migrating from one vault to another you will a period of time where you need to have multiple vault extensions present which may have duplicate secrets.
> 
> 1. Device has both vault extensions
> 2. Secret exists in existing vault
> 3. Configuration pulls from existing vault
> 4. Secret is copied to new vault
> 5. Configuration is updated to use new vault
> 6. At some point later existing vault secret is removed
> 
> In my use case both vault extensions would be present on the device for a long time since each team and script would need to update their references from the existing vault extension to the new one on a secret by secret basis.

In the case of transition from one vault to another, the current code allows for that.  Basically, if you simply request `secret('foo')` and let's assume `foo` exists in Extension1 and Extension2.  As long as the value of `foo` is the same, that will be fine.  It's only an error if `foo` exists in more than one extension and the value is different.

dsc/tests/dsc_functions_secret.tests.ps1
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    It 'Just a secret name' {
+        $configYaml = @'
+            $schema: https://aka.ms/dsc/schemas/v3/bundled/config/document.json
+            resources:
+            - name: Echo
+              type: Microsoft.DSC.Debug/Echo
+              properties:
+                output: "[secret('MySecret')]"
```
michaeltlombardi, line 22: Not the point of this PR, but should we prevent leaking secrets in this way? Accidentally leaking a secure string (and the output of the `secret` function is always secure, right?) to the output seems dangerous.

dsc_lib/locales/en-us.toml
```diff
@@ -174,6 +174,11 @@ resourceManifestSchemaDescription = "Defines the JSON Schema the resource manife
 discoverNoResults = "No results returned for discovery extension '%{extension}'"
 discoverNotAbsolutePath = "Resource path from extension '%{extension}' is not an absolute path: %{path}"
 extensionReturned = "Extension '%{extension}' returned line: %{line}"
+retrievingSecretFromExtension = "Retrieving secret '%{name}' from extension '%{extension}'"
+secretExtensionReturnedInvalidJson = "Extension '%{extension}' returned invalid JSON: %{error}"
+extensionReturnedSecret = "Extension '%{extension}' returned secret"
+extensionReturnedNoSecret = "Extension '%{extension}' did not return a secret"
+secretNoResults = "Extension '%{extension}' returned no output"
```
michaeltlombardi, line 183: Should these message also include the secret name?

dsc/tests/dsc_functions_secret.tests.ps1
SteveL-MSFT: `secureString` internally is just a string (as it's a JSON type), the only control DSC has is that itself it won't write it to console or traces.

dsc_lib/locales/en-us.toml
SteveL-MSFT: Sure

dsc/tests/dsc_functions_secret.tests.ps1
michaeltlombardi: I'm specifically wondering whether we want to add an issue to prevent things like this - afaik there's only two ways to get a secret value into a configuration document:

1. As a parameter.
2. With the `secret()` function.

In both cases, DSC is aware that the string value is a secret and shouldn't be emitted (consider the case where the desired state for a resource requires a credential object) - should we mask the value even in the actually-emitted JSON output?

dsc_lib/locales/en-us.toml
SteveL-MSFT: These messages are more for debugging when extensions get called, although line 181 is an `info` level (aka verbose) message.  If no secret is returned from any extension then the `secret()` function itself has an error message that contains the secret name.

SteveL-MSFT: image

dsc/tests/dsc_functions_secret.tests.ps1
SteveL-MSFT: For example, internally we could keep it as a JSON object with a single `secureString` (or `secureObject`) property, but resources that receive it could unwrap it and DSC wouldn't know.  I think it's best effort at best.

dsc_lib/src/functions/secret.rs
```diff
@@ -0,0 +1,85 @@
...
+            DscError::Function("secret".to_string(), t!("functions.secret.notString").to_string())
+        })?.to_string();
+        let vault_name: Option<String> = if args.len() > 1 {
+            args[1].as_str().map(std::string::ToString::to_string)
+        } else {
+            None
+        };
+
+        // if no vault name is provided, we query all extensions supporting the secret method
+        // to see if any of them can provide the secret.  If none can or if multiple can, we return an error.
```
tgauth:
```suggestion
        // we query all extensions supporting the secret method to see if any of them can provide the secret.
        // if none can or if multiple can, we return an error.
```

Bpoe:
I'm concerned with how this diverges from how Azure handles secrets with templates. I'm also concerned that this function forces this value to come from a secret manager. Shouldn't it just be a parameter for templates? The parameter file can pull the value from a secret manager. But I should also be able to provide the value via some other input method as well.

SteveL-MSFT:
> I'm concerned with how this diverges from how Azure handles secrets with templates. I'm also concerned that this function forces this value to come from a secret manager. Shouldn't it just be a parameter for templates? The parameter file can pull the value from a secret manager. But I should also be able to provide the value via some other input method as well.

Remember that DSC is cloud agnostic and also for on-premise scenarios, so it WILL deviate from ARM.

SteveL-MSFT:
After discussing with @Bpoe will add support for parameter file `reference` .  However, that will be part of https://github.com/PowerShell/DSC/issues/867

dsc_lib/src/functions/secret.rs
```diff
@@ -0,0 +1,93 @@
...
+            DscError::Function("secret".to_string(), t!("functions.secret.notString").to_string())
+        })?.to_string();
+        let vault_name: Option<String> = if args.len() > 1 {
+            args[1].as_str().map(std::string::ToString::to_string)
+        } else {
+            None
+        };
+
+        // we query all extensions supporting the secret method to see if any of them can provide the secret.
+        // if none can or if multiple can, we return an error.
```
tgauth:
```suggestion
        // if none can or if multiple provide different values, we return an error.
```

tgauth (approved):
LGTM

PR #910, opened 2025-06-26
Fix ClearCache example

### LINKED ISSUE #909: ClearCache reference example not working
### Prerequisites

### Summary

Running the ClearCache example in both `Microsoft.Windows/WindowsPowerShell` and `Microsoft.DSC/PowerShell` reference documentation doesn't work. 

### Steps to reproduce

Run:

```powershell
$adapterScript = dsc resource list Microsoft.DSC/PowerShell |
    ConvertFrom-Json |
    Select-Object -ExpandProperty directory |
    Join-Path

& $adapterScript -Operation ClearCache
```

### Expected behavior

```console
The cache is cleared.
```

### Actual behavior

```console
The input object cannot be bound because it did not contain the information required to bind all mandatory parameters:  ChildPath
```

### Error details

```console

```

### Environment data

```PowerShell
PSVersion                      7.5.1
GitCommitId                    7.5.1
```

### Version

3.1.0

### Visuals

_No response_

Gijsreyn:
# PR Summary

This fixes the ClearCache example in the reference documentation for both PowerShell adapters. It uses the .NET type to support cross-platform, rather than using "\" or "/". Also implemented the same on WindowsPowerShell to keep it identical.

## PR Context

Resolves #909

### DIFF
--- a/docs/reference/resources/Microsoft/DSC/PowerShell/index.md
+++ b/docs/reference/resources/Microsoft/DSC/PowerShell/index.md
@@ -86,7 +86,7 @@ set to `ClearCache`:
 $adapterScript = dsc resource list Microsoft.DSC/PowerShell |
     ConvertFrom-Json |
     Select-Object -ExpandProperty directory |
-    Join-Path
+    Join-Path -ChildPath ([System.IO.Path]::Combine("psDscAdapter", "powershell.resource.ps1"))
 
 & $adapterScript -Operation ClearCache
 ```
--- a/docs/reference/resources/Microsoft/Windows/WindowsPowerShell/index.md
+++ b/docs/reference/resources/Microsoft/Windows/WindowsPowerShell/index.md
@@ -83,7 +83,7 @@ set to `ClearCache`:
 $adapterScript = dsc resource list Microsoft.Windows/WindowsPowerShell |
     ConvertFrom-Json |
     Select-Object -ExpandProperty directory |
-    Join-Path
+    Join-Path -ChildPath ([System.IO.Path]::Combine("psDscAdapter", "powershell.resource.ps1"))
 
 & $adapterScript -Operation ClearCache
 ```


docs/reference/resources/Microsoft/DSC/PowerShell/index.md
```diff
@@ -86,7 +86,7 @@ set to `ClearCache`:
 $adapterScript = dsc resource list Microsoft.DSC/PowerShell |
     ConvertFrom-Json |
     Select-Object -ExpandProperty directory |
-    Join-Path
+    Join-Path -ChildPath ([System.IO.Path]::Combine("psDscAdapter", "powershell.resource.ps1"))
```
michaeltlombardi: Isn't this automatically handled in PowerShell?

docs/reference/resources/Microsoft/Windows/WindowsPowerShell/index.md
```diff
@@ -83,7 +83,7 @@ set to `ClearCache`:
 $adapterScript = dsc resource list Microsoft.Windows/WindowsPowerShell |
     ConvertFrom-Json |
     Select-Object -ExpandProperty directory |
-    Join-Path
+    Join-Path -ChildPath ([System.IO.Path]::Combine("psDscAdapter", "powershell.resource.ps1"))
```
michaeltlombardi: This only runs on Windows so I don't think we need the `[System.IO.Path]` usage here, either. We should just be able to do:

```suggestion
    Join-Path -ChildPath 'psDscAdapter\powershell.resource.ps1'
```

docs/reference/resources/Microsoft/DSC/PowerShell/index.md
michaeltlombardi: I think we should be able to do:

```suggestion
    Join-Path -ChildPath 'psDscAdapter' -AdditionalChildPath 'powershell.resource.ps1'
```

PR #911, opened 2025-06-26
Some doc fixes

SteveL-MSFT:
# PR Summary

- Some extraneous whitespace
- Extra trailing slash
- Change `where-object` to use the more common syntax

### DIFF
--- a/docs/reference/resources/Microsoft/Windows/WindowsPowerShell/index.md
+++ b/docs/reference/resources/Microsoft/Windows/WindowsPowerShell/index.md
@@ -141,7 +141,7 @@ You can also retrieve more information by directly reading it from the cache fil
 $cache = Get-Content -Path "$env:LOCALAPPDATA\dsc\WindowsPSAdapterCache.json" |
     ConvertFrom-Json
 
-($cache.ResourceCache | Where-Object { $_.type -eq '<moduleName>/<resourceName>' }).DscResourceInfo.Properties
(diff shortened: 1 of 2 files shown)


docs/reference/resources/Microsoft/Windows/WindowsPowerShell/index.md
```diff
@@ -141,7 +141,7 @@ You can also retrieve more information by directly reading it from the cache fil
 $cache = Get-Content -Path "$env:LOCALAPPDATA\dsc\WindowsPSAdapterCache.json" |
     ConvertFrom-Json
 
-($cache.ResourceCache | Where-Object -Property type -EQ '<moduleName>/<resourceName>').DscResourceInfo.Properties
+($cache.ResourceCache | Where-Object { $_.type -eq '<moduleName>/<resourceName>' }).DscResourceInfo.Properties
```
Gijsreyn: I think the previous line is the more simplified syntax, as we are not having multiple properties to filter on. If there are more, I would have used the FilterScript approach.

michaeltlombardi: For the documentation site, we generally write our examples with the full parameters unless we're specifically showing positional parameters for that purpose.

So if we were to make this change, it would be:

 ```suggestion
($cache.ResourceCache | Where-Object -FilterScript { $_.type -eq '<moduleName>/<resourceName>' }).DscResourceInfo.Properties
```

But I think the existing form ends up more intuitive for less-experienced readers.

michaeltlombardi (approved):

Issue #885, opened 2025-06-13
Have a PSScript resource


SteveL-MSFT:
### Summary of the new feature / enhancement

As part of helping customers transition from script to declarative and fill in the gap where a resource doesn't exist yet, we should have a Transitional resource that allows inline PSScript.

### Proposed technical implementation details (optional)

Some considerations:

- Probably allow for separate get/set/test/export
- Probably not allow just referencing a .ps1 as it hides the script even though the script block could do this
- Probably emit a warning it's not idempotent
- Have separate resources for WinPS vs PS7
- Probably detect if any output is not JSON and convert it
- Auto-convert Verbose, Info, Debug streams to equivalent traces (Debug would be Trace level as it might have sensitive info)

theJasonHelmick:
I agree to limit the .ps1 script access -- although this may be confusing to some at first to have the restriction. I think the point that this is  "transitional" for early learning and development of a resource POC before writing the resource.  The restriction then seems a benefit -- by reducing the rabbit holes.

PR #937, opened 2025-07-02
PSScript resource

SteveL-MSFT:
# PR Summary

New PS7 and WinPS5.1 script resources.  Allow inline script execution for `get`, `set`, and `test` operations.  `export` isn't supported due to how it works and is reserved for actual resources.

If an operation isn't implemented, it is ignored with an `info` level message written and empty JSON being returned.  However, if `test` is not implemented, then it always returns that it is `_inDesiredState = true`.

I did it async to preserve order of traces that could come out of the script.  Note that `Write-Information` is mapped to `trace` level while `Write-Verbose` is mapped to `info` to align with how they are defined in DSC.  `Write-Host` is mapped to an `info` message.

Separate `Microsoft.DSC.Transitional/PowerShellScript` and `Microsoft.DSC.Transitional/WindowsPowerShellScript` resources (latter is only on Windows) but they use the same underlying script as the resource.

Output is expected to not already be JSON and will be converted automatically.  `$VerbosePreference` is `continue`, `$DebugPreference` is `continue`, and `$ErrorActionPreference` is `stop`.

Input can be any valid JSON and is passed to the single parameter declared in the param block as PSCustomObject.  If the script has a param and no input is provided, that's an error.  If input is provided and the script doesn't have a param, that is also an error.  If more than one param is declared, that is also an error.

Example config using it:

```yaml
# Example configuration mixing native app resources with classic PS resources
parameters:
  myName:
    type: string
    defaultValue: Steve
  myObject:
    type: object
    defaultValue:
      color: green
      number: 10
resources:
- name: Use PS script
  type: Microsoft.DSC.Transitional/PowerShellScript
  properties:
    input:
      - name: "[parameters('myName')]"
      - object: "[parameters('myObject')]"
    getScript: |
      param($inputArray)

      Write-Warning "This is a warning message"
      # any output will be collected and returned
      "My name is " + $inputArray[0].name
      "My color is " + $inputArray[1].object.color
```

## PR Context

Fix https://github.com/PowerShell/DSC/issues/885


resources/PSScript/winpsscript.dsc.resource.json
```diff
@@ -0,0 +1,79 @@
+{
+    "$schema": "https://aka.ms/dsc/schemas/v3/bundled/resource/manifest.json",
+    "type": "Microsoft.DSC.Transitional/WindowsPowerShellScript",
+    "description": "Enable running Windows PowerShell 5.1 scripts inline",
+    "version": "0.1.0",
+    "get": {
+        "executable": "pwsh",
```
Gijsreyn: Shouldn't this be powershell?

resources/PSScript/winpsscript.dsc.resource.json
```diff
@@ -0,0 +1,79 @@
+{
+    "$schema": "https://aka.ms/dsc/schemas/v3/bundled/resource/manifest.json",
+    "type": "Microsoft.DSC.Transitional/WindowsPowerShellScript",
+    "description": "Enable running Windows PowerShell 5.1 scripts inline",
+    "version": "0.1.0",
```
Gijsreyn, line 5: Would it make sense to bump it up to 1.0.0?

SteveL-MSFT: When 3.2 hits RC, it would make sense at that time

resources/PSScript/winpsscript.dsc.resource.json
SteveL-MSFT: Oops, cut and paste error

resources/PSScript/psscript.tests.ps1
```diff
@@ -0,0 +1,308 @@
...
+          "Input: $inputObj"
+        input: "This is a string"
+'@
+        $result = dsc resource get -r $resourceType -i $yaml 2> $TestDrive/error.txt | ConvertFrom-Json
+        $LASTEXITCODE | Should -Be 0 -Because (Get-Content $TestDrive/error.txt -Raw | Out-String)
+        $result.actualState.output.Count | Should -Be 1 -Because ($result | ConvertTo-Json -Depth 10 | Out-String)
+        $result.actualState.output[0] | Should -BeExactly "Input: This is a string"
+    }
+
+    It 'Input without param block is an error for <resourecType>' -TestCases $testCases {
```
Gijsreyn: Nit: Small typo on `<resourecType>` to `<resourceType>`

resources/PSScript/winpsscript.dsc.resource.json
```diff
@@ -0,0 +1,83 @@
...
+            "$input | ./psscript.ps1",
+            "test"
+        ],
+        "input": "stdin",
+        "return": "state"
+    },
+    "exitCodes": {
+        "0": "Success",
+        "1": "PowerShell script execution failed",
+        "2": "PowerShell exception occurred",
```
Gijsreyn, line 54: Out of curiosity, were is exit code 2 defined in the code?

SteveL-MSFT: Exit code 2 comes from PowerShell itself when it exits due to unhandled exception

resources/PSScript/psscript.tests.ps1
SteveL-MSFT:
```suggestion
    It 'Input without param block is an error for <resourceType>' -TestCases $testCases {
```

SteveL-MSFT: Good catch!

resources/PSScript/psscript.ps1
```diff
@@ -0,0 +1,166 @@
...
+    $ast.ParamBlock.Parameters[0].Name.VariablePath.UserPath
+} else {
+    $null
+}
+
+$ps = [PowerShell]::Create().AddScript({
+    $DebugPreference = 'Continue'
+    $VerbosePreference = 'Continue'
+    $ErrorActionPreference = 'Stop'
+}).AddScript($script)
```
SeeminglyScience:
```suggestion
}).AddStatement().AddScript($script)
```

Default is to pipe, so currently this would be like `. { $DebugPreference = etc } | . $script`

resources/PSScript/psscript.ps1
```diff
@@ -0,0 +1,166 @@
...
+        Write-DscTrace -Now -Level Error -Message 'Input was provided but script does not have a parameter to accept input.'
+        exit 3
+    }
+    $null = $ps.AddParameter($paramName, $scriptObject.input)
+} elseif ($null -ne $paramName) {
+    Write-DscTrace -Now -Level Error -Message "Script has a parameter '$paramName' but no input was provided."
+    exit 3
+}
+
+$ps.Streams.Error.add_DataAdded({
```
SeeminglyScience: I would recommend switching to `Register-ObjectEvent`. When an event handler is added like this, the scriptblock is just wrapped in a delegate and added directly. If the event is fired in a different thread (which will be the case here) this is what happens:

1. Check if there is a default runspace for the current thread
    - No: then throw immediately
    - Yes: Continue to 2
2. Check if the scriptblock to be invoked has affinity to this thread's runspace
    - Yes: Run normally in the current thread
    - No: Continue to 3
3. Attempt to marshal the scriptblock back to it's origin thread. Only attempt this for 300ms
    - Success: origin thread was able to product a pulse (event processing) pipeline within 300ms, typically means the origin thread was not busy or was in a loop
    - Failure: origin runspace is currently blocking, unsafely access it from this thread as well with no locks

The way it's currently structured will *probably* always hit "Step 3 Success", but if a blocking call is ever added later on, this could easily end up with some nasty state corruption.

For reference: PowerShell/PowerShell#4003

resources/PSScript/psscript.ps1
```diff
@@ -0,0 +1,166 @@
...
+    param($sender, $args)
+    Write-DscTrace -Level Info -Message $sender.Message
+})
+$ps.Streams.Debug.add_DataAdded({
+    param($sender, $args)
+    Write-DscTrace -Level Debug -Message $sender.Message
+})
+$outputObjects = [System.Collections.Generic.List[Object]]::new()
+
+function write-traces() {
```
SeeminglyScience:
```suggestion
function Write-TraceList {
```

(optional nit to quiet the PowerShell pedants like myself (PascalCase names, no plurals))

resources/PSScript/psscript.ps1
```diff
@@ -0,0 +1,166 @@
...
+        if ($traceQueue.TryDequeue([ref] $trace)) {
+            $host.ui.WriteErrorLine($trace)
+        }
+    }
+}
+
+try {
+    $asyncResult = $ps.BeginInvoke()
+    while (-not $asyncResult.IsCompleted) {
+        write-traces
```
SeeminglyScience:
```suggestion
        Write-TraceList
```

(if you took the optional nit)

resources/PSScript/psscript.ps1
```diff
@@ -0,0 +1,166 @@
...
+        Start-Sleep -Milliseconds 100
+    }
+    $outputCollection = $ps.EndInvoke($asyncResult)
+    write-traces
```
SeeminglyScience:
```suggestion
    Write-TraceList
```

(if you took the optional nit)

resources/PSScript/psscript.ps1
```diff
@@ -0,0 +1,166 @@
...
+    param($sender, $args)
+    Write-DscTrace -Level Error -Message $sender.Message
+})
+$ps.Streams.Warning.add_DataAdded({
+    param($sender, $args)
+    Write-DscTrace -Level Warn -Message $sender.Message
+})
+$ps.Streams.Information.add_DataAdded({
+    param($sender, $args)
+    Write-DscTrace -Level Trace -Message $sender.MessageData.ToString()
```
SeeminglyScience:
```suggestion
    Write-DscTrace -Level Trace -Message $sender.MessageData?.ToString()
```

While rare, `InformationRecord` can be manually created with a null message data

SeeminglyScience (commented):
Lookin' good, I do have a couple suggestions

resources/PSScript/psscript.ps1
```diff
@@ -0,0 +1,176 @@
...
+} elseif ($null -ne $paramName) {
+    Write-DscTrace -Now -Level Error -Message "Script has a parameter '$paramName' but no input was provided."
+    exit 3
+}
+
+$traceQueue = [System.Collections.Concurrent.ConcurrentQueue[object]]::new()
+
+$null = Register-ObjectEvent -InputObject $ps.Streams.Error -EventName DataAdding -MessageData $traceQueue -Action {
+    $traceQueue = $Event.MessageData
+    $traceQueue.Enqueue((@{ error = $EventArgs.ItemAdded.Message } | ConvertTo-Json -Compress))
```
SeeminglyScience: `ErrorRecord` doesn't have a `Message` property so I'm guessing this isn't working correctly atm

```suggestion
    $traceQueue.Enqueue((@{ error = ($EventArgs.ItemAdded | Out-String -Width 9999) } | ConvertTo-Json -Compress))
```

or

```suggestion
    $traceQueue.Enqueue((@{ error = [string]$EventArgs.ItemAdded } | ConvertTo-Json -Compress))
```

resources/PSScript/psscript.ps1
```diff
@@ -0,0 +1,176 @@
...
+        Write-DscTrace -Now -Level Error -Message 'Errors occurred during script execution.'
+        exit 1
+    }
+
+    foreach ($output in $outputCollection) {
+        $outputObjects.Add($output)
+    }
+}
+catch {
+    Write-DscTrace -Now -Level Error -Message $_.Exception.Message
```
SeeminglyScience:
```suggestion
    Write-DscTrace -Now -Level Error -Message $_
```

Since the parameter is typed as `string` no cast is needed, and allowing `ErrorRecord` to implicitly cast will account for things like `ErrorDetails.Message` overrides

Issue #829, opened 2025-05-22
Remove UTF-8 BOM from file stream


SteveL-MSFT:
### Summary of the new feature / enhancement

When reading a file, it might have the UTF-8 BOM  (0xEF, 0xBB, 0xBF).  This currently causes deserialization to fail.

### Proposed technical implementation details (optional)

Before deserialization, we can probably check if the first 3 bytes is the BOM and remove it.

PR #944, opened 2025-07-06
Fix removing UTF-8 BOM from file

SteveL-MSFT:
# PR Summary

When reading a file, remove the UTF-8 BOM if it exists

## PR Context

Fix https://github.com/PowerShell/DSC/issues/829


tgauth (approved):
LGTM

PR #946, opened 2025-07-08
Add WMI set capability

Gijsreyn:
# PR Summary

This PR adds the `set` capability on the WMI adapter #475. Traditionally, resources should have set, but in the case of WMI resources, it requires a method name and optionally parameters. These are filtered out during a `Get` operation, but canonical during a `Set`.


SteveL-MSFT:
@Gijsreyn I appreciate your attempt here, but I think the problem is more complex.  Some WMI classes actually support `set` like `Win32_Environment` (and gets even more complex due to `Create or Update`) and others like `Win32_NetworkAdapter` has a method `SetPowerState()` to power cycle it.  I don't think it's worth trying to abstract this as it requires a knowledge base somewhere.  Instead, it may make sense to have a different resource like `Microsoft.Windows.WMI/MethodInvocation` that takes a namespace, classname, method, and parameters and have `set` just try to call the WMI/CIM `set` operation.

Gijsreyn:
Thanks @SteveL-MSFT. If you want me to hit that one up, let me know. Before doing so, let's have a quick talk on the design on the thing.

SteveL-MSFT:
@Gijsreyn let's queue this up for next WG meeting

Gijsreyn:
It was more work than expected @SteveL-MSFT, but as discussed during WG meeting, I have implemented the `Set` method. There have been some changes to the function to support both create/update. That's the reason some logic is moved to functions. 

Probably in the documentation, we can make a note that it does not apply to `CimClassMethods` and static methods. I would appreciate it if you got time to do a review :)

wmi-adapter/Tests/wmi.tests.ps1
```diff
@@ -52,10 +52,58 @@ Describe 'WMI adapter resource tests' {
...
         $res.results[1].result.actualState.result[0].properties.Name | Should -Not -BeNullOrEmpty
         $res.results[1].result.actualState.result[0].properties.BootupState | Should -BeNullOrEmpty
         $res.results[1].result.actualState.result[1].properties.Caption | Should -Not -BeNullOrEmpty
         $res.results[1].result.actualState.result[1].properties.BuildNumber | Should -BeNullOrEmpty
-        $res.results[1].result.actualState.result[4].properties.AdapterType | Should -BeLike "Ethernet*"
+        $res.results[1].result.actualState.result[4].properties.Name | Should -Not -BeNullOrEmpty
+    }
+
+    It 'Set does not work without input for resource' -Skip:(!$IsWindows) {
+        $out = dsc resource set --resource root.cimv2/Win32_Environment --input '{}' 2>&1
```
SteveL-MSFT: My preference is to redirect STDERR to a file and not mix it with STDOUT:

```powershell
$out = dsc resource set --resource root.cimv2/Win32_Environment --input '{}' 2>$TestDrive/error.log
$out | Should -BeNullOrEmpty
(Get-Content $TestDrive/error.log _Raw) | Should -BeLike "*No valid properties found in the CIM class 'Win32_Environment' for the provided properties.*"
```

wmi-adapter/Tests/wmi.tests.ps1
```diff
@@ -52,10 +52,58 @@ Describe 'WMI adapter resource tests' {
...
+    }
+
+    It 'Set does not work without a key property' -Skip:(!$IsWindows) {
+        $i = @{
+            VariableValue = "TestValue"
+            UserName = ("{0}\{1}" -f $env:USERDOMAIN, $env:USERNAME) # Read-only property is key, but we require a key property to be set
+        } | ConvertTo-Json
+        
+        $out = dsc resource set -r root.cimv2/Win32_Environment -i $i 2>&1
+        $out[0] | Should -BeLike "*All key properties in the CIM class 'Win32_Environment' are read-only, which is not supported.*"
```
SteveL-MSFT: Since `UserName` is a key, the adapter needs it to find the right instance, so I would expect this to error that not all key properties are provided?

wmi-adapter/wmi.dsc.resource.json
```diff
@@ -35,6 +35,20 @@
...
+        "-NoLogo",
+        "-NonInteractive",
+        "-NoProfile",
+        "-ExecutionPolicy",
+        "Bypass",
+        "-Command",
+        "$Input | ./wmi.resource.ps1 Set"
+      ],
+      "input": "stdin",
+      "implementsPretest": true
```
SteveL-MSFT: `implementsPretest = true` would mean that the adapter would not actually perform a set if the input matches current state, I don't see that in the code.  May be simpler to set to `false` and let DSC engine perform a get to determine if a set is necessary, although it would be more efficient if the adapter did this.  So depends on if you want to add that code.

wmi-adapter/wmiAdapter.psm1
```diff
@@ -35,6 +35,188 @@ function Get-DscResourceObject {
...
+        [switch] $SkipReadOnly,
+
+        [Parameter()]
+        [switch] $ValidateKeyProperty
+    )
+
+    $availableProperties = $CimClass.CimClassProperties | Where-Object -Property Name -in $Properties.psobject.Properties.name
+    $validatedProperties = [System.Collections.Generic.List[Array]]::new()
+
+    $keyProperty = ($availableProperties.flags -like "Property, Key*")
```
SteveL-MSFT: Better to do a check like:

```powershell
$keyProperties = $availableProperties | Where-Object {$_.Flags.Hasflag([Microsoft.Management.Infrastructure.CimFlags]::Key)}
```

instead of relying on the string representation

wmi-adapter/wmiAdapter.psm1
```diff
@@ -35,6 +35,188 @@ function Get-DscResourceObject {
...
+        exit 1
+    }
+
+    if ($ValidateKeyProperty.IsPresent) {
+        # Check if any key property is also read-only
+        $keyProps = $availableProperties | Where-Object { $_.Flags.ToString() -like "*Key*" }
+        $readOnlyKeyProps = $keyProps | Where-Object { $_.Flags.ToString() -like "*ReadOnly*" }
+
+        if ($readOnlyKeyProps.Count -eq $keyProps.Count) {
+            "All key properties in the CIM class '$ClassName' are read-only, which is not supported." | Write-DscTrace -Operation Error
```
SteveL-MSFT: I think we should just ignore read-only properties instead of erroring so that you can do a `get`, modify, `set`.  I think a warning or maybe verbose (meaning `info` in DSC terms) message would be ok.

wmi-adapter/wmiAdapter.psm1
```diff
@@ -35,6 +35,188 @@ function Get-DscResourceObject {
...
+        
+        [Parameter(Mandatory = $true)]
+        [psobject]$DesiredStateProperties,
+        
+        [Parameter()]
+        [switch]$KeyPropertiesOnly
+    )
+    
+    $targetProperties = if ($KeyPropertiesOnly.IsPresent) {
+        $Properties | Where-Object { $_.Flags.ToString() -like "*Key*" }
```
SteveL-MSFT: Same as above, use the `CimFlag` type

wmi-adapter/wmiAdapter.psm1
```diff
@@ -113,24 +260,86 @@ function GetCimSpace {
...
+                $wmi_instance = ValidateCimMethodAndArguments -DesiredState $r
+                $properties = @{}
 
+                $wmi_instance.Properties | ForEach-Object {
+                    if ($r.properties.psobject.properties.name -contains $_.Name) {
+                        $properties[$_.Name] = $r.properties.$($_.Name)
+                    }
+                }
+
+                $readOnlyProperties = $wmi_instance.Properties | Where-Object -Property Flags -like "*ReadOnly*"
```
SteveL-MSFT: use CimFlag

wmi-adapter/wmiAdapter.psm1
```diff
@@ -113,24 +260,86 @@ function GetCimSpace {
...
+                "Test operation is not implemented for WMI/CIM methods." | Write-DscTrace -Operation Error 
+                exit 1
             }
         }
     }
 
     return $result
 }
 
+function ValidateCimMethodAndArguments {
```
SteveL-MSFT: This function needs to be renamed now, right?

wmi-adapter/wmi.dsc.resource.json
Gijsreyn: I needed that refresher. I'll keep the test method out for now and plan to implement it in another commit. Turning it to `false` now.

wmi-adapter/wmiAdapter.psm1
Gijsreyn: Renamed.

wmi-adapter/wmiAdapter.psm1
Gijsreyn: Only readOnly key properties aren't working when you try to do a set:

image

The same applies to the previous comment on the adapter finding the right instance. That's why we derive it once again using the key property only (without the readOnly one)

wmi-adapter/wmiAdapter.psm1
```diff
@@ -35,6 +35,189 @@ function Get-DscResourceObject {
...
+    if ($ValidateKeyProperty.IsPresent) {
+        # Check if any key property is also read-only
+        if ($keyProperties.Count -eq 0) {
+            "No key properties found in the CIM class '$ClassName'." | Write-DscTrace -Operation Error
+            exit 1
+        }
+        $readOnlyKeyProps = $keyProperties | Where-Object { $_.Flags.HasFlag([Microsoft.Management.Infrastructure.CimFlags]::ReadOnly) }
+
+        if ($readOnlyKeyProps.Count -eq $keyProperties.Count) {
+            "All key properties in the CIM class '$ClassName' are read-only, which is not supported." | Write-DscTrace -Operation Error
```
SteveL-MSFT: I see your intent now.  I think you should change this message to be more clear:

```suggestion
            "All properties specified in the CIM class '$ClassName' are read-only, which is not supported." | Write-DscTrace -Operation Error
```

Gijsreyn:
Should be good to go @SteveL-MSFT . Hope this can be merged in before the preview release :)

SteveL-MSFT (approved):

PR #948, opened 2025-07-08
Example-linux-assertion

theJasonHelmick:


SteveL-MSFT:
@theJasonHelmick for each example config, can you add a pester test under the configurations folder so we can be sure they don’t get broken?

configurations/linux/linux_assertion.dsc.yaml
```diff
@@ -0,0 +1,12 @@
+# Assertion for Linux
+$schema: https://aka.ms/dsc/schemas/v3/bundled/config/document.json
```
Gijsreyn, line 2: I was just wondering after Mikey's PR got in on the schema, that it should be targeted at v3.1?

SteveL-MSFT:
Similar to my feedback on the Windows assertion, it would make sense to have a more complete config example that uses linux assertion

theJasonHelmick:
@Gijsreyn @SteveL-MSFT - Added depends on with the Echo resource.

theJasonHelmick:

PR #950, opened 2025-07-08
Example Windows assertion

theJasonHelmick:


Gijsreyn:
Hey @theJasonHelmick, there is also an example for Windows available at: https://github.com/PowerShell/DSC/blob/main/dsc/examples/assertion.dsc.yaml

SteveL-MSFT:
Instead of a standalone example of this assertion, I think it makes more sense to have it on top of any Windows specific examples

theJasonHelmick:
Added dependsOn for Echo resource

theJasonHelmick:

PR #951, opened 2025-07-08
Example_Windows_Darkmode

theJasonHelmick:


SteveL-MSFT:
I think light mode and dark mode should be a single config that uses parameters

theJasonHelmick:

PR #955, opened 2025-07-10
Fix trace message for function result

SteveL-MSFT:
# PR Summary

Previously the code was using `result` as the trace parameter name, but it's actually `results`.  However, it also used the debug format which although more human readable, I think JSON is easier as you can pretty print or even convert to YAML easily.


dsc_lib/src/parser/expressions.rs
```diff
@@ -113,7 +113,9 @@ impl Expression {
     /// This function will return an error if the expression fails to execute.
     pub fn invoke(&self, function_dispatcher: &FunctionDispatcher, context: &Context) -> Result<Value, DscError> {
         let result = self.function.invoke(function_dispatcher, context)?;
-        trace!("{}", t!("parser.expression.functionResult", result = result : {:?}));
+        // TODO: check if `secret()` funciton and don't emit result
```
tgauth:
```suggestion
        // TODO: check if `secret()` function and don't emit result
```

tgauth (approved):

PR #958, opened 2025-07-11
sshdconfig: add tracing and update constants

tgauth:
# PR Summary
- add json tracing to sshdconfig (based on registry implementation)
- update consts lists in metadata.rs and add related tests to tree-sitter grammar and parser


sshdconfig/src/main.rs
```diff
@@ -22,14 +25,25 @@ mod util;
...
+    enable_tracing();
+
     let args = Args::parse();
 
     let result = match &args.command {
-        Command::Export => invoke_export(),
-        Command::Get { setting } => invoke_get(setting),
-        Command::Set { input } => invoke_set(input),
+        Command::Export => {
+            debug!("Export command");
```
SteveL-MSFT: should i18n all strings

SteveL-MSFT (changes requested):
I noticed in local builds this message:

```
Building tree-sitter-ssh-server-config ... for current
Warning: rule empty_line is just a `seq` or `choice` rule with a single element. This is unnecessary.
```

Do you want to fix that in this PR as well?

Issue #50, opened 2023-04-04
Handle resources that require restarts


SteveL-MSFT:
### Summary of the new feature / enhancement

Some changes to the system may require a reboot before additional changes can be applied.  `dsc` itself shouldn't perform a reboot, but we may need a standardized way for resources to return that a reboot is required like `_rebootRequested: true` and then `dsc` just reports that as a standard result.  Then something else performs the reboot when ready and re-runs the config until it completes or hits another reboot.

ThomasNieto:
Even if a higher level agent is managing the reboots there is a need for a configuration/resource to indicate that the reboot must occur first before continuing applying the configuration.

SteveL-MSFT:
WG discussed this and current proposal is for resources that require a reboot to emit `_rebootRequired=true` as part of their output and `dsc` will inform the user that a reboot is required.  If the user ignores that and reruns the config, it will be up to the resource to know if a reboot has been satisfied or do nothing except re-emit `_rebootRequired=true`.

SteveL-MSFT:
Based on a new scenario that came up, I'm renaming this to `restart` instead of `reboot` to handle other cases like restarting a service, window manager, or even computer system.

In this case, my proposal is if a restart is required, we handle 3 cases (more can be added later, like clusters):

- If a process needs to be restarted, the resource returns a `_restartRequired` property with contents:

```yaml
_restartRequired:
  process:
    name: explorer.exe
    id: 1234
```

- if a service needs to be restarted, the resource returns:

```yaml
_restartRequired:
  service:
    name: sshd
    id: 1234
```

- if the computer system needs to be restarted:

```yaml
_restartRequired:
  computer:
    name: MyComputer
```

Resources that can handle restarting should have a property that indicates a restart should be performed if needed:

```yaml
_restartIfNeeded: true
```

The default value is `false` and returns the `_restartRequired` as above.

michaeltlombardi:
Is there a case for a single resource returning an array of processes or services that need to be restarted?

SteveL-MSFT:
> Is there a case for a single resource returning an array of processes or services that need to be restarted?

I can't think of one, but we can allow a choice of an array if that situation comes up.

SteveL-MSFT:
I think it may be better to have this returned as metadata than a property

PR #975, opened 2025-07-16
Enable resources to return _restartRequired metadata

SteveL-MSFT:
# PR Summary

Allow resources to return `_restartRequired` as part of `_metadata`.  This is an array that can contain any of:

```json
{ "system": "computerName" }
```

```json
{ "service": "serviceName" }
```

```json
{ "process":
  {
    "name": "processName",
    "id": "pid"
  }
}
```

The metadata is pulled into each resources metadata as part of result object, but also collected into the top level `Microsoft.DSC` metadata:

image

Any items in the array that don't conform to the above results in a warning.  `dsc schema -t restart-required` returns the JSONSchema.

## PR Context

Fix https://github.com/PowerShell/DSC/issues/50

### DIFF
--- a/dsc_lib/src/configure/config_doc.rs
+++ b/dsc_lib/src/configure/config_doc.rs
@@ -34,6 +34,21 @@ pub enum ExecutionKind {
     WhatIf,
 }
 
+#[derive(Debug, Clone, PartialEq, Deserialize, Serialize, JsonSchema)]
+#[serde(rename_all = "camelCase")]
+pub struct Process {
+    pub name: String,
+    pub id: u32,
+}
+
+#[derive(Debug, Clone, PartialEq, Deserialize, Serialize, JsonSchema)]
+#[serde(rename_all = "camelCase")]
+pub enum RestartRequired {
+    System(String),
+    Service(String),
+    Process(Process),
+}
+
 #[derive(Debug, Default, Clone, PartialEq, Deserialize, Serialize, JsonSchema)]
 pub struct MicrosoftDscMetadata {
     /// Version of DSC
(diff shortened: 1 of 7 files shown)


dsc_lib/src/configure/config_doc.rs
```diff
@@ -57,6 +72,9 @@ pub struct MicrosoftDscMetadata {
     /// The security context of the configuration operation, can be specified to be required
     #[serde(rename = "securityContext", skip_serializing_if = "Option::is_none")]
     pub security_context: Option<SecurityContextKind>,
+    /// Indicates what needs to be restarted after the configuration operation
+    #[serde(rename = "restartRequired", skip_serializing_if = "Option::is_none")]
+    pub restart_required: Option<Vec<RestartRequired>>,
```
tgauth, line 77: Did you consider displaying `restartRequired` as a struct that supports arrays for `System`, `Process`, and `Service` instead of repeating `Service`, for instance, in the output array, if there is multiple?

dsc_lib/src/configure/mod.rs
```diff
@@ -225,14 +225,24 @@ fn check_security_context(metadata: Option<&Metadata>) -> Result<(), DscError> {
...
+fn get_metadata_from_result(mut context: Option<&mut Context>, result: &mut Value, metadata: &mut Metadata) -> Result<(), DscError> {
     if let Some(metadata_value) = result.get("_metadata") {
         if let Some(metadata_map) = metadata_value.as_object() {
             for (key, value) in metadata_map {
                 if key.starts_with("Microsoft.DSC") {
                     warn!("{}", t!("configure.mod.metadataMicrosoftDscIgnored", key = key));
                     continue;
                 }
+                if let Some(ref mut context) = context {
+                    if key == "_restartRequired" {
```
tgauth, line 237: Any particular reason for using `_restartRequired` as the resource key but `restartRequired` in the top-level metadata, when both are displayed in the output?

dsc_lib/src/configure/config_doc.rs
SteveL-MSFT: I did consider that, but thought that an array of objects was slightly better if we add a new object type in the future vs a consumer of this having to know different properties (although they'd have to know different object types I guess).

dsc_lib/src/configure/mod.rs
SteveL-MSFT: So we want to be consistent in that properties that start with underscore are canonical meaning they have specific DSC meaning and reserved for DSC use.  However, that is only needed for properties returned or accepted by resources.  Within the `metadata`, anything under `Microsoft.DSC` is already reserved, so I think for consumers of output of config, it made sense to not have the underscore.  I do realize it's not symmetric, but it's consistent within resources and within config consumers.

Issue #976, opened 2025-07-17
Initial features needed for Bicep


SteveL-MSFT:
### Summary of the new feature / enhancement

[ ] Types
[ ] utcNow()
[ ] uniqueString()
[ ] empty()
[ ] union()
[ ] not()
[ ] or()
[ ] and()
[ ] contains()
[ ] endsWith()
[ ] flatten()?
[ ] map() - lambda
[ ] loop (aka copy)
[ ] filter() -lambda
[ ] length()
[ ] reduce() -lambda
[ ] output

Gijsreyn:
@SteveL-MSFT - I think this is also part of the ARM language v2 support. It would be pretty cool if something like the following can be done:

```bicep
extension dsc
targetScope = 'desiredStateConfiguration'

param companyName string = 'MyCompany'
param valueData registryValueDataType

resource registry 'Microsoft.Windows/Registry@1.0.0' = {
  keyPath: 'HKLM\\Software\\${toLower(companyName)}'
  valueName: 'InstallPath'
  valueData: valueData
}

// =============== //
//   Definitions   //
// =============== //

@export()
@description('The type for registry value data. Can be one of several variants based on the registry value type.')
type registryValueDataType = {
  @description('Optional. A string value (REG_SZ).')
  String: string?

  @description('Optional. An expandable string value (REG_EXPAND_SZ).')
  ExpandString: string?

  @description('Optional. Binary data (REG_BINARY).')
  Binary: int[]?

  @description('Optional. A 32-bit number (REG_DWORD).')
  DWord: int?

  @description('Optional. An array of strings (REG_MULTI_SZ).')
  MultiString: string[]?

  @description('Optional. A 64-bit number (REG_QWORD).')
  QWord: int?
}

```

PR #979, opened 2025-07-17
Add remaining logical functions: and, bool, false, true, not, or

SteveL-MSFT:
# PR Summary

Implement all the remaining logical functions: `and()`, `bool()`, `false()`, `true()`, `not()`, `or()` 

## PR Context

### DIFF
--- a/dsc_lib/src/functions/bool.rs
+++ b/dsc_lib/src/functions/bool.rs
@@ -1,0 +1,83 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+use crate::DscError;
+use crate::configure::context::Context;
+use crate::functions::{AcceptedArgKind, Function, FunctionCategory};
+use rust_i18n::t;
+use serde_json::Value;
+use tracing::debug;
+
+#[derive(Debug, Default)]
+pub struct Bool {}
+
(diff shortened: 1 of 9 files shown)


dsc_lib/src/functions/bool.rs
```diff
@@ -0,0 +1,83 @@
...
+    fn invoke(&self, args: &[Value], _context: &Context) -> Result<Value, DscError> {
+        debug!("{}", t!("functions.bool.invoked"));
+        if let Some(arg) = args[0].as_str() {
+            match arg.to_lowercase().as_str() {
+                "true" => Ok(Value::Bool(true)),
+                "false" => Ok(Value::Bool(false)),
+                _ => Err(DscError::Parser(t!("functions.invalidArguments").to_string())),
+            }
+        } else if let Some(num) = args[0].as_i64() {
+            Ok(Value::Bool(num != 0))
```
tgauth, line 44: Should bool support numeric values that are not `0` or `1`, as it does here, or should it error?

SteveL-MSFT: Good question.  Looking at the ARM code, they are also simply comparing if != 0.

tgauth (approved):

PR #981, opened 2025-07-19
Add reference docs for logical functions

Gijsreyn:
# PR Summary

PR #979 got merged, so I decided to hit up this PR to get it documented.


Gijsreyn:
@SteveL-MSFT, as it is still fresh in your mind, if you have time, please review this one.

SteveL-MSFT (commented):
I focused on reviewing the description and they are all technically accurate.  Will wait for @michaeltlombardi to perform editorial review

docs/reference/schemas/config/functions/bool.md
```diff
@@ -0,0 +1,132 @@
...
+```Syntax
+bool(<value>)
+```
+
+## Description
+
+The `bool()` function converts a string or number to a boolean value. For string arguments,
+it accepts "true" (case-insensitive) which converts to `true`, and "false" (case-insensitive)
+which converts to `false`. For numeric arguments, zero converts to `false` and any non-zero
+value converts to `true`.
```
michaeltlombardi, line 25: What happens when I specify `bool("blue")`? Does that raise an error?

michaeltlombardi (commented):
Minor comment about the behavior for strings, but otherwise this looks great!

docs/reference/schemas/config/functions/bool.md
SteveL-MSFT: Yes, that's an error.  For strings only `true` and `false` are accepted (case insensitive).  The error will come from the function parser which has declarative checking for parameters of functions before the function itself gets invoked:

```console
PS>         $configYaml = @'
>>         $schema: https://aka.ms/dsc/schemas/v3/bundled/config/document.json
>>         resources:
>>           - name: test
>>             type: Microsoft.DSC.Debug/Echo
>>             properties:
>>               output: "[bool('hello')]"
>> '@
[5.98ms] /Users/steve/repos/DSC [origin/output]
PS> dsc config get -i $configYaml
2025-07-28T16:20:25.782494Z ERROR Parser: Invalid argument(s)
```

michaeltlombardi: Then I think I'm happy with this PR once we explicitly clarify that any other value raises an error.

Gijsreyn: Note added!

michaeltlombardi (approved):

PR #990, opened 2025-07-22
Add new comparison operators

### LINKED ISSUE #988: Comparison functions

As a user I would like to have additional comparison functions so I can perform evaluation on if a resource should be executed.

https://docs.github.com/en/actions/reference/evaluate-expressions-in-workflows-and-actions

Gijsreyn:
# PR Summary

Just trying to learn some Rust on the side and address #988 and check some boxes in #57 

## PR Context

This pull request adds the following comparison operators:

* Less
* LessOrEquals
* Greater
* GreaterOrEquals

fix #988

### DIFF
--- a/dsc/src/args.rs
+++ b/dsc/src/args.rs
@@ -280,6 +280,7 @@ pub enum SchemaType {
     TestResult,
     ResolveResult,
     DscResource,
+    Resource,
     ResourceManifest,
     Include,
     Configuration,
--- a/dsc/src/util.rs
+++ b/dsc/src/util.rs
@@ -10,6 +10,7 @@ use dsc_lib::{
     configure::{
         config_doc::{
             Configuration,
+            Resource,
             RestartRequired,
         },
         config_result::{
@@ -163,6 +164,9 @@ pub fn get_schema(schema: SchemaType) -> Schema {
         }
         SchemaType::DscResource => {
             schema_for!(DscResource)
+        },
+        SchemaType::Resource => {
+            schema_for!(Resource)
         },
         SchemaType::ResourceManifest => {
             schema_for!(ResourceManifest)
--- a/dsc/tests/dsc_resource_schema.tests.ps1
+++ b/dsc/tests/dsc_resource_schema.tests.ps1
@@ -1,0 +1,30 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+Describe 'tests for the Resource schema within a configuration' {
+    It 'Unknown properties are an error' {
+        $yaml = @'
+            $schema: https://aka.ms/dsc/schemas/v3/bundled/config/document.json
+            resources:
+            - name: test
+              type: Microsft.Dsc.Debug/Echo
+              unknownProperty: true
+              properties:
+              output: "Hello World"
+'@
+        dsc config get -i $yaml 2>$TestDrive/error.log
+        $LASTEXITCODE | Should -Be 2
+        (Get-Content $TestDrive/error.log -Raw) | Should -Match 'Error: JSON: unknown field `unknownProperty`'
+
+    }
+
+    It 'dsc schema returns a valid schema' {
+        $schema = dsc schema -t resource
+        $LASTEXITCODE | Should -Be 0
+        $schema | Should -Not -BeNullOrEmpty
+        $schema = $schema | ConvertFrom-Json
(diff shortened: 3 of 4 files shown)


Gijsreyn:
@SteveL-MSFT or @tgauth, I know it is considered for v3.3, but if you have time, would you mind a flexible review?

dsc_lib/src/functions/greater.rs
```diff
@@ -0,0 +1,78 @@
...
+
+        if let (Some(num1), Some(num2)) = (first.as_f64(), second.as_f64()) {
+            return Ok(Value::Bool(num1 > num2));
+        }
+
+        if let (Some(str1), Some(str2)) = (first.as_str(), second.as_str()) {
+            return Ok(Value::Bool(str1 > str2));
+        }
+
+        Ok(Value::Bool(false))
```
SteveL-MSFT: if both args aren't numbers or both aren't strings, it should return an error of a type mismatch

dsc_lib/src/functions/greater.rs
```diff
@@ -0,0 +1,78 @@
...
+        2
+    }
+
+    fn invoke(&self, args: &[Value], _context: &Context) -> Result<Value, DscError> {
+        debug!("{}", t!("functions.greater.invoked"));
+
+        let first = &args[0];
+        let second = &args[1];
+
+        if let (Some(num1), Some(num2)) = (first.as_f64(), second.as_f64()) {
```
SteveL-MSFT: according to ARM template functions doc, we should use i64

dsc_lib/src/functions/greater_or_equals.rs
```diff
@@ -0,0 +1,77 @@
...
+        debug!("{}", t!("functions.greaterOrEquals.invoked"));
+        
+        let num1 = match &args[0] {
+            Value::Number(n) => n.as_f64().ok_or_else(|| DscError::Parser(t!("functions.invalidArguments").to_string()))?,
+            _ => return Err(DscError::Parser(t!("functions.invalidArguments").to_string())),
+        };
+
+        let num2 = match &args[1] {
+            Value::Number(n) => n.as_f64().ok_or_else(|| DscError::Parser(t!("functions.invalidArguments").to_string()))?,
+            _ => return Err(DscError::Parser(t!("functions.invalidArguments").to_string())),
```
SteveL-MSFT: This logic should be same as `greater()` except for testing `>=`

SteveL-MSFT (changes requested):
feedback applies to < and <=

dsc_lib/src/functions/greater_or_equals.rs
Gijsreyn: I have changed the logic and introduced a new localization message for it. According to the documentation, all functions should support both integers and strings. Additional Pester tests added. Can you review it once again? Thanks!

dsc/tests/dsc_expressions.tests.ps1
```diff
@@ -112,6 +112,28 @@ resources:
...
+        @{ expression = "[greaterOrEquals(5, 3)]"; expected = $true }
+        @{ expression = "[greaterOrEquals(3, 5)]"; expected = $false }
+        @{ expression = "[greaterOrEquals(5, 5)]"; expected = $true }
+        @{ expression = "[less(3, 5)]"; expected = $true }
+        @{ expression = "[less(5, 3)]"; expected = $false }
+        @{ expression = "[less(5, 5)]"; expected = $false }
+        @{ expression = "[lessOrEquals(3, 5)]"; expected = $true }
+        @{ expression = "[lessOrEquals(5, 3)]"; expected = $false }
+        @{ expression = "[lessOrEquals(5, 5)]"; expected = $true }
+        @{ expression = "[greater('b', 'a')]"; expected = $true }
```
SteveL-MSFT, line 127: Can you add or change one of the string tests to be `a` and `A` or `aa` and `Aa` to make sure casing is handled correctly?  Might even be worth having a test like `a` compared to `1`.  The result should be equivalent to `[string]::Compare('a', 'A')` in PowerShell

SteveL-MSFT (changes requested):

dsc/tests/dsc_expressions.tests.ps1
Gijsreyn: Added a couple more tests and also the type mismatch in Pester.

SteveL-MSFT (approved):

Issue #980, opened 2025-07-19
Config document resource schema is missing additionalProperties true


ThomasNieto:
The resource section of the config document does not prevent additional properties from being added.

```
resources:
- name: test
  type: Microsoft.Windows/Registry
  myRandomProperty: should not be here!
  properties:
    keyPath: HKCU\test
    valueName: abc
    valueData:
     String: def
```

<details>

<summary>Original issue about versioning but is duplicate</summary>

### Summary of the new feature / enhancement

As a user I want to be able to specify which version or version range of a resource I want to use. Then when running the configuration document if the resource version is not present error.

### Proposed technical implementation details (optional)

```yaml
type: owner[.group][.area]/name@1.2.3
name: instance_name
properties:
  propertyName: propertyValue
```

NuGet range syntax

```yaml
type: owner[.group][.area]/name@[1.0,2.0)
name: instance_name
properties:
  propertyName: propertyValue
```

</details>

SteveL-MSFT:
In the YAML, you would need to specify it as the `version` or `apiVersion` property, but in Bicep, it would take the `@version` syntax

ThomasNieto:
The schema for config document doesn't show anything about a `version` or `apiVersion` property. Adding version doesn't throw an error is the resource version doesn't exist. Also, this may have revealed a bug in the schema file where it doesn't set `additionalProperties` to `false`.

```
resources:
- name: test
  type: Microsoft.Windows/Registry
  version: 0.2.0
  properties:
    keyPath: HKCU\test
    valueName: abc
    valueData:
     String: def
```

SteveL-MSFT:
@ThomasNieto I thought I had made the change to add `apiVersion` to the schema, but I guess I had not yet.  You're right that `additionalProperties` should be `false`

This is all waiting on https://github.com/PowerShell/DSC/issues/543 to get implemented

ThomasNieto:
@SteveL-MSFT going to rename this issue to be about the `additionalProperties` since the original issue is a duplicate of #543

PR #995, opened 2025-07-23
Update `Resource` schema to match ARM

SteveL-MSFT:
# PR Summary

Schema for `Resource` within a config now matches ARM.  This means:

- unknown properties are rejected
- many unused ARM properties are allowed, but ignored currently
- decided not to alias `apiVersion` to `version` since that only helps YAML authors and intent is for people to use Bicep so adding the alias would just make it confusing in the future
- added `resource` to list of returned schemas: dsc schema -t resource

## PR Context

Fix https://github.com/PowerShell/DSC/issues/980


dsc/tests/dsc_resource_schema.tests.ps1
```diff
@@ -0,0 +1,30 @@
...
+    It 'dsc schema returns a valid schema' {
+        $schema = dsc schema -t resource
+        $LASTEXITCODE | Should -Be 0
+        $schema | Should -Not -BeNullOrEmpty
+        $schema = $schema | ConvertFrom-Json
+        $schema.'$schema' | Should -BeExactly 'https://json-schema.org/draft/2020-12/schema'
+        $schema.title | Should -BeExactly 'Resource'
+        $schema.additionalProperties | Should -Be $false
+    }
+}
```
tgauth:
```suggestion
}

```
nit: add newline

PR #997, opened 2025-07-24
Enable passing Bicep files directly to dsc

SteveL-MSFT:
# PR Summary

New `import` extension where the output to stdout needs to be a JSON string representing a valid DSC config file.
Bicep extension that implements this requiring the `bicep` exe to be in PATH to transpile the bicep file to JSON.
Still need to resolve with Bicep to not strictly validate the version to be in date format to allow semver or the word `latest`.

Some other changes:
- reformatted/ordered some of the `use` statements to make it more organized
- Discovery and DscManager `new()` no longer needs to return a DscError since no code is run for creation, so removed the `?` operator in many places
- Add new `get_extensions()` function to discovery which is used by `dsc` host since that is where file reading happens rather than in `dsc_lib`
- since `bicep` takes a positional arg for the file path, the args specified in the manifest can be an empty string and no arg name is passed, just the arg value (file path)

Note that until DSC support in Bicep is out of Experimental status, you must have a `bicepconfig.json` that enables that feature:

```json
{
  "experimentalFeaturesEnabled": {
    "desiredStateConfiguration": true
  }
}
```

Also made change to PSScript multi-warning test since traces are async, order is not guaranteed.

## PR Context

### DIFF
--- a/dsc_lib/src/extensions/dscextension.rs
+++ b/dsc_lib/src/extensions/dscextension.rs
@@ -41,6 +41,8 @@ pub enum Capability {
     Discover,
     /// The extension aids in retrieving secrets.
     Secret,
+    /// The extension imports configuration from a different format.
+    Import,
 }
 
 impl Display for Capability {
@@ -48,6 +50,7 @@ impl Display for Capability {
         match self {
             Capability::Discover => write!(f, "Discover"),
             Capability::Secret => write!(f, "Secret"),
+            Capability::Import => write!(f, "Import"),
         }
     }
 }
--- a/dsc_lib/src/extensions/dscextension.rs
+++ b/dsc_lib/src/extensions/dscextension.rs
@@ -2,13 +2,14 @@
 // Licensed under the MIT License.
 
 use rust_i18n::t;
+use path_absolutize::Absolutize;
 use serde::{Deserialize, Serialize};
 use serde_json::Value;
 use schemars::JsonSchema;
 use std::{fmt::Display, path::Path};
 use tracing::{debug, info, trace};
 
-use crate::{discovery::command_discovery::{load_manifest, ImportedManifest}, dscerror::DscError, dscresources::{command_resource::{invoke_command, process_args}, dscresource::DscResource}, extensions::secret::SecretResult};
+use crate::{discovery::command_discovery::{load_manifest, ImportedManifest}, dscerror::DscError, dscresources::{command_resource::{invoke_command, process_args}, dscresource::DscResource}, extensions::{import::ImportArgKind, secret::SecretResult}};
 
 use super::{discover::DiscoverResult, extension_manifest::ExtensionManifest, secret::SecretArgKind};
 
@@ -22,6 +23,8 @@ pub struct DscExtension {
     pub version: String,
     /// The capabilities of the resource.
     pub capabilities: Vec<Capability>,
+    /// The extensions supported for importing.
+    pub import_extensions: Option<Vec<String>>,
     /// The file path to the resource.
     pub path: String,
     /// The description of the resource.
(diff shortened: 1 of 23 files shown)


extensions/bicep/bicep.tests.ps1
```diff
@@ -0,0 +1,38 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+BeforeDiscovery {
+    $foundBicep = if ($null -ne (Get-Command bicep -ErrorAction Ignore)) {
+
+        $true
+    } else {
+        $false
+    }
```
tgauth:
```suggestion
    $foundBicep = if (Get-Command bicep -ErrorAction Ignore) {
        $true
    } else {
        $false
    }
```

tgauth: could also rename this `bicepNotFound` and flip the logic to avoid the `!` in the `-Skip` on line 13 below

dsc_lib/src/extensions/dscextension.rs
```diff
@@ -129,6 +136,62 @@ impl DscExtension {
...
+            let (_exit_code, stdout, _stderr) = invoke_command(
+                &import.executable,
+                args,
+                None,
+                Some(self.directory.as_str()),
+                None,
+                extension.exit_codes.as_ref(),
+            )?;
+            if stdout.is_empty() {
+                info!("{}", t!("extensions.dscextension.importNoResults", extension = self.type_name));
```
tgauth, line 185: After this info message, it will fall through to the unsupported capability error below - should it Error with a more relevant message or still return Ok?

dsc_lib/src/extensions/dscextension.rs
```diff
@@ -22,6 +23,8 @@ pub struct DscExtension {
     pub version: String,
     /// The capabilities of the resource.
     pub capabilities: Vec<Capability>,
+    /// The extensions supported for importing.
+    pub import_extensions: Option<Vec<String>>,
```
tgauth: I think extension is getting overloaded - maybe `import_file_extensions` or something like that would be more clear?

dsc_lib/src/extensions/import.rs
```diff
@@ -0,0 +1,28 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+use schemars::JsonSchema;
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Default, Clone, PartialEq, Deserialize, Serialize, JsonSchema)]
+pub struct ImportMethod {
+    /// The extensions to import.
+    pub extensions: Vec<String>,
```
tgauth: consider renaming to `file_extensions` to be more explicit

dsc_lib/src/extensions/import.rs
```diff
@@ -0,0 +1,28 @@
...
+    /// The command to run to get the state of the resource.
+    pub executable: String,
+    /// The arguments to pass to the command to perform a Get.
```
tgauth:
```suggestion
    /// The arguments to pass to the command to perform an Import.
```

extensions/bicep/bicep.tests.ps1
SteveL-MSFT: I tend to find negative named variables harder to read

dsc_lib/src/extensions/dscextension.rs
SteveL-MSFT: So the logic in the `dsc` host is to get all importers and try to import.  If they error, it just moves onto the next one until one successfully imports.  So this code supports that.

tgauth (approved):

PR #999, opened 2025-07-24
Comparison and object funcions coalesce, create_object and null

Gijsreyn:
# PR Summary

This PR adds the following functions:

* `coalesce`
* `create_object`
* `null`

Part of #57.


dsc_lib/src/functions/coalesce.rs
```diff
@@ -0,0 +1,111 @@
...
+#[derive(Debug, Default)]
+pub struct Coalesce {}
+
+impl Function for Coalesce {
+    fn description(&self) -> String {
+        t!("functions.coalesce.description").to_string()
+    }
+
+    fn category(&self) -> FunctionCategory {
+        FunctionCategory::Logical
```
SteveL-MSFT: The docs say this is `comparison`https://learn.microsoft.com/en-us/azure/azure-resource-manager/templates/template-functions-comparison#coalesce

dsc_lib/src/functions/coalesce.rs
```diff
@@ -0,0 +1,111 @@
...
+        Ok(Value::Null)
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use crate::configure::context::Context;
+    use crate::parser::Statement;
+    use super::*;
+    // TODO: Add tests for direct function calls with nulls and mixed types if the parser accept it
```
SteveL-MSFT: I would suggest removing this TODO and having these types of tests as Pester tests.  We're missing some functions to create objects and arrays within the expression.

SteveL-MSFT (changes requested):
Add Pester tests in dsc_functions.tests.ps1

dsc_lib/src/functions/coalesce.rs
Gijsreyn: I have now included the `null()` and `create_object` functions also. Moved up the tests a little to seggregate based on the types.

Note that I have also removed the documents. I'll create a separate PR for it.

Gijsreyn:
> Add Pester tests in dsc_functions.tests.ps1. Use `parameters` with default values for objects, arrays, etc...

Are the current tests fine, or you still want more in this file?

dsc_lib/src/functions/null.rs
```diff
@@ -0,0 +1,83 @@
...
+
+    #[test]
+    fn null_in_object() {
+        let mut parser = Statement::new().unwrap();
+        let result = parser.parse_and_execute("[createObject('key', null())]", &Context::new()).unwrap();
+        assert_eq!(result.to_string(), r#"{"key":null}"#);
+    }
+
+    #[test]
+    fn null_in_array() {
```
SteveL-MSFT: Is this test supposed to be the same as line 64?

dsc/tests/dsc_expressions.tests.ps1
```diff
@@ -138,6 +135,53 @@ resources:
         @{ expression = "[lessOrEquals('b', 'a')]"; expected = $false }
         @{ expression = "[lessOrEquals('a', 'a')]"; expected = $true }
         @{ expression = "[lessOrEquals('aa', 'Aa')]"; expected = $false }
+        @{ expression = "[coalesce('hello', 'world')]" ; expected = 'hello' }
```
SteveL-MSFT: You should add some tests using `null()` with `coalesce()` here

SteveL-MSFT (changes requested):

tree-sitter-dscexpression/.gitignore
```diff
@@ -1,27 +1,25 @@
 # Rust artifacts
 target/
+Cargo.lock
```
SteveL-MSFT: Why are these tree-sitter-dscexpression files part of this PR?

dsc_lib/src/functions/null.rs
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    #[test]
+    fn null_in_object() {
+        let mut parser = Statement::new().unwrap();
+        let result = parser.parse_and_execute("[createObject('key', null())]", &Context::new()).unwrap();
+        assert_eq!(result.to_string(), r#"{"key":null}"#);
+    }
+
+    // TODO: Fix this test when createArray supports null
```
SteveL-MSFT: https://learn.microsoft.com/en-us/azure/azure-resource-manager/templates/template-functions-array#createarray doesn't indicate that `null` is accepted unless that's changed for language 2.0

SteveL-MSFT (changes requested):

tree-sitter-dscexpression/.gitignore
Gijsreyn: Somehow these are added when I run the build script. Removed.

PR #1002, opened 2025-07-24
Fix to allow default value to be set as null

### LINKED ISSUE #1000: Parameter null value is not allowed
### Prerequisites

### Summary

While developing the `coalesce` function and writing the documentation, I stumbled on the following error from example 3:

Parser: Parameter 'customValue' not found in context.

It looks like parameter default values set to `null` are not allowed.

### Steps to reproduce

Run the following example from the branch comparison-function-coalesce:

```powershell
$in = @'
parameters:
  customValue:
    type: string
    defaultValue: null
  timeout:
    type: int
    defaultValue: 0
resources:
- name: Configuration with fallbacks
  properties:
    output:
      configValue: "[coalesce(parameters('customValue'), 'default-config')]"
      timeout:     "[coalesce(parameters('timeout'), 30)]"
'@
dsc config get --input $in
```

### Expected behavior

```console
The example would have returned null on `configValue`
```

### Actual behavior

```console
The error: Parser: Parameter 'customValue' not found in context occurs
```

### Error details

```console

```

### Environment data

```PowerShell
PSVersion                      7.5.2
GitCommitId                    7.5.2
```

### Version

dsc 3.2.0-preview.3

### Visuals

_No response_

Gijsreyn:
# PR Summary

Allows default values to be null in parameters. This allows the coalesce function to work correctly.

## PR Context

Fix #1000.

> [!NOTE]
> Used the following as reference: https://github.com/serde-rs/serde/issues/1098

### DIFF
--- a/dsc/tests/dsc_parameters.tests.ps1
+++ b/dsc/tests/dsc_parameters.tests.ps1
@@ -236,6 +236,25 @@ Describe 'Parameters tests' {
         $out.results[3].result.actualState.output | Should -BeExactly @('hello', 'world')
     }
 
+    It 'Null default value is handled correctly' {
+        $config_yaml = @"
+          `$schema: https://aka.ms/dsc/schemas/v3/bundled/config/document.json
+          parameters:
+            customValue:
+              type: string
+              defaultValue: null
+          resources:
+          - name: Configuration with fallbacks
+            type: Microsoft.DSC.Debug/Echo
+            properties:
+              output:
+                configValue: "[parameters('customValue')]"
+"@
+        $out = $config_yaml | dsc config get -f - | ConvertFrom-Json
+        $LASTEXITCODE | Should -Be 0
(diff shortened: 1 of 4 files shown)


Gijsreyn:
Probably this raises bigger questions or concerns, which I cannot oversee.  However, the end goal should be able to do something like:

```bicep
targetScope = 'desiredStateConfiguration'

parameter optionalParameter string?

resource echo 'Microsoft.DSC.Debug/Echo@2025-01-01' = {
  name: 'exampleEcho'
  properties: {
    output: 'Hello, world!'
    mystring: ?optionalParameter
  }
```

SteveL-MSFT:
`nullable` is an ARM language 2.0 feature, so this wouldn't be the way to add that capability in the engine.  There's some other language 2.0 items that need to be added as part of Bicep work, so I'll take care of that later

PR #1004, opened 2025-07-24
Add sshdconfig get

tgauth:
# PR Summary
- add get support to sshdconfig resource:
  - if no input is provided, get returns all settings from `SSHD -T` including defaults
  - if input is provided, get returns the subset of settings from `SSHD -T` included in the input
  - if `defaults` false is passed in via `_metadata`, get returns the subset of settings from `SSHD -T` that are not defaults
  - if `filepath` is passed in via `_metadata`, get runs `SSHD -T -f <filepath>` 
 - add pester tests for sshdconfig get and export
 - update resource schema to allow additional properties


dsc/tests/dsc_sshdconfig.tests.ps1
```diff
@@ -0,0 +1,105 @@
...
+- name: sshdconfig
+  type: Microsoft.OpenSSH.SSHD/sshd_config
+  properties:
+'@
+        # set a non-default value in a temporary sshd_config file
+        "LogLevel Debug3" | Set-Content -Path $TestDrive/test_sshd_config
+    }
+
+    AfterAll {
+        if (Test-Path $TestDrive/test_sshd_config) {
```
SteveL-MSFT: You don't need to explicitly cleanup as Pester will automatically delete `$TestDrive`

dsc/tests/dsc_sshdconfig.tests.ps1
```diff
@@ -0,0 +1,105 @@
...
+            Remove-Item -Path $TestDrive/test_sshd_config -Force
+        }
+    }
+
+    It 'Export works' {
+        if ($runTest) {
```
SteveL-MSFT: `Describe` now has a `-Skip` you can use with `$runTest` instead of having it in the test itself, this way you can see in the results tests were skipped

dsc/tests/dsc_sshdconfig.tests.ps1
```diff
@@ -0,0 +1,105 @@
...
+            $out = dsc config export -i "$yaml" | ConvertFrom-Json -Depth 10
+            $LASTEXITCODE | Should -Be 0
+            $out.resources.count | Should -Be 1
+            $out.resources[0].properties | Should -Not -BeNullOrEmpty
+            $out.resources[0].properties.port[0] | Should -Be 22
+        }
+    }
+
+    It 'Get works'{
```
SteveL-MSFT: Is there any real difference in this case between `get` and `export`?  If not, just have it as one test with `-TestCases` to perform the different operations but have the same validation code

dsc/tests/dsc_sshdconfig.tests.ps1
```diff
@@ -0,0 +1,105 @@
...
+  type: Microsoft.OpenSSH.SSHD/sshd_config
+  properties:
+    passwordauthentication: 'no'
+'@
+            $out = dsc config get -i "$get_yaml" | ConvertFrom-Json -Depth 10
+            $LASTEXITCODE | Should -Be 0
+            $out.results.count | Should -Be 1
+            $out.results.result.actualState.count | Should -Be 1
+            $out.results.result.actualState.passwordauthentication | Should -Be 'yes'
+            $out.results.result.actualState.port | Should -BeNullOrEmpty
```
SteveL-MSFT: If the expectation is that the result only has 1 property `passwordauthentication`, then it would be better to check the count of properties:

```powershell
($out.results.result.actualState.psobject.properties | measure-object).count | Should -Be 1
```

dsc/tests/dsc_sshdconfig.tests.ps1
```diff
@@ -0,0 +1,105 @@
...
+`$schema: https://aka.ms/dsc/schemas/v3/bundled/config/document.json
+metadata:
+  Microsoft.DSC:
+    securityContext: elevated
+resources:
+- name: sshdconfig
+  type: Microsoft.OpenSSH.SSHD/sshd_config
+  properties:
+    _metadata:
+        defaults: false
```
SteveL-MSFT: Maybe `defaults` should be called `includeDefaults`

sshdconfig/src/get.rs
```diff
@@ -9,20 +9,23 @@ use {
...
+use crate::export::invoke_export_to_map;
+use crate::util::{extract_metadata_from_input, extract_sshd_defaults, SshdCommandArgs};
 
 /// Invoke the get command.
 ///
 /// # Errors
 ///
 /// This function will return an error if the desired settings cannot be retrieved.
-pub fn invoke_get(setting: &Setting) -> Result<(), SshdConfigError> {
+pub fn invoke_get(input: Option<&String>, setting: &Setting) -> Result<(), SshdConfigError> {
```
SteveL-MSFT: My preference would be that these helper functions return the result of the operation and the `main` code writes it to STDOUT so ideally any writing of output is in a single place.

sshdconfig/src/get.rs
```diff
@@ -82,3 +85,68 @@ fn get_default_shell() -> Result<(), SshdConfigError> {
...
+                    SshdCommandArgs {
+                        filepath: Some(path.clone()),
+                        additional_args: None,
+                    }
+                );
+            } else {
+                return Err(SshdConfigError::InvalidInput(t!("get.filepathMustBeString").to_string()));
+            }
+        }
+    }
```
SteveL-MSFT: I think handling of metadata should be a separate function and put the results into a struct

sshdconfig/src/util.rs
```diff
@@ -2,11 +2,41 @@
...
 use std::process::Command;
+use tracing::debug;
 use tracing_subscriber::{EnvFilter, filter::LevelFilter, Layer, prelude::__tracing_subscriber_SubscriberExt};
 
 use crate::error::SshdConfigError;
+use crate::parser::parse_text_to_map;
 
+pub struct CommandInfo {
+    /// metadata provided with the command
+    pub metadata: Map<String, Value>,
```
SteveL-MSFT: Having metadata strongly typed helps it to reject anything it doesn't understand

sshdconfig/src/util.rs
```diff
@@ -36,16 +66,26 @@ pub fn enable_tracing() {
 /// # Errors
 ///
 /// This function will return an error if sshd -T fails to validate `sshd_config`.
-pub fn invoke_sshd_config_validation() -> Result<String, SshdConfigError> {
+pub fn invoke_sshd_config_validation(args: Option<SshdCommandArgs>) -> Result<String, SshdConfigError> {
     let sshd_command = if cfg!(target_os = "windows") {
```
SteveL-MSFT: I don't think on Windows you MUST have the `.exe` extension, so it might be sufficient to not have this code and always use `sshd`

SteveL-MSFT (changes requested):

SteveL-MSFT (approved):
LGTM

michaeltlombardi:
It's a bit late, since the implementation is already so far along, but I want to address a few things from the design perspective:

> - if no input is provided, get returns all settings from SSHD -T including defaults
> - if input is provided, get returns the subset of settings from SSHD -T included in the input
> - if `defaults: false` is passed in via `_metadata`, get returns the subset of settings from `SSHD -T` that are not defaults
> - if `filepath: <filepath>` is passed in via `_metadata`, get runs `SSHD -T -f <filepath>`

The differing behavior for input/no-input in the `get` operation here is a break from prior design, user expectation, and integration with higher order tooling. Typically, `get` operations return the _full_ configurable state for a resource instance. This filtering makes more sense to me for the `export` operation. This design also prevents me from retrieving the full configuration but only _verifying or enforcing_ a specific subset.

Similarly, here we seem to be defining write-only properties (behavior-modifying parameters) to be passed through `_metadata`. Arguably, `filepath` is a non-required key property - I would expect a configuration document containing any two instances of this resource with the same `filepath` (or `filepath` not being defined) to raise a conflict in the future.

SteveL-MSFT:
@michaeltlombardi seems like we should have a meeting to discuss this before we commit this change

tgauth:
> @michaeltlombardi seems like we should have a meeting to discuss this before we commit this change

I'll set it up!

SteveL-MSFT (approved):

dsc/tests/dsc_sshdconfig.tests.ps1
```diff
@@ -0,0 +1,136 @@
...
+        $yaml = @"
+`$schema: https://aka.ms/dsc/schemas/v3/bundled/config/document.json
+metadata:
+  Microsoft.DSC:
+    securityContext: elevated
+resources:
+- name: sshdconfig
+  type: Microsoft.OpenSSH.SSHD/sshd_config
+  properties:
+    _metadata:
```
SteveL-MSFT: Input metadata should be a peer to properties.  It could be passed this way since that's how DSC does it internally, but I think users looking at our tests as examples should see it passed at a higher level to the resource.

tgauth: Like this? The input metadata is processed by DSC but doesn't seem to be passed to the resource (at least for `get` or `export`) unless I also need to make a change to the manifest?
```
- name: sshdconfig
  type: Microsoft.OpenSSH.SSHD/sshd_config
  metadata:
    filepath: 
  properties:
```

tgauth: I think I see where the engine needs to be updated: https://github.com/PowerShell/DSC/blob/main/dsc_lib/src/configure/mod.rs#L173
I can go ahead and make that change in a separate PR?

SteveL-MSFT: Yes, create a new PR if that isn't working as expected, then we'll merge that one first and get back to this one.

sshdconfig/src/get.rs
```diff
@@ -29,7 +29,7 @@ use crate::util::{
 ///
 /// This function will return an error if the desired settings cannot be retrieved.
 pub fn invoke_get(input: Option<&String>, setting: &Setting) -> Result<Map<String, Value>, SshdConfigError> {
-    debug!("{}: {:?}", t!("get.debugSetting").to_string(), setting);
+    debug!("{}: {:?} {:?}", t!("get.debugSetting").to_string(), setting, input);
```
SteveL-MSFT: If the raw `input` is being logged, it should be a `trace`.  You can always have a separate `trace` rather than being part of the `debug` message.

PR #1005, opened 2025-07-24
Add `contains()`, `union()`, `length()`, and `empty()` Array functions

SteveL-MSFT:
# PR Summary

Add `contains()`, `union()`, `length()`, and `empty()` array functions

I would suggest using the ARM docs for describing the behaviors which isn't completely obvious:
https://learn.microsoft.com/en-us/azure/azure-resource-manager/templates/template-functions-array#contains
https://learn.microsoft.com/en-us/azure/azure-resource-manager/templates/template-functions-array#union
https://learn.microsoft.com/en-us/azure/azure-resource-manager/templates/template-functions-array#length
https://learn.microsoft.com/en-us/azure/azure-resource-manager/templates/template-functions-array#empty

## PR Context


SteveL-MSFT:
Looks like `length()` and `empty()` are also Array functions, so will add those to this PR

dsc_lib/src/functions/contains.rs
```diff
@@ -0,0 +1,115 @@
...
+    fn accepted_arg_types(&self) -> Vec<AcceptedArgKind> {
+        vec![AcceptedArgKind::Array, AcceptedArgKind::Object, AcceptedArgKind::String, AcceptedArgKind::Number]
+    }
+
+    fn invoke(&self, args: &[Value], _context: &Context) -> Result<Value, DscError> {
+        debug!("{}", t!("functions.contains.invoked"));
+        let mut found = false;
+
+        let (string_to_find, number_to_find) = if let Some(string) = args[1].as_str() {
+            (string.to_string(), 0)
```
tgauth: I think assigning `number_to_find` to `0` here could potentially allow for an edge case when the container is an array of integers. For example, `[contains(createArray(1,0), 'z')]` should return `false`, but would return `true` based on line 57.

SteveL-MSFT: Good catch, I think you're right.  Will rework this.

SteveL-MSFT: Changed logic to use `None` and added a test case for your scenario

tgauth (approved):

PR #1010, opened 2025-07-27
Add `outputs` support in configuration

SteveL-MSFT:
# PR Summary

Currently blocked on https://github.com/Azure/bicep/issues/17670 where bicep is crashing when trying to format the resourceId

Add support to use `outputs` section of configuration to have custom output as part of the result.  This is needed to use DSC bicep files as modules.

## PR Context

### DIFF
--- a/dsc/examples/hello_world.dsc.bicep
+++ b/dsc/examples/hello_world.dsc.bicep
@@ -9,3 +9,5 @@ resource echo 'Microsoft.DSC.Debug/Echo@2025-01-01' = {
     output: 'Hello, world!'
   }
 }
+
+output exampleOutput string = echo.properties.output
--- a/dsc_lib/locales/en-us.toml
+++ b/dsc_lib/locales/en-us.toml
@@ -71,6 +71,11 @@ propertyNotString = "Property '%{name}' with value '%{value}' is not a string"
 metadataMicrosoftDscIgnored = "Resource returned '_metadata' property 'Microsoft.DSC' which is ignored"
 metadataNotObject = "Resource returned '_metadata' property which is not an object"
 metadataRestartRequiredInvalid = "Resource returned '_metadata' property '_restartRequired' which contains invalid value: %{value}"
+skippingOutput = "Skipping output for '%{name}' due to condition evaluating to false"
+outputValueNotDefined = "Output value '%{name}' is not defined"
+secureOutputSkipped = "Secure output '%{name}' is skipped"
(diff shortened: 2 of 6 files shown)


Gijsreyn:
Hey @SteveL-MSFT, 

In an attempt to fix and test this out locally, I stumbled on a few issues. With the change added to the Bicep CLI at: https://github.com/Azure/bicep/pull/17805, the produced JSON looks like:

```json
{
  "contentVersion": "1.0.0.0",
  "metadata": {
    "_generator": {
      "name": "bicep",
      "version": "0.37.52.64608",
      "templateHash": "928342520835393415"
    }
  },
  "resources": [
    {
      "type": "Microsoft.DSC.Debug/Echo",
      "apiVersion": "2025-01-01",
      "name": "exampleEcho",
      "properties": {
        "output": "Hello, world!"
      }
    }
  ],
  "outputs": {
    "exampleOutput": {
      "type": "string",
      "value": "[reference(concat('Microsoft.DSC.Debug/Echo', '/', 'exampleEcho'), '2025-01-01').output]"
    }
  }
}
```

When I'm running that through your branch, another error pops up: `Error: JSON: unknown field `value`, expected one of `condition`, `type`, `value_or_copy` at line 1 column 392`

Looking at the [docs](https://learn.microsoft.com/en-us/azure/azure-resource-manager/templates/outputs?tabs=azure-powershell), I would say that in the output object, it can be a value or a copy, not what the current implementation has. If I'm running it like this:

```json
"outputs": {
    "exampleOutput": {
      "type": "string",
      "value_or_copy": {
        "value": "[reference(concat('Microsoft.DSC.Debug/Echo', '/', 'exampleEcho'), '2025-01-01').output]"
      }
    }
  }
```

It at least proceeds further, but comes with the following error message: `Parser: Function 'reference' requires exactly 1 arguments`. I've seen some changes with that particular `apiVersion` at the end. However, I cannot say much about it with my current knowledge.

Nevertheless, one more time try with `"value": "[reference(concat('Microsoft.DSC.Debug/Echo', '/', 'exampleEcho')).output]"` specified, gives me: `Parser: Invalid resourceId or resource has not executed yet: Microsoft.DSC.Debug/Echo/exampleEcho`. This is probably still a WIP.

I wanted to share this to see if improvements can be made at the Bicep side or DSC. Hope it helps further in the development.

SteveL-MSFT:
@Gijsreyn the `value_or_copy` problem just required use of `flatten` so that it doesn't actually show up in the JSON.  Since this works outside of bicep, won't block on this for now and will revisit later when we have bicep integration working.

PR #1025, opened 2025-08-02
Experimental WinPowerShell DSC import extension

Gijsreyn:
# PR Summary

This PR introduces a Windows PowerShell DSC import extension that enables the passing in of PSDSC v2 configuration documents.

## PR Context

The PR adds three new files to the build:

- `convert-resource.ps1`
- `convertDscResource.psd1`
- `convertDscResource.psm1`

When a user passes in a `.ps1` file with a valid DSC configuration document, an attempt is made to convert it by borrowing content from [DSCParser](https://github.com/microsoft/DSCParser).


extensions/powershell/powershell.dsc.extension.json
```diff
@@ -0,0 +1,23 @@
+{
+    "$schema": "https://aka.ms/dsc/schemas/v3/bundled/resource/manifest.json",
+    "type": "Microsoft.DSC.Extension/PowerShell",
+    "version": "0.1.0",
+    "description": "Enable passing PowerShell v2 configuration document file directly to DSC, but requires pwsh executable to be available.",
+    "import": {
+        "fileExtensions": ["ps1"],
+        "executable": "pwsh",
```
ThomasNieto: Should this be Windows PowerShell since `configuration` keywords are a DSC v1 construct?

Gijsreyn: Yes, I was about to put it back on draft to make those changes. Will finish it later. Thanks for the check.

Gijsreyn:
@SteveL-MSFT Any idea why the Windows build failed on fetching packages?

SteveL-MSFT:
@Gijsreyn I'd hold off on this for now, I'm working on a big PR to fix the Windows CI.  Basically there was one issue in a test which caused Pester to crash so even though Windows tests have been failing, it was never reported.  Since making that fix, many existing tests are failing which I'm now fixing one by one.

Gijsreyn:
Hi @SteveL-MSFT, this one was a pretty tough one, especially with the latest fixes on the `build.ps1` script. Yet, I had a successful run at: https://github.com/PowerShell/DSC/actions/runs/16867533276. Then I thought it should be relatively easy to include `pwsh.exe` as well, which I have now done. However, this raised another issue because both extensions _accept_ .ps1. 

Now there's a conflict between the ones. Is there another indication to tell DSC's engine what extension to use?

extensions/powershell/windowspowershell.dsc.extension.json
```diff
@@ -0,0 +1,23 @@
+{
+    "$schema": "https://aka.ms/dsc/schemas/v3/bundled/resource/manifest.json",
+    "type": "Microsoft.Windows.Extension/WindowsPowerShell",
```
SteveL-MSFT: I think you should call this `Microsoft.DSC.Transitional/PSDesiredStateConfiguration`

SteveL-MSFT (changes requested):

extensions/powershell/windowspowershell.dsc.extension.json
Gijsreyn: Check!

Gijsreyn:
Closed as discussed at the working group; not to be implemented here.

PR #1028, opened 2025-08-06
Refactor functions on how they return metadata

SteveL-MSFT:
# PR Summary

Refactor how functions are implemented so that:

- arg validation happens before function is invoked including remaining args and mandatory args
- metadata is in a single struct that can be extended in the future
- returnType is now added and used for listing functions

### DIFF
--- a/dsc_lib/src/functions/mod.rs
+++ b/dsc_lib/src/functions/mod.rs
@@ -194,41 +194,34 @@ impl FunctionDispatcher {
                 break;
             }
 
-            if value.is_array() && !metadata.accepted_arg_ordered_types[index].contains(&FunctionArgKind::Array) {
-                return Err(DscError::Parser(t!("functions.noArrayArgs", name = name, accepted_args_string = metadata.accepted_arg_ordered_types[index].iter().map(std::string::ToString::to_string).collect::<Vec<_>>().join(", ")).to_string()));
-            } else if value.is_boolean() && !metadata.accepted_arg_ordered_types[index].contains(&FunctionArgKind::Boolean) {
-                return Err(DscError::Parser(t!("functions.noBooleanArgs", name = name, accepted_args_string = metadata.accepted_arg_ordered_types[index].iter().map(std::string::ToString::to_string).collect::<Vec<_>>().join(", ")).to_string()));
-            } else if value.is_null() && !metadata.accepted_arg_ordered_types[index].contains(&FunctionArgKind::Null) {
-                return Err(DscError::Parser(t!("functions.noNullArgs", name = name, accepted_args_string = metadata.accepted_arg_ordered_types[index].iter().map(std::string::ToString::to_string).collect::<Vec<_>>().join(", ")).to_string()));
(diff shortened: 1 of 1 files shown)


dsc_lib/src/functions/mod.rs
```diff
@@ -168,20 +188,43 @@ impl FunctionDispatcher {
...
+                return Err(DscError::Parser(t!("functions.noObjectArgs", name = name, accepted_args_string = metadata.accepted_arg_ordered_types[index].iter().map(std::string::ToString::to_string).collect::<Vec<_>>().join(", ")).to_string()));
+            } else if value.is_string() && !metadata.accepted_arg_ordered_types[index].contains(&FunctionArgKind::String) {
+                return Err(DscError::Parser(t!("functions.noStringArgs", name = name, accepted_args_string = metadata.accepted_arg_ordered_types[index].iter().map(std::string::ToString::to_string).collect::<Vec<_>>().join(", ")).to_string()));
+            }
+        }
+
+        // if we have remaining args, they must match one of the remaining_arg_types
+        if let Some(remaining_arg_types) = metadata.remaining_arg_accepted_types {
+            for value in args.iter().skip(metadata.accepted_arg_ordered_types.len()) {
+                if value.is_array() && !remaining_arg_types.contains(&FunctionArgKind::Array) {
```
tgauth: Might be worth encapsulating this logic in a helper function that checks if the value type is allowed based on a vec of arg types instead of duplicating for `metadata.accepted_arg_ordered_types[index]` & `remaining_arg_types`

dsc/locales/en-us.toml
```diff
@@ -115,7 +115,7 @@ tableHeader_functionName = "Function"
 tableHeader_functionCategory = "Category"
 tableHeader_minArgs = "MinArgs"
 tableHeader_maxArgs = "MaxArgs"
-tableHeader_argTypes = "ArgTypes"
+tableHeader_argTypes = "ReturnTypes"
```
tgauth, line 118: Would it be helpful to include both input args and return types in the table, instead of replacing an existing column?

tgauth:
I wonder if it is worth further abstracting some of the input arg validation for functions like `concat` that accept multiple arg types, but expect all args to be of a single type

dsc/locales/en-us.toml
SteveL-MSFT: I think the accepted types should be part of the returned JSON, but it's probably too much info (as it's now an array of arrrays) to put into the table which is more for human use.

SteveL-MSFT:
> I wonder if it is worth further abstracting some of the input arg validation for functions like `concat` that accept multiple arg types, but expect all args to be of a single type

If there is sufficient functions that have same pattern, then I would say yes, but if it's more of a one-off, I'd leave it to the individual function to validate.  Many of the comparison functions have similar pattern where both have to be string or number, but I think the rust code doesn't change so it doesn't seem to be saving anything if pre-check.

tgauth (approved):

PR #1032, opened 2025-08-07
Add `endsWith`, `startsWith`, `utcNow`, and `uniqueString` functions

SteveL-MSFT:
# PR Summary

Add `endsWith()`, `startsWith()`, `utcNow()`, and `uniqueString()` functions.  Note that `utcNow()` supports a subset of the .NET datetime formatting strings.

Also refactored how functions get registered with DSC so that it's from a vector and takes the name from the metadata so no risk of them being out of sync.

## PR Context


tgauth (approved):
LGTM

Issue #428, opened 2024-04-30
Synthetic `export` using `get`


SteveL-MSFT:
### Summary of the new feature / enhancement

Many resources may not support `export` and something like `registry` it doesn't make sense to export the whole registry.  If you have a configuration that mixes resources that directly support `export` and ones that would work with `get`, it would make sense to allow exporting that entire configuration so it can be applied elsewhere.

PR #1035, opened 2025-08-07
Add support for using `get` if resource doesn't implement `export`

SteveL-MSFT:
# PR Summary

Enable exporting a configuration that specifies resources that only implement `get`.  This change also makes the `OSInfo` resource only implement `get` (since `export` literally did the same operation).  Added a new test resource to help validate the case where input is provided as a filter and where none is provided.

## PR Context

Fix https://github.com/PowerShell/DSC/issues/428


dsc/tests/dsc_export.tests.ps1
```diff
@@ -181,4 +181,42 @@ resources:
...
+      $out.resources[0].properties.name | Should -BeExactly 'two'
+      $out.resources[0].properties.id | Should -Be 2
+    }
+
+    It 'Export can be used with a resource that only implements Get with no filter' {
+      $yaml = @'
+$schema: https://aka.ms/dsc/schemas/v3/bundled/config/document.json
+resources:
+  - name: OS
+    type: Microsoft/OSInfo
```
tgauth: might need to update this test if OSInfo resource continues to implement export

SteveL-MSFT: good catch.  I'll update the test resource to cover this scenario

PR #1038, opened 2025-08-08
Update OSInfo resource to return `_name`

SteveL-MSFT:
# PR Summary

Since we added the ability for resources to return `_name`, update OSInfo resource to use this as "<family> <version> <architecture>"

### DIFF
--- a/osinfo/src/config.rs
+++ b/osinfo/src/config.rs
@@ -72,7 +72,12 @@ impl OsInfo {
             family,
             version,
             architecture.as_deref().unwrap_or("")
-        ));
+        let name = Some(
+            match &architecture {
+                Some(arch) => format!("{:?} {} {}", family, version, arch),
+                None => format!("{:?} {}", family, version),
+            }
+        );
         Self {
             id: ID.to_string(),
             family,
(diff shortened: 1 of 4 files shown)


osinfo/src/config.rs
```diff
@@ -64,14 +77,22 @@ impl OsInfo {
             os_info::Bitness::X64 => Bitness::Bit64,
             _ => Bitness::Unknown,
         };
+        let version = os_info.version().to_string();
+        let name = Some(
+            match &architecture {
+                Some(arch) => format!("{family} {version} {arch}"),
+                None => format!("{family:?} {version}"),
+            }
+        );
```
michaeltlombardi: I think, if I understand correctly, that we only want this surfaced and returned from the `export` operation, not `get`:

- When directly invoking with `dsc resource get`, this property doesn't mean anything / is just noise.
- When invoking for the get operation with `dsc config get`, the name of the resource instance is already defined by the `name` field of the resource instance definition.
- When invoking for the export operation, we _do_ want to surface `_name`, so it can be hoisted to the effective `name` of the exported instance definition.

SteveL-MSFT: makes sense, updated

PR #1041, opened 2025-08-10
Add array(), first(), indexOf() functions

Gijsreyn:
# PR Summary

This pull request adds the `array()`, `first()`, and `indexOf()` functions.

## PR Context

Partially addresses #57

### DIFF
--- a/dsc/tests/dsc_functions.tests.ps1
+++ b/dsc/tests/dsc_functions.tests.ps1
@@ -2,18 +2,18 @@
 # Licensed under the MIT License.
 
 Describe 'tests for function expressions' {
-    It 'function works: <text>' -TestCases @(
-        @{ text = "[concat('a', 'b')]"; expected = 'ab' }
-        @{ text = "[concat('a', 'b', 'c')]"; expected = 'abc' }
-        @{ text = "[concat('a', concat('b', 'c'))]"; expected = 'abc' }
-        @{ text = "[base64('ab')]"; expected = 'YWI=' }
-        @{ text = "[base64(concat('a','b'))]"; expected = 'YWI=' }
-        @{ text = "[base64(base64(concat('a','b')))]"; expected = 'WVdJPQ==' }
-    ) {
-        param($text, $expected)
+  It 'function works: <text>' -TestCases @(
+    @{ text = "[concat('a', 'b')]"; expected = 'ab' }
+    @{ text = "[concat('a', 'b', 'c')]"; expected = 'abc' }
+    @{ text = "[concat('a', concat('b', 'c'))]"; expected = 'abc' }
+    @{ text = "[base64('ab')]"; expected = 'YWI=' }
+    @{ text = "[base64(concat('a','b'))]"; expected = 'YWI=' }
+    @{ text = "[base64(base64(concat('a','b')))]"; expected = 'WVdJPQ==' }
+  ) {
+    param($text, $expected)
 
-        $escapedText = $text -replace "'", "''"
-        $config_yaml = @"
+    $escapedText = $text -replace "'", "''"
+    $config_yaml = @"
             `$schema: https://aka.ms/dsc/schemas/v3/bundled/config/document.json
             resources:
             - name: Echo
@@ -21,17 +21,17 @@ Describe 'tests for function expressions' {
               properties:
                 output: '$escapedText'
 "@
-        $out = $config_yaml | dsc config get -f - | ConvertFrom-Json
-        $out.results[0].result.actualState.output | Should -Be $expected
-    }
+    $out = $config_yaml | dsc config get -f - | ConvertFrom-Json
+    $out.results[0].result.actualState.output | Should -Be $expected
+  }
 
-    It 'path(<path>) works' -TestCases @(
-        @{ path = "systemRoot(), 'a'"; expected = "$PSHOME$([System.IO.Path]::DirectorySeparatorChar)a" }
-        @{ path = "'a', 'b', 'c'"; expected = "a$([System.IO.Path]::DirectorySeparatorChar)b$([System.IO.Path]::DirectorySeparatorChar)c" }
-    ) {
-        param($path, $expected)
+  It 'path(<path>) works' -TestCases @(
+    @{ path = "systemRoot(), 'a'"; expected = "$PSHOME$([System.IO.Path]::DirectorySeparatorChar)a" }
+    @{ path = "'a', 'b', 'c'"; expected = "a$([System.IO.Path]::DirectorySeparatorChar)b$([System.IO.Path]::DirectorySeparatorChar)c" }
+  ) {
+    param($path, $expected)
 
-        $config_yaml = @"
+    $config_yaml = @"
             `$schema: https://aka.ms/dsc/schemas/v3/bundled/config/document.json
             resources:
             - name: Echo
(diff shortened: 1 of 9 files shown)


SteveL-MSFT:
@Gijsreyn can you resolve the merge conflict then I'll review this.  Thanks!

Gijsreyn:
> @Gijsreyn can you resolve the merge conflict then I'll review this. Thanks!

Oops, didn't see it had a merge conflict. Should be resolved now!

dsc_lib/src/functions/array.rs
```diff
@@ -0,0 +1,103 @@
...
+#[derive(Debug, Default)]
+pub struct Array {}
+
+impl Function for Array {
+    fn get_metadata(&self) -> FunctionMetadata {
+        FunctionMetadata {
+            name: "array".to_string(),
+            description: t!("functions.array.description").to_string(),
+            category: FunctionCategory::Array,
+            min_args: 0,
```
SteveL-MSFT: The docs https://learn.microsoft.com/en-us/azure/azure-resource-manager/templates/template-functions-array#array indicate that it only takes 1 argument?

dsc_lib/src/functions/first.rs
```diff
@@ -0,0 +1,116 @@
...
+#[cfg(test)]
+mod tests {
+    use crate::configure::context::Context;
+    use crate::parser::Statement;
+
+    #[test]
+    fn array_of_strings() {
+        let mut parser = Statement::new().unwrap();
+        let result = parser.parse_and_execute("[first(createArray('hello', 'world'))]", &Context::new()).unwrap();
+        assert_eq!(result.to_string(), "\"hello\"");
```
SteveL-MSFT: I would expect the result is a string without quotes

dsc_lib/src/functions/index_of.rs
```diff
@@ -0,0 +1,130 @@
...
+            return_types: vec![FunctionArgKind::Number],
+        }
+    }
+
+    fn invoke(&self, args: &[Value], _context: &Context) -> Result<Value, DscError> {
+        debug!("{}", t!("functions.indexOf.invoked"));
+        
+        let array = args[0].as_array().ok_or_else(|| {
+            DscError::Parser(t!("functions.indexOf.invalidArrayArg").to_string())
+        })?;
```
SteveL-MSFT: Prefer this syntax:

```suggestion
        let Some(array) = args[0].as_array() else {
            return DscError::Parser(t!("functions.indexOf.invalidArrayArg").to_string())
        };
```

dsc_lib/src/functions/index_of.rs
```diff
@@ -0,0 +1,130 @@
...
+
+        for (index, item) in array.iter().enumerate() {
+            let matches = match (item_to_find, item) {
+                // String comparison (case-sensitive)
+                (Value::String(find_str), Value::String(item_str)) => find_str == item_str,
+                (Value::Number(find_num), Value::Number(item_num)) => find_num == item_num,
+                (Value::Array(find_arr), Value::Array(item_arr)) => find_arr == item_arr,
+                (Value::Object(find_obj), Value::Object(item_obj)) => find_obj == item_obj,
+                _ => false,
+            };
```
SteveL-MSFT: Don't you just need to check if `item_to_find == item` and return the current index?

tree-sitter-dscexpression/.gitignore
```diff
@@ -1,27 +1,25 @@
 # Rust artifacts
 target/
+Cargo.lock
```
SteveL-MSFT: this shouldn't be in ignore

dsc_lib/src/functions/index_of.rs
Gijsreyn: You're right. Modified!

dsc_lib/src/functions/array.rs
```diff
@@ -0,0 +1,103 @@
...
+pub struct Array {}
+
+impl Function for Array {
+    fn get_metadata(&self) -> FunctionMetadata {
+        FunctionMetadata {
+            name: "array".to_string(),
+            description: t!("functions.array.description").to_string(),
+            category: FunctionCategory::Array,
+            min_args: 1,
+            max_args: usize::MAX,
```
SteveL-MSFT: This is still not correct, it appears that `array()` is solely like casting a single value into an array of 1 element.  However, here you're duplicating `createArray()`

dsc/tests/dsc_functions.tests.ps1
```diff
@@ -274,70 +274,138 @@ Describe 'tests for function expressions' {
...
+            - name: Echo
+              type: Microsoft.DSC.Debug/Echo
+              properties:
+                output: "$expression"
+"@
+    $out = dsc -l trace config get -i $config_yaml 2>$TestDrive/error.log | ConvertFrom-Json
+    $LASTEXITCODE | Should -Be 0 -Because (Get-Content $TestDrive/error.log -Raw)
+    ($out.results[0].result.actualState.output | Out-String) | Should -BeExactly ($expected | Out-String)
+  }
+}
```
SteveL-MSFT: add newline at the end

dsc_lib/locales/en-us.toml
```diff
@@ -217,6 +217,11 @@ invoked = "add function"
 description = "Evaluates if all arguments are true"
 invoked = "and function"
 
+[functions.array]
+description = "Creates an array from the given elements of mixed types"
```
SteveL-MSFT:
```suggestion
description = "Convert the value to an array"
```

dsc_lib/src/functions/first.rs
```diff
@@ -0,0 +1,116 @@
...
+                return Err(DscError::Parser(t!("functions.first.emptyArray").to_string()));
+            }
+            return Ok(array[0].clone());
+        }
+
+        if let Some(string) = args[0].as_str() {
+            if string.is_empty() {
+                return Err(DscError::Parser(t!("functions.first.emptyString").to_string()));
+            }
+            return Ok(Value::String(string.chars().next().unwrap().to_string()));
```
SteveL-MSFT, line 42: This seems unnecessary, wouldn't `Value::String(string.to_string())` be sufficient?

dsc_lib/src/functions/index_of.rs
```diff
@@ -0,0 +1,121 @@
...
+#[cfg(test)]
+mod tests {
+    use crate::configure::context::Context;
+    use crate::parser::Statement;
+
+    #[test]
+    fn find_string_in_array() {
+        let mut parser = Statement::new().unwrap();
+        let result = parser.parse_and_execute("[indexOf(createArray('apple', 'banana', 'cherry'), 'banana')]", &Context::new()).unwrap();
+        assert_eq!(result.to_string(), "1");
```
SteveL-MSFT: Since the result should be a number, you shouldn't convert to string as it might hide an issue:
```suggestion
        assert_eq!(result, 1);
```

dsc_lib/src/functions/array.rs
```diff
@@ -0,0 +1,103 @@
...
+            return_types: vec![FunctionArgKind::Array],
+        }
+    }
+
+    fn invoke(&self, args: &[Value], _context: &Context) -> Result<Value, DscError> {
+        debug!("{}", t!("functions.array.invoked"));
+        
+        let value = &args[0];
+        
+        if value.is_number() || value.is_string() || value.is_array() || value.is_object() {
```
SteveL-MSFT: You don't need to check the type since the engine will already do that, you can just return it as an array.

dsc/tests/dsc_functions.tests.ps1
```diff
@@ -274,70 +274,137 @@ Describe 'tests for function expressions' {
...
+                output: "$expression"
+"@
+    $out = dsc -l trace config get -i $config_yaml 2>$TestDrive/error.log | ConvertFrom-Json
+    $LASTEXITCODE | Should -Be 0 -Because (Get-Content $TestDrive/error.log -Raw)
+    ($out.results[0].result.actualState.output | Out-String) | Should -BeExactly ($expected | Out-String)
+  }
+
+  It 'indexOf function works for: <expression>' -TestCases @(
+    @{ expression = "[indexOf(createArray('apple', 'banana', 'cherry'), 'banana')]"; expected = 1 }
+    @{ expression = "[indexOf(createArray(10, 20, 30), 20)]"; expected = 1 }
```
SteveL-MSFT, line 390: Should have one of the tests like this one check for the last item

SteveL-MSFT (changes requested):

Gijsreyn:
Just patched up your comments @SteveL-MSFT, but looks like there is an authentication issue.

SteveL-MSFT:
@Gijsreyn not an authn issue directly, but there's a new version of `serde_json` that isn't in our feed, so I just need to add it

SteveL-MSFT (approved):

PR #1044, opened 2025-08-11
Add new conceptual docs for configuring DSC settings

Gijsreyn:
# PR Summary

This PR adds a new conceptual article on how to configure DSC settings.


michaeltlombardi:
I think we may want to defer this until we have JSON Schemas we can reference - there's a lot of caveats, currently, and the meta configuration isn't schematized or obvious. Primarily, I'm concerned that documenting the current implementation without the schemas will lead to more adoption, which we'll have to help folks migrate around once we fix the way we do meta configuration.

For some more context, see #1081.

PR #1048, opened 2025-08-12
(REPRO) Trace messaging issues repro

michaeltlombardi:
# PR Summary

This change creates a repro resource, tests, and example configuration to enumerate issues with the current model for bubbling up trace messages from resources.

The `resources/repro.trace` folder contains the relevant files:

- `repro.trace.dsc.resource.json` - resource manifest
- `repro.trace.resource.ps1` - resource script
- `repro.trace.tests.ps1` - failing tests for behavior
- `repro.config.dsc.yaml` - config file with every repro case
- `repro.config.result.yaml` - snapshot of output from config

The repro covers several test cases to show how DSC behaves when a resource emits trace messages as compressed JSON to stderr:

- `simpleMessaging` shows the behavior for emitting a string for a given trace level, like `{"warn":"my warning"}`.
- `minimalStruct` shows the behavior for emitting a struct that defines the `level`, `timestamp`, and `fields.message` properties.
- `structWithMetadata` also defines the `target` and `lineNumber` properties to bubble up for the resource.
- `structWithAdditionalFields` defines fields.`extraString` and `fields`extraInteger` to bubble up related context.
- `structWithMetadataAndAdditionalFields` includes defines `target`, `lineNumber`, `fields.extraString`, and `fields.extraInteger`.

The resource itself returns the name of the test case and uses the `emittedData` read-only property to return the data that it emitted to stderr for DSC to interpret as a trace message.

Findings:

- For the structured representation, DSC only retains the trace level and message contents. The `timestamp`, `target`, `lineNumber`, and additional fields are all ignored or dropped.
- Regardless of emitted JSON, DSC munges the message to insert the process ID like `PID 1234: <message>`.
- For structured representation, DSC bubbles up the message in the `trace_message` field instead of `message`.
- For `dsc config get`, DSC doesn't seem to capture the emitted messages in the result object, only emitting them to stderr.

This repro shows that we have some unexpected behaviors for resource messaging that may guide a review and update of the implementation.

## PR Context

While working on the resource authoring design guide, I started to write up how and when a resource author should implement trace messaging for the resource. To accurately advise authors, I needed a better understanding of how to implement tracing and how it would behave when invoked through DSC.

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -324,7 +324,8 @@ if (!$SkipBuild) {
         "sshdconfig",
         "tools/dsctest",
         "tools/test_group_resource",
-        "y2j"
+        "y2j",
+        "resources/repro.trace"
     )
     $pedantic_unclean_projects = @()
     $clippy_unclean_projects = @("tree-sitter-dscexpression", "tree-sitter-ssh-server-config")
--- a/resources/repro.trace/copy_files.txt
+++ b/resources/repro.trace/copy_files.txt
@@ -1,0 +1,1 @@
+./repro.trace.resource.ps1
--- a/resources/repro.trace/repro.config.dsc.yaml
+++ b/resources/repro.trace/repro.config.dsc.yaml
@@ -1,0 +1,51 @@
+$schema: https://aka.ms/dsc/schemas/v3/bundled/config/document.vscode.json
+resources:
+  - type: Dsc.Repro/Trace
+    name: Simple messaging
+    properties:
+      case: simpleMessage
+      # emittedData: # Output from repro testing
+      #   warn: Simple message
+  - type: Dsc.Repro/Trace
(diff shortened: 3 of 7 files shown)


SteveL-MSFT:
Discussed with @michaeltlombardi, conclusions:

- Use the extended macro syntax to pass along the additional properties
- Put the PID in a custom field
- instead of passing `trace_message` variable directly to macro, need to specify `trace` field: https://github.com/PowerShell/DSC/blob/main/dsc_lib/src/dscresources/command_resource.rs#L860
- ResourceMessage is not being used so is always empty, need to collect the traces and put it in https://github.com/PowerShell/DSC/blob/0577c7e5c5bcc680f926544befa0bc08abe5d518/dsc_lib/src/configure/config_result.rs#L19
-

PR #1057, opened 2025-08-18
Add contribution workflow and synopsis

Gijsreyn:
# PR Summary

For newcomers, it's helpful to understand how they can contribute and run the `build.ps1` script. This pull request adds a new section to the `CONTRIBUTING.md` with additional steps. Also added a SYNOPSIS in the build script with examples + removed the `-Verbose` parameter and added the `[CmdletBinding()]` attribute


michaeltlombardi:
Thinking about this a bit more in depth, we probably want to use this information for a contributor's guide - which would require placing the majority of the information in Markdown files in the `docs` folder - likely `docs/contributing`.

Then we could independently add documents for testing, building from source, and so on. I don't consider this a blocking concern for this PR, just something to be mindful of.

kilasuit:
@michaeltlombardi I like your thinking as this aligns with how the layout is in https://github.com/PowerShell/PowerShell & makes it easier to maintain docs that may be useful in other repos across across the PowerShell Org (or additional ones too) from a centralised repo in future

Kinda annoying though that as part of common repo layouts everything seems to just be `thrown on the floor` directly at the repo root, like a messy bedroom 😂 & that UI links to certain docs are hardcoded (atm)

@Gijsreyn I'd recommend as Mikey suggests and follow the PowerShell repo layout. Especially as we can then make use of features like CODEOWNERS for that whole directory easier and add an [issue-template for repo documentation updates like this PR should add to the PowerShell Repo](https://github.com/PowerShell/PowerShell/pull/25601)

PR #1059, opened 2025-08-18
Update the release pipeline

SteveL-MSFT:
# PR Summary

- Comment out aarch linker for now to see if resulting binaries still work since the linker isn't found now
- Enable Release pipeline to work for Official and NonOfficial runs via a parameter
- Change URL for RustSDK download to new feed within MSAzure
- Update subscription being used
- Use variable for feed URL
- Update build script to install tree-sitter CLI from main script
- Add condition in build script to not run msrustup commands that update environment since they aren't needed in ADO
- Add additional tracing and checks if native commands failed
- Update cargo install command to explicitly use redirected feed from config.toml

### DIFF
--- a/.pipelines/DSC-Official.yml
+++ b/.pipelines/DSC-Official.yml
@@ -179,7 +179,7 @@ extends:
           condition: succeeded()
 
       - job: SignMsixBundle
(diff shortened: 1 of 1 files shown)


SteveL-MSFT:
Reviewed and tested internally

PR #1069, opened 2025-08-20
pass input metadata to resource

tgauth:
# PR Summary
- extract existing schema validation logic in `subcommand.rs` into helper function to call from there and `add_metadata()`
- modify `add_metadata()` to:
  - insert `_metadata` key into `properties` for resources that permit it in their schema
  - extend existing `metadata` key for adapters
- add related pester tests

## PR Context
- any input metadata for a resource should be passed to the resource invocation, if resources permit it in their schema


tgauth:
@SteveL-MSFT - looks like the key name is `metadata` for adapters, but the test resource uses `_metadata`. Should I standardize to one name?

dsc_lib/src/configure/mod.rs
```diff
@@ -185,6 +190,18 @@ fn add_metadata(kind: &Kind, mut properties: Option<Map<String, Value>> ) -> Res
         return Ok(serde_json::to_string(&properties)?);
     }
 
+    if let Some(resource_metadata) = resource_metadata {
+        let other_metadata = resource_metadata.other;
+        if let Some(mut properties) = properties {
+            properties.insert("_metadata".to_string(), Value::Object(other_metadata));
```
SteveL-MSFT: I think we need to check if the resource has a `_metadata` property before we add it otherwise it will invalidate their schema.  There should be a test added for this as well.  In the case the resource doesn't handle `_metadata`, then we don't send it.

SteveL-MSFT (approved):

Gijsreyn:
@michaeltlombardi should this have a doc-impact label and requires to be documented for resource authors?

PR #1071, opened 2025-08-21
Experimental PowerShell discover extension

### LINKED ISSUE #913: Discover extension for PSModulePath

It would make sense for some PS Modules to include extensions or even resources.  `SecretManagement` makes sense to have a secret extension and `AzPS` also (for AzKeyVault).

Add a new discover extension that searches PSModulePath for resources and extensions.  Need to have some caching so that it doesn't crawl all the paths each time.  It may be more efficient to also require the module manifest to declare it.

Gijsreyn:
# PR Summary

Implements an optimized approach to discovering PowerShell resources using .NET with parallel processing and direct result synchronization. 

The previous approach using Get-ChildItem had unnecessary object overhead and required post-processing with `Where-Object`. The same is implied when using a thread-safe collection (ConcurrentBag), which adds unnecessary complexity.

## PR Context

Fix #913

> [!NOTE]
> While I recognize the need for caching in the issue, I left it out for now, assuming we can increment it over time.


Gijsreyn:
With PR #1025 around, it might be worthwhile to rename the directories.

Gijsreyn:
> A few questions and requests:
> 
> 1. Do I correctly understand that this extension _only_ for resources implemented in PowerShell and excluding Windows PowerShell? If so, do we need a second extension for discovering those resources, or can we handle that in this one? I'm not sure I see an immediate reason for why we can't handle both here, given the manifest has to indicate whether the resource is invoked with `powershell` or `pwsh`.
> 2. Can we slightly restructure this script to define (currently empty) `param` block and put the implementation into the `process` block? That can help with organization and testing later on.
> 3. The current implementation _can't_ run in Windows PowerShell, I think, given the use of `ForEach-Object -Parallel` - but we can still _discover_ resources in Windows PowerShell modules.

1. The one currently written is only compatible with PowerShell 7+ because of the `[System.IO.EnumerationOptions]`. We might be able to rewrite it, but as you also mentioned (from point 3), the `Foreach-Object -Parallel` is not implemented. It would be best to create a second extension for it with different logic.
2. Sure, I'll implement this change. Thanks for the comment!
3. Answered in point 1.

extensions/powershell/powershell.discover.ps1
```diff
@@ -0,0 +1,25 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+[CmdletBinding()]
+param ()
+
+begin {
+    $psPaths = $env:PSModulePath -split [System.IO.Path]::PathSeparator | Where-Object { $_ -notmatch 'WindowsPowerShell' }
```
michaeltlombardi: Are we throwing away paths with `WindowsPowerShell` in them for performance reasons or because we don't want to support discovering resources shipped in Windows PowerShell modules?

If the latter, I think we _might_ be better off retrieving the folder paths for discovered modules and searching those.

extensions/powershell/powershell.discover.ps1
```diff
@@ -0,0 +1,25 @@
...
+                [System.IO.Directory]::EnumerateFiles($_, $pattern, $enumOptions) | ForEach-Object {
+                    @{ manifestPath = $_ }
+                }
+            } catch { }
+        }
+    } -ThrottleLimit 10
+}
+end {
+    $manifests | ForEach-Object { $_ | ConvertTo-Json -Compress }
+}
```
michaeltlombardi: nit: missing trailing newline.
```suggestion
}

```

extensions/powershell/powershell.discover.ps1
Gijsreyn: Quite honestly, the only reason that I'm filtering it out is because if I can remember correctly, there was a bug in the past where we filtered out `WindowsPowerShell` in the adapter (or tests). Looking through the codebase, I can see this was the change: https://github.com/PowerShell/DSC/pull/777/files

If we opt for searching folder paths, I suppose it needs to match the specific operating systems, which becomes a _ugly_ `if/elseif/else` block. It also limits custom installation, which will be lost. For example:

image

SteveL-MSFT:
The WG discussed this and agree to not have WinPS support initially until there is a customer need

Gijsreyn:
@SteveL-MSFT - by any chance, if you have time to review this one, I would appreciate it.

extensions/powershell/powershell.discover.ps1
```diff
@@ -0,0 +1,25 @@
...
+[CmdletBinding()]
+param ()
+
+begin {
+    $psPaths = $env:PSModulePath -split [System.IO.Path]::PathSeparator | Where-Object { $_ -notmatch 'WindowsPowerShell' }
+}
+process {
+    $manifests = $psPaths | ForEach-Object -Parallel {
+        $searchPatterns = @('*.dsc.resource.json', '*.dsc.resource.yaml', '*.dsc.resource.yml')
+        $enumOptions = [System.IO.EnumerationOptions]@{ IgnoreInaccessible = $false; RecurseSubdirectories = $true }
```
SteveL-MSFT: I'm concerned about the perf impact.  Can we have some caching that stores the last modified date of a PSModulePath segment and only search it if it has changed from last time?  This also requires currently discovered ones to be persisted on disk.

SteveL-MSFT (changes requested):

extensions/powershell/powershell.discover.ps1
Gijsreyn: Added caching.

Gijsreyn:
@SteveL-MSFT - caching added. Took a couple of rounds because of _returning_ nothing. Added this to the docs after discussing with Mikey.

extensions/powershell/powershell.discover.tests.ps1
```diff
@@ -0,0 +1,125 @@
...
+    
+    $cacheFilePath = if ($IsWindows) {
+        Join-Path $env:LocalAppData "dsc\PowerShellDiscoverCache.json"
+    } else {
+        Join-Path $env:HOME ".dsc" "PowerShellDiscoverCache.json"
+    }
+    $script:cacheFilePath = $cacheFilePath
+
+    Remove-Item -Force -ErrorAction SilentlyContinue -Path $script:cacheFilePath
+}
```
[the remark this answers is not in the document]
Gijsreyn: Other tests do the same.

docs/reference/schemas/extension/stdout/discover.md
```diff
@@ -22,13 +22,16 @@ Type:          object
 
 ## Description
 
-Represents the actual state of a resource instance in DSCpath to a discovered DSC resource or
+Represents the actual state of a resource instance in DSC path to a discovered DSC resource or
 extension manifest on the system. DSC expects every JSON Line emitted to stdout for the
 **Discover** operation to adhere to this schema.
```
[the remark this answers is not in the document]
Gijsreyn: Earlier suggstion already taken from Mikey.

extensions/powershell/powershell.discover.ps1
```diff
@@ -0,0 +1,113 @@
...
+                return $false
+            }
+            
+            $currentLastWrite = (Get-Item $path).LastWriteTimeUtc
+            $cachedLastWrite = [DateTime]$entry.Value
+            
+            if ($currentLastWrite -ne $cachedLastWrite) {
+                return $false
+            }
+        }
```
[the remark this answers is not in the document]
Gijsreyn: Fixed.

Gijsreyn:
@SteveL-MSFT - if you've time, do you mind taking a stab at this one? Would love to see this get in GA 3.2! :)

docs/reference/schemas/extension/stdout/discover.md
```diff
@@ -22,13 +22,16 @@ Type:          object
...
 
 The output must be a JSON object. The object must define the full path to the discovered manifest.
 If an extension returns JSON that is invalid against this schema, DSC raises an error.
 
+If the extension doesn't discover any manifests, it must return nothing to stdout. An empty output
+indicates no resources were found.
```
SteveL-MSFT: Should add that if nothing is found, a zero exit code should be returned indicating success unless that is already covered elsewhere

dsc/tests/dsc_extension_discover.tests.ps1
```diff
@@ -148,4 +156,4 @@ Describe 'Discover extension tests' {
             $env:DSC_RESOURCE_PATH = $null
         }
     }
-}
+}
```
SteveL-MSFT: Please add newline at end

extensions/powershell/powershell.discover.ps1
```diff
@@ -0,0 +1,122 @@
...
+    }
+    end {
+        if ($manifests.Count -gt 0) {
+            $manifests | ForEach-Object { $_ | ConvertTo-Json -Compress }
+        }
+    }
+}
+
+Invoke-DscResourceDiscovery
```
SteveL-MSFT: Add newline to end

docs/reference/schemas/extension/stdout/discover.md
Gijsreyn: Added additional section on exit codes.

Issue #543, opened 2024-09-09
Support version pinning in a configuration document


michaeltlombardi:
### Summary of the new feature / enhancement

> As an infrastructure engineer, I want to be able to define version pins for the resoures in my configuration documents to ensure my configuration management is as deterministic as possible. 

### Proposed technical implementation details (optional)

A few options come to mind.

The easiest-for-users would be to allow them to append a version pin to the end of the resource type name for a resource, like `Microsoft.DSC/PowerShell@v1.0.0`. This would require introduce complexity to the engine for handling and impact the way configuration functions work, which makes it a harder sell to me.

The more explicit option is to support a `version` field on the resource instance declaration:

```yaml
- type:    Microsoft.DSC/PowerShell
  version: v1.0.0
```

Another option would be to define it under resource metadata:

```yaml
- type: Microsoft.DSC/PowerShell
  metadata:
    Microsoft.DSC:
      version: v1.0.0
```

Of the available options, I lean more towards the metadata implementation, even though it requires defining per-resource metadata. However, we've discussed other uses for this property bag, and defining a structure for this would pave the way for those use cases. It would also enable users to declare version pins in the top-level metadata with per-resource overrides (if we choose to support that).

I don't think this would require any further work for adapters, since they already report available resources and versions to DSC.

anmenaga:
Currently discovery in the engine and resource invocation (i.e. `find and invoke this resource with this specific version` scenario) does **not** consider resource versions at all;
so either way this will need some work in the engine.
And, most likely, an internal resource identifier will be changed to look like `Microsoft.DSC/PowerShell@v1.0.0`,
so probably we should use the same for user-facing resource identifier.

michaeltlombardi:
Arguably, it might be better to have a `pin` or `require` (or similar) field in the resource instance declaration or metadata to support pinning by version/sha/signature/etc as that metadata about resources works its way into the engine so we don't have to overload the type name further. I'm also not sure if `<namespace>/<name>@<version>` is valid ARM syntax or not.

```yaml
metadata:
  Microsoft.DSC:
    requireResources:
      - type: MyOrg/Database
        version: 1.2
        sha: <sha>
      - type: MyOrg.Web/*
        version: "[2.0,3.0)"
        signed_by: <trusted-persona>
resources:
  # Using `require` field on instance
  - type: MyOrg.Web/Server
    name: Setup web server
    require:
      version: 2.4.2
    properties: {}
  # Using `require` field in metadata
[...]

PR #1077, opened 2025-08-24
Enable specifying version for resource

SteveL-MSFT:
# PR Summary

Allows specifying semver requirements via the `apiVersion` property.  Uses the https://docs.rs/semver/latest/semver/struct.VersionReq.html syntax, some examples:

- `1.1`
- `1.1.*`
- `*`
- `>1.1.2`
- `<=1.3.2`
- `>1.1.2, <2.0.0`
- `~1.1`
- `^1.1`
- `=1.1`

Note that if the version requirement doesn't start with a number, then an `=` operator is prefixed indicating exact version.  Also, adapted resource versioning is NOT part of this PR and will come separately since this PR is already big enough.

`dsc resource list` will now show multiple versions of a resource and `dsc resource` has a new `--version` parameter.

## PR Context

Fix https://github.com/PowerShell/DSC/issues/543


dsc_lib/src/discovery/discovery_trait.rs
```diff
@@ -1,17 +1,43 @@
...
 
 #[derive(Debug, PartialEq)]
 pub enum DiscoveryKind {
     Resource,
     Extension,
 }
 
+#[derive(Debug, Clone, Eq, Hash, PartialEq)]
+pub struct DiscoveryFilter {
+    resource_type: String,
```
michaeltlombardi: Should this be `r#type` to support (eventual) filtering for extensions by version during discovery?

SteveL-MSFT: That makes sense, wasn't thinking about extensions

dsc/tests/dsc_discovery.tests.ps1
```diff
@@ -168,19 +166,19 @@ Describe 'tests for resource discovery' {
 
     It 'Verify non-zero exit code when resource not found' {
 
-        $out = dsc resource get -r abc/def
+        $out = dsc resource get -r abc/def 2>$null
         $LASTEXITCODE | Should -Be 7
-        $out = dsc resource get --all -r abc/def
+        $out = dsc resource get --all -r abc/def >$null
```
tgauth:
```suggestion
        $out = dsc resource get --all -r abc/def 2>$null
```

tgauth (approved):
LGTM

Issue #1082, opened 2025-08-27
Author secret extensions


Gijsreyn:
### Summary of the new feature / enhancement

As an extension author for secrets, I want to get Intellisense when authoring secret extensions.

Gijsreyn:
Resolved in #1083.

PR #1083, opened 2025-08-27
(SCHEMA) Add extension secret to schema source

Gijsreyn:
# PR Summary

This changeset only adds the extension secret schema.

## PR Context

Partially addresses #1082

### DIFF
--- a/schemas/src/extension/manifest.secret.yaml
+++ b/schemas/src/extension/manifest.secret.yaml
@@ -1,0 +1,105 @@
+# yaml-language-server: $schema=https://json-schema.org/draft/2020-12/schema
+$schema: https://json-schema.org/draft/2020-12/schema
+$id:     <HOST>/<PREFIX>/<VERSION>/extension/manifest.secret.yaml
+
+title: Secret operation command
+description: >-
+  Defines how DSC must call the DSC extension to retrieve a secret value.
+markdownDescription: | # VS Code only
+  ***
+  [_Online Documentation_][00]
+  ***
+
+  Defines how DSC must call the DSC extension to retrieve a secret value. An
+  extension that defines this field in its manifest has the `secret` capability.
+
+  The secret operation is expected to output a single line to stdout containing
+  the secret text. If the extension outputs no data to stdout, DSC treats it as
+  "no secret returned". If the extension outputs multiple lines, DSC treats that
+  as an error.
+
+  For details about the output format, see
+  [Secret extension operation stdout][01].
+
+  [00]: <DOCS_BASE_URL>/reference/schemas/extension/manifest/secret?<DOCS_VERSION_PIN>
+  [01]: <DOCS_BASE_URL>/reference/schemas/extension/stdout/secret?<DOCS_VERSION_PIN>
+
+type: object
+required:
+  - executable
+properties:
+  executable:
+    $ref: /<PREFIX>/<VERSION>/definitions/commandExecutable.yaml
+    markdownDescription: |
+      ***
+      [_Online Documentation_][01]
+      ***
+
+      Defines the name of the command to run. The value must be the name of a command discoverable
+      in the system's `PATH` environment variable or the full path to the command. A file extension
+      is only required when the command isn't recognizable by the operating system as an
+      executable.
+
+      [01]: <DOCS_BASE_URL>/reference/schemas/extension/manifest/secret?<DOCS_VERSION_PIN>#executable
+  args:
+    title: Arguments
+    description: >-
+      Defines an ordered list of arguments to pass to the command.
+    markdownDescription: |
+      ***
+      [_Online Documentation_][01]
+      ***
+
+      Defines an ordered list of arguments to pass to the command. Items can be plain strings or
+      structured entries indicating which argument name should receive the secret name or the vault
+      name.
+
+      DSC expands structured entries as follows:
+
+      - `{ "nameArg": "<flag>" }` expands to `"<flag>", "<secret-name>"`.
+      - `{ "vaultArg": "<flag>" }` expands to `"<flag>", "<vault-name>"` (only when a vault is
+        specified by the caller).
+
+      [01]: <DOCS_BASE_URL>/reference/schemas/extension/manifest/secret?<DOCS_VERSION_PIN>#args
+    type: array
+    items:
+      anyOf:
+        - type: string
+        - type: object
+          additionalProperties: false
+          required:
+            - nameArg
+          properties:
+            nameArg:
+              type: string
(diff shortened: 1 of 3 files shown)


michaeltlombardi (dismissed):

michaeltlombardi:
@SteveL-MSFT - this looks correct to me from a docs perspective, just want the technical details double-checked.

schemas/src/extension/manifest.secret.yaml
```diff
@@ -0,0 +1,105 @@
...
+  [_Online Documentation_][00]
+  ***
+
+  Defines how DSC must call the DSC extension to retrieve a secret value. An
+  extension that defines this field in its manifest has the `secret` capability.
+
+  The secret operation is expected to output a single line to stdout containing
+  the secret text. If the extension outputs no data to stdout, DSC treats it as
+  "no secret returned". If the extension outputs multiple lines, DSC treats that
+  as an error.
```
SteveL-MSFT: We should probably expand on this a bit more:

- Extensions MUST return the secret as a single line in plaintext with a zero exit code indicating success
- If the secret does NOT exist, then the extension MUST return no output AND a zero exit code indicating no failure of the extension
- If the extension cannot query/retrieve the secret (for example, needs the user to authenticate first), then the extension SHOULD return a detailed error message as a JSONLine to STDERR and MUST have a non-zero exit code

SteveL-MSFT: @michaeltlombardi can help with the actual proposed text

schemas/src/extension/manifest.secret.yaml
```diff
@@ -0,0 +1,105 @@
...
+      [_Online Documentation_][01]
+      ***
+
+      Defines an ordered list of arguments to pass to the command. Items can be plain strings or
+      structured entries indicating which argument name should receive the secret name or the vault
+      name.
+
+      DSC expands structured entries as follows:
+
+      - `{ "nameArg": "<flag>" }` expands to `"<flag>", "<secret-name>"`.
```
SteveL-MSFT: @michaeltlombardi do we specify in other cases that the `<flag>` or parameter name MUST include any dashes, etc... as it's used verbatim?

schemas/src/extension/manifest.secret.yaml
```diff
@@ -0,0 +1,105 @@
...
+  The secret operation is expected to output a single line to stdout containing
+  the secret text. If the extension outputs no data to stdout, DSC treats it as
+  "no secret returned". If the extension outputs multiple lines, DSC treats that
+  as an error.
+
+  For details about the output format, see
+  [Secret extension operation stdout][01].
+
+  [00]: <DOCS_BASE_URL>/reference/schemas/extension/manifest/secret?<DOCS_VERSION_PIN>
+  [01]: <DOCS_BASE_URL>/reference/schemas/extension/stdout/secret?<DOCS_VERSION_PIN>
```
michaeltlombardi:
```suggestion
  DSC expects extensions implementing the `secret` capability to adhere to the
  following contract:

  1. If the extension retrieves the secret, the extension must emit the secret
     to stdout as a single line of plaintext and exit with code `0`. DSC
     consumes the emitted output and makes the secret available in the
     configuration document.

     If the extension emits more than one line to stdout, DSC raises an error.
  1. If the extension cannot retrieve the secret because the secret doesn't
     exist, the extension must not emit any text to stdout and must exit with
     code `0`. DSC interpets this result as the secret not existing in the
     vault.
  1. If the extension cannot retrieve the secret for any other reason, such
     as invalid credentials or an API error, the extension should emit
     a descriptive error message as a JSON Line to stderr and exit with a
     nonzero exit code. DSC interprets the nonzero exit code as an operational
     failure and surfaces that information and any emitted error messages to
     the user.

  When the exit code for the operation is `0`, DSC interprets the operation as
  completing without errors. For extensions, failure to retrieve a secret
  because it doesn't exist is _not_ an error. Failure to retrieve a secret
  for any other reason _is_ an error and the extension should exit with a
  nonzero code. For an improved user experience, the extension should define
  the `exitCodes` field in the extension manifest to indicate what the nonzero
  exit code means.

  For more information about how DSC validates the data for stdout, see
  [Secret extension operation stdout][01]. For more information about defining
  exit codes for the extension, see [`exitCodes`][02] in the extension manifest
  schema reference.

  [00]: <DOCS_BASE_URL>/reference/schemas/extension/manifest/secret?<DOCS_VERSION_PIN>
  [01]: <DOCS_BASE_URL>/reference/schemas/extension/stdout/secret?<DOCS_VERSION_PIN>
  [02]: <DOCS_BASE_URL>/reference/schemas/extension/manifest/root?<DOCS_VERSION_PIN>#exitcodes
```

schemas/src/extension/manifest.secret.yaml
michaeltlombardi: We don't explicitly say so, but we have examples that show the behavior. For example, in `set.args`, we define the `markdownDescription` with the following content:

https://github.com/PowerShell/DSC/blob/4750f779ecb27eb480f7ab37b8e8a526ed5054f5/schemas/src/resource/manifest.set.yaml#L80-L106

Gijsreyn:
Added your remark Mikey, thanks for the rewritten version :)

michaeltlombardi:
@Gijsreyn - I'm going to adopt this PR to make some changes to the argument definitions. Adding them as review comments will be more confusing/difficult than updating in place, but I'll sync with you to explain these changes.

michaeltlombardi:
This PR should be ready for review again.

schemas/src/extension/stdout/secret.yaml
```diff
@@ -0,0 +1,49 @@
...
+       the secret and noting that the secret doesn't exist. DSC surfaces emitted messages to the
+       caller depending on the message level.
+    1. Exit with code `0` to indicate success. For DSC, an extension not returning a secret value
+       because no secret with the given name exists is _not_ a failure.
+
+  - If the extension can't retrieve the secret for any other reason, such as requiring the vault
+    to be unlocked or encountering an API error, DSC expects the extension to:
+
+    1. Emit _no output_ to stdout.
+    1. Optionally emit detailed error messages to stderr as [JSON Lines][01] to provide contextual
```
Gijsreyn, line 38: Nit: Just thinking if it's helpful for the user to have an example here?

michaeltlombardi: I think we leave it out for now - when we have the docs on emitting messages published, we can link to that.

michaeltlombardi (approved):

PR #1085, opened 2025-08-29
Add skip() function

Gijsreyn:
# PR Summary

This pull request adds the `skip()` function, including reference documentation.

## PR Context

Partially addresses https://github.com/PowerShell/DSC/issues/57


docs/reference/schemas/config/functions/skip.md
```diff
@@ -0,0 +1,147 @@
...
+## Description
+
+The `skip()` function returns the tail of an array or string by skipping a
+specified number of items from the start.
+
+- For arrays: returns a new array containing elements after the specified index
+- For strings: returns a new string containing characters after the specified index
+
+Both parameters are required. `originalValue` must be an array or a string.
+`numberToSkip` must be an integer greater than or equal to zero. If the number
```
SteveL-MSFT: The ARM docs https://learn.microsoft.com/en-us/azure/azure-resource-manager/templates/template-functions-array#skip indicate this could be negative, but treated the same as zero.  This could be useful with other functions that might return a negative value.

dsc/tests/dsc_functions.tests.ps1
```diff
@@ -408,4 +408,27 @@ Describe 'tests for function expressions' {
     $LASTEXITCODE | Should -Be 0 -Because (Get-Content $TestDrive/error.log -Raw)
     ($out.results[0].result.actualState.output | Out-String) | Should -BeExactly ($expected | Out-String)
   }
+
+  It 'skip function works for: <expression>' -TestCases @(
+    @{ expression = "[skip(createArray('a','b','c','d'), 2)]"; expected = @('c','d') }
+    @{ expression = "[skip('hello', 2)]"; expected = 'llo' }
+    @{ expression = "[skip(createArray('a'), 0)]"; expected = @('a') }
```
SteveL-MSFT: Would be better to have more than one element in the array, two is sufficient

dsc/tests/dsc_functions.tests.ps1
```diff
@@ -408,4 +408,27 @@ Describe 'tests for function expressions' {
...
+    @{ expression = "[skip('a', 0)]"; expected = 'a' }
```
SteveL-MSFT: Same here, have more than 1 character

dsc/tests/dsc_functions.tests.ps1
```diff
@@ -408,4 +408,27 @@ Describe 'tests for function expressions' {
...
+    @{ expression = "[skip(createArray('a','b'), 5)]"; expected = @() }
+    @{ expression = "[skip('', 1)]"; expected = '' }
```
SteveL-MSFT, line 418: This test is fine, but would also like to see `[skip('abc', 5)]` and `[skip(createArray('a','b'), 5)]`

dsc_lib/src/functions/skip.rs
```diff
@@ -0,0 +1,84 @@
...
+
+    fn invoke(&self, args: &[Value], _context: &Context) -> Result<Value, DscError> {
+        debug!("{}", t!("functions.skip.invoked"));
+
+        let Some(count_i64) = args[1].as_i64() else {
+            return Err(DscError::Parser(t!("functions.skip.invalidNumberToSkip").to_string()));
+        };
+        if count_i64 < 0 {
+            return Err(DscError::Parser(t!("functions.skip.negativeNotAllowed").to_string()));
+        }
```
SteveL-MSFT:
```suggestion
        if count_i64 < 0 {
            count_i64 = 0;
        }
```

`count_i64` would need to be `mut` for this to work

dsc_lib/src/functions/skip.rs
```diff
@@ -0,0 +1,84 @@
...
+            ],
+            remaining_arg_accepted_types: None,
+            return_types: vec![FunctionArgKind::Array, FunctionArgKind::String],
+        }
+    }
+
+    fn invoke(&self, args: &[Value], _context: &Context) -> Result<Value, DscError> {
+        debug!("{}", t!("functions.skip.invoked"));
+
+        let Some(count_i64) = args[1].as_i64() else {
```
SteveL-MSFT: The general parameter validation will ensure args[1] is a i64, so you could just wrap the whole thing as:

```
if let Some(count_i64) = args[1].as_i64() {
...
}
```

dsc_lib/src/functions/skip.rs
```diff
@@ -0,0 +1,84 @@
...
+        let count: usize = count_i64.try_into().unwrap_or(usize::MAX);
+
+        if let Some(array) = args[0].as_array() {
+            if count >= array.len() { return Ok(Value::Array(vec![])); }
+            let skipped = array.iter().skip(count).cloned().collect::<Vec<Value>>();
+            return Ok(Value::Array(skipped));
+        }
+
+        if let Some(s) = args[0].as_str() {
+
```
SteveL-MSFT, line 51:
```suggestion
```

SteveL-MSFT (changes requested):

dsc_lib/src/functions/skip.rs
Gijsreyn: Slightly changed as clap was complaining on this change.

dsc/tests/dsc_functions.tests.ps1
Gijsreyn: Added additional tests, also for negative integers. Can you please check again?

SteveL-MSFT:
@Gijsreyn can you fix the merge conflict

PR #1092, opened 2025-09-03
Enable initial DSC MCP Server support

SteveL-MSFT:
# PR Summary

Enable DSC to be MCP Server.  First tool is `list_resources`

### DIFF
--- a/dsc/src/mcp/list_resources.rs
+++ b/dsc/src/mcp/list_resources.rs
@@ -43,7 +43,7 @@ impl McpServer {
         let result = task::spawn_blocking(move || {
             let mut dsc = DscManager::new();
             let mut resources = Vec::new();
-            for resource in dsc.list_available(&DiscoveryKind::Resource, "*", "*", ProgressFormat::None) {
+            for resource in dsc.list_available(&DiscoveryKind::Resource, "*", "", ProgressFormat::None) {
                 resources.push(resource);
             }
             ResourceListResult { resources }
--- a/dsc/src/mcp/list_resources.rs
+++ b/dsc/src/mcp/list_resources.rs
@@ -1,0 +1,54 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+use crate::mcp::McpServer;
+use dsc_lib::{
+    DscManager,
+    discovery::{
+        command_discovery::ImportedManifest,
+        discovery_trait::DiscoveryKind,
+    },
+    progress::ProgressFormat,
+};
+use rmcp::{ErrorData as McpError, Json, tool, tool_router};
+use schemars::JsonSchema;
+use serde::{Serialize, Deserialize};
+use tokio::task;
+
+#[derive(Serialize, Deserialize, JsonSchema)]
+pub struct ResourceListResult {
+    pub resources: Vec<ImportedManifest>,
+}
+
+#[tool_router]
+impl McpServer {
+    #[must_use]
+    pub fn new() -> Self {
+        Self {
+            tool_router: Self::tool_router()
+        }
+    }
+
+    #[tool(
+        description = "List all DSC resources available on the local machine",
+        annotations(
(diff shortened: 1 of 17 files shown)


Gijsreyn:
image

dsc/src/mcp/list_resources.rs
```diff
@@ -0,0 +1,54 @@
...
+        description = "List all DSC resources available on the local machine",
+        annotations(
+            title = "Enumerate all available DSC resources on the local machine",
+            read_only_hint = true,
+            destructive_hint = false,
+            idempotent_hint = true,
+            open_world_hint = true,
+        )
+    )]
+    async fn list_resources(&self) -> Result<Json<ResourceListResult>, McpError> {
```
JohnMcPMS: It feels like a tool that uses the noun "resource" and shares a symbol with `ServerHandler` will be monumentally confusing...

I honestly don't know how the agent is going to handle the dual meaning of resource in the context of DSC+MCP.  But I do think it would be helpful for the humans reading this code for the tool to be named something else, even if it is:
```suggestion
    async fn list_dsc_resources(&self) -> Result<Json<ResourceListResult>, McpError> {
```

This might also be a better fit for MCP resources, possibly:
- A resource that is like `dsc resource list`; contains all resources with maybe `{ Type, Description }` to keep things short
- A resource template that is like `dsc resource list <Type> -o json` to have a more complete description.  This really only adds capabilities, but maybe more will go here eventually.

I think that the schema being part of the MCP resource template would be great, but I also think that the line between MCP resources and tools should be the need to invoke the DSC resource command or not.

dsc/src/mcp/list_resources.rs
```diff
@@ -0,0 +1,54 @@
...
+        let result = task::spawn_blocking(move || {
+            let mut dsc = DscManager::new();
+            let mut resources = Vec::new();
+            for resource in dsc.list_available(&DiscoveryKind::Resource, "*", "", ProgressFormat::None) {
```
JohnMcPMS: How would adapted resources be enumerated?

JohnMcPMS (commented):
Since I'm here, leaving my thoughts on a `find_dsc_resource` tool.  My expectations are:
1. There will be a very large number of DSC resources eventually
2. The context window will not be able to accommodate all of the information about every DSC resource

The general flow of the tool would be:
1. Takes in a string `user_query` that is intended to be a distillation of the user request.
2. Server uses sampling to request a list of keywords to use when searching.
3. Server filters resources based on given keywords.
4. Server uses sampling to have model choose the best resource from the results of 3.
5. Server responds with resource information sufficient for agent to use `invoke_dsc_resource` tool (includes capabilities and schema).

dsc/src/mcp/list_resources.rs
SteveL-MSFT: Adapted resource enumeration is more expensive so probably a separate tool

dsc/src/mcp/list_resources.rs
SteveL-MSFT: tools have descriptions, so I don't think the name of the tool is that critical?  I'm ok with changing the name, however.  For concerns about size/token constraints, my expectation is that the LLM would do a list which gives a short version of info and then a `get_resource` type tool to get full information

Gijsreyn:
@SteveL-MSFT forgot to ask during the community call, but will this be just as simple to invoke as the Bicep experimental MCP server? Would love to see it in VSCode, or instructions how to enable it.

SteveL-MSFT:
@JohnMcPMS add your thoughts here https://github.com/PowerShell/DSC/issues/1093 on MCP strategy for DSC

SteveL-MSFT:
@Gijsreyn to play with this, see https://code.visualstudio.com/docs/copilot/customization/mcp-servers on how to add MCP servers to VSCode.  This one is simply executed as `dsc mcp`.

tgauth (approved):

PR #1094, opened 2025-09-05
Mixed PowerShell example fix

Gijsreyn:
# PR Summary

~~This pull request adds a known issue when running `dsc resource list --adapter Microsoft.Windows/WindowsPowerShell` with the `$env:PSModulePath` containing a ";" at the end.~~ It also fixes up an example, which was reported on the dsc channel (Discord).

### DIFF
--- a/.pipelines/DSC-Official.yml
+++ b/.pipelines/DSC-Official.yml
@@ -246,7 +246,7 @@ extends:
       - job: BuildLinuxArm64Musl
         dependsOn: SetPackageVersion
         variables:
-          LinuxContainerImage: 'onebranch.azurecr.io/linux/ubuntu-2004-arm64:latest'
+          LinuxContainerImage: 'onebranch.azurecr.io/linux/ubuntu-2204-arm64:latest'
           PackageVersion: $[ dependencies.SetPackageVersion.outputs['Package.Version'] ]
           AzToken: $[ dependencies.SetPackageVersion.outputs['AzToken'] ]
           ob_outputDirectory: '$(Build.ArtifactStagingDirectory)'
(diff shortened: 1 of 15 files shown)


Gijsreyn:
PSDesiredStateConfiguration v2.0.7 doesn't have this issue.

docs/troubleshooting/known-issues.md
```diff
@@ -146,6 +146,39 @@ When executing `dsc.exe` commands in Windows PowerShell:
 - Use `ConvertTo-Json` without the `-Compress` parameter.
 - Consider using PowerShell 7+ for improved JSON handling compatibility.
 
+## Zero results when listing resources with Windows PowerShell adapter
+
+When running `dsc resource list --adapter Microsoft.Windows/WindowsPowerShell`, you
+may see zero resources returned. In trace or debug logs, an error like the following can
+appear: "Cannot bind argument to parameter 'Path' because it is an empty string."
```
michaeltlombardi:
``````suggestion
appear:

```console
Cannot bind argument to parameter 'Path' because it is an empty string.
```
``````

michaeltlombardi (commented):
Aside from a small formatting change, this documentation looks good. I'm going to file a separate issue for updating the adapter to return a specific exit code for this failure state, but that's nonblocking for this PR.

Gijsreyn:
@michaeltlombardi - actually after PR #1095 got merged, the issue becomes obsolete. I wasn't assuming we didn't want to specify a _version_ for example when it gets solved. I have removed the whole block. Only the example fix remains. I'll update the title.

SteveL-MSFT (changes requested):
Since this is a test config and not intended to be an example, I can still approve this as it probably makes more sense to test with WinPS but if the intent is an example, ideally we'd have a more scenario complete example we put into configurations folder

PR #1097, opened 2025-09-07
Fix empty path handling for Windows PowerShell adapter

### LINKED ISSUE #1095: Handle empty path segment in `PSModulePath` for `Microsoft.Windows/WindowsPowerShell` adapter resource

> As a user, I want to use adapted PSDSC resources in DSC without getting confusing errors when trying to list resources because my `PSModulePath` environment variable contains any empty segments or ends with a trailing separator (`;`).

As identified in #1094, the Windows PowerShell adapter currently fails when trying to list adapted PSDSC resources when the `PSModulePath` contains any empty segments or ends with a trailing separator (`;`).

We should either update the adapter to munge the path and avoid this issue or define a semantic exit code for this failure state and provide useful information to the caller.

#### Option 1: Munge `PSModulePath`

In this case, we should update the following lines for `Invoke-DscCacheRefresh` in `win_psDscAdapter.psm1` from:

https://github.com/PowerShell/DSC/blob/4750f779ecb27eb480f7ab37b8e8a526ed5054f5/powershell-adapter/psDscAdapter/win_psDscAdapter.psm1#L80

https://github.com/PowerShell/DSC/blob/4750f779ecb27eb480f7ab37b8e8a526ed5054f5/powershell-adapter/psDscAdapter/win_psDscAdapter.psm1#L262

To:

```powershell
$m = $env:PSModulePath -split [IO.Path]::PathSeparator | 
    Where-Object -FilterScript {-not [string]::IsNullOrEmpty($_) } |
    ForEach-Object -Process {
        Get-ChildItem -Directory -Path $_ -Depth 1 -ErrorAction Ignore
    }
```

Or define a helper function to invoke in those locations instead, like:

```powershell
function Get-ModuleDirectory {
    [CmdletBinding()]
    [OutputType([System.IO.DirectoryInfo])
    param()

    end {
        $env:PSModulePath -split [IO.Path]::PathSeparator | 
            Where-Object -FilterScript {-not [string]::IsNullOrEmpty($_) } |
            ForEach-Object -Process {
                Get-ChildItem -Directory -Path $_ -Depth 1 -ErrorAction Ignore
            }
        }
    }
}
```

And update the call sites to:

```powershell
$m = Get-ModuleDirectory
```

This would obviate the error behavior for users.

#### Option 2: Define semantic exit code

In this case, we should update the manifest to define exit code `2` with descriptive text like:

```jsonc
{
    // elided manifest fields
    "exitCodes": {
        "0": "Success",
        "1": "Unhandled error",
        "2": "Malformed PSModulePath"
    }
}
```

Then, we should add a `try {} catch {}` construction to the `win_psDscAdapter.psm1` wherever we use `Get-DscResource` and catch the specific case for this error. In the catch block, we should:

1. Emit a descriptive error message indicating the problem with `PSModulePath` (empty segment or trailing separator or both).
2. Exit with code `2`.

### Summary

I have a slight preference for intentionally handling the malformed `PSModulePath` over erroring. It's a failure state we can readily predict and avoid. It might be worth emitting a warning or info message indicating discovery of a malformed `PSModulePath`.

Gijsreyn:
# PR Summary

This pull request addresses an issue when `$env:PSModulePath` contains empty entries, the PSDesiredStateConfiguration v1.1 module errors out and returns empty results.

Rather than the suggestion in the fixed issue, a reset should always happen because of the line 158 and 170

## PR Context

Fix #1095

### DIFF
--- a/powershell-adapter/Tests/win_powershellgroup.tests.ps1
+++ b/powershell-adapter/Tests/win_powershellgroup.tests.ps1
@@ -118,7 +118,7 @@ class PSClassResource {
       New-Item -Path $modulePath -ItemType File -Value $module -Force | Out-Null
     }
 
-    $env:PSModulePath = $windowsPowerShellPath + [System.IO.Path]::PathSeparator + $env:PSModulePath
+    $env:PSModulePath = $windowsPowerShellPath + [System.IO.Path]::PathSeparator + $env:PSModulePath + [System.IO.Path]::PathSeparator
   }
 
   AfterAll {
--- a/powershell-adapter/psDscAdapter/win_psDscAdapter.psm1
+++ b/powershell-adapter/psDscAdapter/win_psDscAdapter.psm1
@@ -711,7 +711,8 @@ function Set-ValidPSModulePath {
     end {
         if (($env:PSModulePath -split [System.IO.Path]::PathSeparator) -contains '') {
             "Removing empty entry from PSModulePath: '$env:PSModulePath'" | Write-DscTrace -Operation Debug
-            $env:PSModulePath = ($env:PSModulePath -split [System.IO.Path]::PathSeparator | Where-Object { $_ -ne '' }) -join [System.IO.Path]::PathSeparator
+            $env:PSModulePath = ($env:PSModulePath -split [System.IO.Path]::PathSeparator | 
+                Where-Object { $_ -ne '' }) -join [System.IO.Path]::PathSeparator
         }
     }
 }
--- a/powershell-adapter/psDscAdapter/win_psDscAdapter.psm1
+++ b/powershell-adapter/psDscAdapter/win_psDscAdapter.psm1
@@ -711,7 +711,7 @@ function Repair-ValidPSModulePath {
     end {
         if (($env:PSModulePath -split [System.IO.Path]::PathSeparator) -contains '') {
             "Removing empty entry from PSModulePath: '$env:PSModulePath'" | Write-DscTrace -Operation Debug
-            $env:PSModulePath = [String]::Join([System.IO.Path]::PathSeparator, ($env:PSModulePath.Split([System.IO.Path]::PathSeparator, [System.StringSplitOptions]::RemoveEmptyEntries)))
+            $env:PSModulePath = [String]::Join([System.IO.Path]::PathSeparator, ($env:PSModulePath.Split([System.IO.Path]::PathSeparator, [System.StringSplitOptions]::RemoveEmptyEntries))).TrimEnd([System.IO.Path]::PathSeparator)
         }
     }
 }
--- a/powershell-adapter/psDscAdapter/win_psDscAdapter.psm1
+++ b/powershell-adapter/psDscAdapter/win_psDscAdapter.psm1
@@ -60,8 +60,8 @@ function Invoke-DscCacheRefresh {
     $namedModules = [System.Collections.Generic.List[Object]]::new()
     $cacheFilePath = Join-Path $env:LocalAppData "dsc\WindowsPSAdapterCache.json"
 
-    # Reset PSModulePath to not contain empty entries
-    Set-ValidPSModulePath
+    # Repair PSModulePath to not contain empty entries
+    Repair-ValidPSModulePath
 
     if (Test-Path $cacheFilePath) {
         "Reading from Get-DscResource cache file $cacheFilePath" | Write-DscTrace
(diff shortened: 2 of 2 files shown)


powershell-adapter/psDscAdapter/win_psDscAdapter.psm1
```diff
@@ -701,6 +704,20 @@ function GetClassBasedCapabilities {
     }
 }
 
+function Set-ValidPSModulePath {
```
SteveL-MSFT: Might be better to call this `Repair-PSModulePath`

powershell-adapter/psDscAdapter/win_psDscAdapter.psm1
```diff
@@ -701,6 +704,20 @@ function GetClassBasedCapabilities {
...
+    [CmdletBinding()]
+    param()
+
+    end {
+        if (($env:PSModulePath -split [System.IO.Path]::PathSeparator) -contains '') {
```
SteveL-MSFT, line 712: Seems like this could be simpler and faster just doing:

`$env:PSModulePath = [String]::Join([System.IO.Path]::PathSeparator, ($env:PSModulePath.Split([System.IO.Path]::PathSeparator, [System.StringSplitOptions]::RemoveEmptyEntries)))`

powershell-adapter/psDscAdapter/win_psDscAdapter.psm1
```diff
@@ -244,10 +247,10 @@ function Invoke-DscCacheRefresh {
             }
 
             $dscResourceCacheEntries.Add([dscResourceCacheEntry]@{
-                Type            = "$moduleName/$($dscResource.Name)"
-                DscResourceInfo = $DscResourceInfo
-                LastWriteTimes  = $lastWriteTimes
-            })
+                    Type            = "$moduleName/$($dscResource.Name)"
```
SteveL-MSFT: This indentation is wrong

powershell-adapter/Tests/win_powershellgroup.tests.ps1
```diff
@@ -118,7 +118,7 @@ class PSClassResource {
       New-Item -Path $modulePath -ItemType File -Value $module -Force | Out-Null
     }
 
-    $env:PSModulePath = $windowsPowerShellPath + [System.IO.Path]::PathSeparator + $env:PSModulePath
+    $env:PSModulePath = $windowsPowerShellPath + [System.IO.Path]::PathSeparator + $env:PSModulePath + [System.IO.Path]::PathSeparator
```
SteveL-MSFT, line 121: For the test case, would be good to append two pathseparators at the end to cover the empty case as well

SteveL-MSFT (changes requested):

powershell-adapter/Tests/win_powershellgroup.tests.ps1
Gijsreyn: I think that's why I already added that single one to cover it already (instead of two). See the red line (empty):

image

powershell-adapter/psDscAdapter/win_psDscAdapter.psm1
Gijsreyn: Agreed!

powershell-adapter/psDscAdapter/win_psDscAdapter.psm1
Gijsreyn: Quite frankly, I have returned it once again to the earlier indentation. Still, I find it pretty strange that when formatting, it constantly returns back with the `"powershell.codeFormatting.preset": "OTBS"` set. I think it is correct.

powershell-adapter/psDscAdapter/win_psDscAdapter.psm1
Gijsreyn: Missed the end separator. Now: `$env:PSModulePath = [String]::Join([System.IO.Path]::PathSeparator, ($env:PSModulePath.Split([System.IO.Path]::PathSeparator, [System.StringSplitOptions]::RemoveEmptyEntries))).TrimEnd([System.IO.Path]::PathSeparator)`

powershell-adapter/psDscAdapter/win_psDscAdapter.psm1
```diff
@@ -701,6 +704,19 @@ function GetClassBasedCapabilities {
...
 }
 
+function Repair-ValidPSModulePath {
+    [CmdletBinding()]
+    param()
+
+    end {
+        if (($env:PSModulePath -split [System.IO.Path]::PathSeparator) -contains '') {
+            "Removing empty entry from PSModulePath: '$env:PSModulePath'" | Write-DscTrace -Operation Debug
+            $env:PSModulePath = [String]::Join([System.IO.Path]::PathSeparator, ($env:PSModulePath.Split([System.IO.Path]::PathSeparator, [System.StringSplitOptions]::RemoveEmptyEntries))).TrimEnd([System.IO.Path]::PathSeparator)
```
SteveL-MSFT, line 714: I don't think the Trim is necessary if using the RemoveEmptyEntries?

Gijsreyn: Unfortunately, it was:

image

Issue #496, opened 2024-07-22
Copy loop iteration


Gijsreyn:
### Summary of the new feature / enhancement

As a user, I would expect you to be able to use the `copy` operation with a counter in the resource configuration document. This allows me to either drive resource configurations through the `--parameters-file` parameter or `--path` when doing an operation.

What I would expect is the following:

```yaml
# test.parameters.yaml
parameters:
  applications:
      - name: PowerShell
        id: 'Microsoft.PowerShell.Preview'

# test.config.yaml
$schema: https://raw.githubusercontent.com/PowerShell/DSC/main/schemas/2024/04/config/document.json
parameters:
  applications:
    type: object
resources:
  - name: Use class PowerShell resources
    type: Microsoft.DSC/PowerShell

    properties:
      copy:
        resources:
          - name: PowerShell 7 Preview
            type: Microsoft.WinGet.DSC/WinGetPackage
            properties:
              Id: "[parameters('applications.name').id]"
              Ensure: "Absent"
```

Other scenarios I'm thinking of, are the famously known DTAP street with configuration for Computer Management. 

Imagine you want to add users/groups to the relevant groups like _Adminstrator,_ or _Remote Desktop Users_. Organizations can follow naming conventions as `<groupname.<environmentName>.<regionName>', helping drive configurations from parameter files. 

You will in turn get something like:

```
# dev.group.parameters.json
parameters:
 groups:
      - name: Admin Group
        groupName: 'Administrators'
        members: ['testuser1', 'testuser2']
      - name: 'Remote Users'
        # truncated

# group.config.json
$schema: https://raw.githubusercontent.com/PowerShell/DSC/main/schemas/2024/04/config/document.json
parameters:
  groups:
    type: object

resources:
  - name: Use class PowerShell resources
    type: Microsoft.DSC/PowerShell
    properties:
      copy:
        resources:
          - name: PowerShell 7 Preview
            type: PSDscResources/GroupSet
            properties:
              GroupName "[parameters('name').groupName]"
              Members "[parameters('name').members]"
              Ensure: "Present"
```

It's a bit short, but if you want to expand on the idea, just give me a headsup.

SteveL-MSFT:
If I understand what you're trying to do, you want to use a resource that takes a single value and be able to specify an array so that single resource gets used for each element in the array?  Basically like a pipeline?

Gijsreyn:
[...]

PR #1099, opened 2025-09-10
Add `Copy` support for resources

SteveL-MSFT:
# PR Summary

Add `copy` support for resources.  This PR does not include `copy` support for variables nor properties which will come later.  The `mode` and `batchSize` properties for `copy` are not supported.

Symbolic names aren't supported, that'll be added as part of ARM Language 2.0 work

## PR Context

Fix https://github.com/PowerShell/DSC/issues/496

### DIFF
--- a/dsc_lib/src/configure/mod.rs
+++ b/dsc_lib/src/configure/mod.rs
@@ -904,7 +904,7 @@ impl Configurator {
                     self.context.copy.insert(copy.name.clone(), i);
                     let mut new_resource = resource.clone();
                     let Value::String(new_name) = self.statement_parser.parse_and_execute(&resource.name, &self.context)? else {
-                        return Err(DscError::Parser(t!("configure.mod.copyNameResultNotString", name = &copy.name).to_string()))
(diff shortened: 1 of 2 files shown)


dsc_lib/src/configure/context.rs
```diff
@@ -9,40 +9,56 @@ use std::{collections::HashMap, path::PathBuf};
...
+    Copy,
+    Normal,
+    NoExpressionEvaluation,
+    ParametersDefault,
+    UserFunction,
+}
+
+#[derive(Clone)]
 pub struct Context {
+    pub copy: HashMap<String, i64>,
```
tgauth, line 23: If `copy` is tracking multiple loop names, should there be a test for multiple?

SteveL-MSFT: Multiple (as in nested) only occurs when `copy` is implemented for properties.  Since that is more complicated, didn't want to make this PR any bigger so the tracking is just scaffolding for now.

Issue #1093, opened 2025-09-04
MCP Server


SteveL-MSFT:
### Summary of the new feature / enhancement

`dsc.exe` will be a MCP server started as `dsc mcp`.  It will support STDIO transport.  This is to support future AI scenarios where the user wants to use MCP clients like CoPilot to make changes ot the system which results in invoking a DSC resource.

### Proposed technical implementation details (optional)

As LLMs are still evolving and we want to support SLM use cases as well, we should be thoughtful about the number of tokens returned as to not exceed limits.  Additionally, we should assume in the future that a system may have thousands of DSC resources (locally or perhaps through remote discovery from a repository), so need a strategy for this to work effectively.

Proposed tools in implementation order:

## `list_dsc_resources([adapter])`

This will provide a small list of resources found on the current system.  Should include the `typeName`, `description` and `kind`.  Only 1 version should be returned.

## `show_dsc_resource(<typeName>)`

This retrieves all the information for one specific DSC resource.  Only latest version is returned.  Includes schema.

## `validate_dsc_resource_schema(<typeName>, <input>)`

Used to validate potential input JSON against the resource JSONSchema.

## `invoke_dsc_resource(<operation>, <typeName>, [input])`

Invoke `get`, `set`, `test`, or `export` against a specific resource.

## `list_configuration_examples()`

Returns a list of embedded example configurations (YAML) returning `name` and `description`.  This is intended to give the LLM some example of what a proper configuration doc should look like.

## `show_configuration_example(<name>)`

Returns one example configuration (YAML)

## `validate_configuration(<input>)`

Perform static analysis of the configuration.

## `invoke_dsc_configuration(<operation>, <input>)`

Invoke `get`, `set`, `test`, or `export` against the configuration.

JohnMcPMS:
I think that some of these tools might be better served as MCP resources (or resource templates), in particular the examples but also potentially the list.

`show_dsc_resource` "retrieves all the information" suggests no need for `show_dsc_resource_schema`, and I would agree that it is generally good to keep the tool count low.

Regarding limitations on output, my thoughts on a `find_dsc_resource` tool with the goal of sifting through a large number of DSC resources without requiring that their data all fits within the context window.

The general flow of the tool would be:
1. Takes in a string `user_query` that is intended to be a distillation of the user request.
2. Server uses sampling to request a list of keywords to use when searching.
[...]

PR #1105, opened 2025-09-15
Add `show_dsc_resource()` and enable `list_dsc_resource([adapter])` for MCP

SteveL-MSFT:
# PR Summary

As discussed in https://github.com/PowerShell/DSC/issues/1093#issuecomment-3289113811 add an optional parameter to `list_dsc_resource([adapter])` instead of having a separate tool for listing adapted resources.  Add `show_dsc_resource()` tool to get detailed information including schema for a dsc resource.

## PR Context

Part of https://github.com/PowerShell/DSC/issues/1093


dsc/src/mcp/show_dsc_resource.rs
```diff
@@ -0,0 +1,81 @@
...
+    /// The author of the resource.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub author: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub schema: Option<Value>,
+}
+
+#[derive(Deserialize, JsonSchema)]
+pub struct ShowResourceRequest {
+    #[schemars(description = "The type name of the resource to get detailed information.")]
```
tgauth, line 42: does this need to be localized?

SteveL-MSFT: Good question.  LLMs currently tend to be English only.  I think we can just accept the MCP code to have English text and not localized for now.

PR #1117, opened 2025-09-18
Support expression validation in resource name

### LINKED ISSUE #1116: Resource names doesnt support expression validation
### Prerequisites

### Summary

Looking from a Bicep perspective, the following example showcases how to give a name based on a parameter:

```bicep
param prefix string = 'myapp'
param environment string = 'dev'

resource storage 'Microsoft.Storage/storageAccounts@2023-01-01' = {
  name: '${prefix}${environment}sa'
  location: resourceGroup().location
  sku: {
    name: 'Standard_LRS'
  }
  kind: 'StorageV2'
}
```

However, trying to do the same in DSC v3 doesn't evaluate the expression:

```yaml
parameters:
  resourceName:
    type: string
    defaultValue: name
resources:
- name: "[concat('start-', parameters('resourceName'))]"
  properties:
    output:
      hasSvcPrefix: "[concat('test-', parameters('resourceName'))]"
```

Output:

```plaintext
results:
- metadata:
    Microsoft.DSC:
      duration: PT0.0492017S
  name: '[concat(''start-'', parameters(''resourceName''))]'
  result:
    actualState:
      output:
        hasSvcPrefix: test-name
```

### Steps to reproduce

Run the example from above.

### Expected behavior

```console
results:
- metadata:
    Microsoft.DSC:
      duration: PT0.1481929S
  name: start-name
  result:
    actualState:
      output:
        hasSvcPrefix: test-name
```

### Actual behavior

```console
The expression is not evaluated
```

### Error details

```console

```

### Environment data

```PowerShell
PSVersion                      7.5.3
GitCommitId                    7.5.3
```

### Version

dsc 3.2.0-preview.5

### Visuals

_No response_

Gijsreyn:
# PR Summary

Moved resource name expression evaluation from validate_config() to a new process_resource_names() method. Added the method after set_context(0 to ensure parameters are available.

## PR Context

Fix #1116


dsc/src/subcommand.rs
```diff
@@ -381,6 +381,12 @@ pub fn config(subcommand: &ConfigSubCommand, parameters: &Option<String>, parame
         exit(EXIT_INVALID_INPUT);
     }
 
+    // process resource names after parameters are available
+    if let Err(err) = configurator.process_resource_names() {
+        error!("Error processing resource names: {err}");
```
SteveL-MSFT: String should be localized

dsc_lib/src/configure/mod.rs
```diff
@@ -912,6 +912,39 @@ impl Configurator {
...
+    /// have been processed, ensuring that parameter references work correctly.
+    ///
+    /// # Errors
+    ///
+    /// This function will return an error if:
+    /// - Resource name expression evaluation fails
+    /// - Expression does not result in a string value
+    /// - Statement parser encounters invalid syntax
+    pub fn process_resource_names(&mut self) -> Result<(), DscError> {
+        let mut config = self.config.clone();
```
SteveL-MSFT: Can you avoid making a clone and another copy?

dsc_lib/src/configure/mod.rs
```diff
@@ -912,6 +912,39 @@ impl Configurator {
...
+        let config_copy = config.clone();
+        
+        for resource in config_copy.resources {
+            // skip resources that were created from copy loops (they already have evaluated names)
+            if resource.copy.is_none() && resource.name.starts_with('[') && resource.name.ends_with(']') {
```
SteveL-MSFT: You shouldn't have to check if the name looks like an expression as `parse_and_execute` will handle non-expressions as literals and return it

dsc_lib/src/configure/mod.rs
```diff
@@ -912,6 +912,39 @@ impl Configurator {
...
+                // process resource name expressions for non-copy resources
+                let Value::String(new_name) = self.statement_parser.parse_and_execute(&resource.name, &self.context)? else {
+                    return Err(DscError::Parser(t!("configure.mod.nameResultNotString").to_string()))
+                };
+                // find and update the resource name in the config
+                if let Some(config_resource) = config.resources.iter_mut().find(|r| r.name == resource.name && r.resource_type == resource.resource_type) {
```
SteveL-MSFT: This also seems expensive, ideally, I think what you're doing here should be when a logical resource in the config is being invoked then the name is evaluated as an expression first before the properties are actually sent to the resource.  This should be somewhere near where `condition` is processed.

dsc_lib/src/configure/mod.rs
```diff
@@ -912,6 +920,37 @@ impl Configurator {
...
+    ///
+    /// # Errors
+    ///
+    /// This function will return an error if:
+    /// - Resource name expression evaluation fails
+    /// - Expression does not result in a string value
+    /// - Statement parser encounters invalid syntax
+    fn evaluate_resource_name(&mut self, resource: &Resource) -> Result<String, DscError> {
+        // skip resources that were created from copy loops (they already have evaluated names)
+        if resource.copy.is_some() {
```
SteveL-MSFT: You don't need to pass in the whole resource, just a ref to the name should be sufficient.  To check if `copy` is happening, you can check of `context.process_mode == ProcessMode::Copy`

SteveL-MSFT (changes requested):

SteveL-MSFT (approved):

Issue #931, opened 2025-06-30
New adapter interface for resource invocation


SteveL-MSFT:
### Summary of the new feature / enhancement

The current adapter model currently expects adapters to implement the equivalent of `dsc config ...` where the original motivation was to only create a single hosting process that all the nested resources would execute within.  However, there have been several problems from this:

- implicit use of adapted resources requires the engine to generate a nested config at runtime to pass to the adapter and this incurs another discovery (there's no clean way to resolve this since the nested one is deep in the runtime stack)
- users are more likely to use adapted resources implicitly than wrap them
- adapters need to understand and implement config level features

### Proposed technical implementation details (optional)

We add a new resource adapter model that can be identified in the resource manifest as:

```json
{
  "kind": "resourceAdapter"
}
```

This would require these kind of adapters to perform the equivalent as `dsc resource ...`.  Benefits:

- no extra discovery perf penalty when used implicilty
- process isolation is better for security and reliability reasons
- simpler for adapters to implement
- engine can resolve all expressions and pass context to this adapter which can pass to resources

michaeltlombardi:
Related:

- #267
- #381
- #730
- #849

Ideally, I would want to be able to _both_ individually invoke a single resource with an adapter and handle `n` instances _without_ being responsible for any configuration expressions or munging output as much as possible.

Does it make sense to support two different kinds of adapters, or can we find a way to extend/rework the current (data/manifest) model to support this?

Alternative ideas:

- Add the `invoke` field under `adapter`, which indicates direct invocation (with sub-properties for the operation commands) where the top-level fields are for the grouped form.
- As above, but make the top-level fields be for direct invocation of adapted resources and the nested fields be for the group form.
- Add a field to `adapter` to clarify whether the resource supports single or group invocation.
- Add a new argument object to operation fields to support passing a conditional discriminator argument, like `--group`.

SteveL-MSFT:
I originally wanted to encapsulate it within the `adapter` kind, but I think it would actually make it more confusing since it has to default to current model which is why I went with an alternate kind instead.  Multi is tricky without some look-ahead and then what if you have expressions that depend on previous execution first (like the output via `reference()`) via implied dependency order.

[...]

PR #1124, opened 2025-09-20
Add new inputKind for adapters to work against single resource instead of config

SteveL-MSFT:
# PR Summary

- the `config` property within `adapter` section of resource manifest is renamed to `inputKind` which accepts `full` (meaning config) or `single` (meaning just a single resource).  However, `config` is an allowed alias for backwards compat.
- as noted above, the resource manifest format changed, best example is the [test adapter manifest](https://github.com/PowerShell/DSC/compare/main...SteveL-MSFT:adapter-single?expand=1#diff-4bb99b52642918c9657d7684361a095bc88088d2f143fbb79e6eef34776ada8f) which shows specifying `single` for `inputKind`, but also for the args, there's a new object:

```json
            {
                "resourceTypeArg": "--resource-type"
            }
```

  where this arg object will use the value of `resourceTypeArg` as the parameter nae and supply the target resource type for the adapter as the parameter value.

- the `launch.json` change only applies to macOS and leverages an OSS script to pretty print variables in VSCode under debugging instead of showing the memory address
- the previous embedded code for `get`, `set`, `test`, `delete`, and `export` for adapters got moved to help functions which handles `full` or `single` adapter invocation
- since `process_args()` needs the target resource type to fill in the parameter value, some plumbing was added in the existing functions to ensure this gets passed through
- helper function added to retreive the adapter `inputKind`
- test adapter added to `dsctest` which doesn't adapt anything but holds the resources internally

## PR Context

Fix https://github.com/PowerShell/DSC/issues/931

### DIFF
--- a/dsc_lib/src/dscresources/dscresource.rs
+++ b/dsc_lib/src/dscresources/dscresource.rs
@@ -415,32 +415,7 @@ impl Invoke for DscResource {
     fn test(&self, expected: &str) -> Result<TestResult, DscError> {
         debug!("{}", t!("dscresources.dscresource.invokeTest", resource = self.type_name));
         if let Some(adapter) = &self.require_adapter {
-            let mut configurator = self.clone().create_config_for_adapter(adapter, expected)?;
-            let result = configurator.invoke_test()?;
-            let TestResult::Resource(ref resource_result) = result.results[0].result else {
-                return Err(DscError::Operation(t!("dscresources.dscresource.invokeReturnedWrongResult", operation = "test", resource = self.type_name).to_string()));
-            };
-            let mut desired_state = resource_result.desired_state
-                .as_object().ok_or(DscError::Operation(t!("dscresources.dscresource.propertyIncorrectType", property = "desiredState", property_type = "object").to_string()))?
-                .get("resources").ok_or(DscError::Operation(t!("dscresources.dscresource.propertyNotFound", property = "resources").to_string()))?
-                .as_array().ok_or(DscError::Operation(t!("dscresources.dscresource.propertyIncorrectType", property = "resources", property_type = "array").to_string()))?[0]
-                .as_object().ok_or(DscError::Operation(t!("dscresources.dscresource.propertyIncorrectType", property = "resources", property_type = "object").to_string()))?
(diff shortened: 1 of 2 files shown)


dsc_lib/src/dscresources/dscresource.rs
```diff
@@ -129,6 +132,149 @@ impl DscResource {
...
+        };
+        let properties = resource_result.actual_state
+            .as_object().ok_or(DscError::Operation(t!("dscresources.dscresource.propertyIncorrectType", property = "actualState", property_type = "object").to_string()))?
+            .get("result").ok_or(DscError::Operation(t!("dscresources.dscresource.propertyNotFound", property = "result").to_string()))?
+            .as_array().ok_or(DscError::Operation(t!("dscresources.dscresource.propertyIncorrectType", property = "result", property_type = "array").to_string()))?[0]
+            .as_object().ok_or(DscError::Operation(t!("dscresources.dscresource.propertyIncorrectType", property = "result", property_type = "object").to_string()))?
+            .get("properties").ok_or(DscError::Operation(t!("dscresources.dscresource.propertyNotFound", property = "properties").to_string()))?.clone();
+        let get_result = GetResult::Resource(ResourceGetResponse {
+            actual_state: properties.clone(),
+        });
```
tgauth, line 156: I see that this is just copied from the existing `get()` function but defining an `actual_state` struct (or more strongly typing the existing struct) could simplify serialization and improve readability, with something like:  

```
#[derive(Serialize, Deserialize)]
struct Result {
    properties: Value,
    #[serde(flatten)]
    extra: Value
}

#[derive(Serialize, Deserialize)]
struct ActualState {
    result: Vec<Result>,
    #[serde(flatten)]
    extra: Value
}

let actual_state: ActualState = serde_json::from_value(resource_result.actual_state.clone())?;
let properties = &actual_state.result.get(0)
    .ok_or_else(|| DscError::Operation("result array empty".into()))?
    .properties;
```

SteveL-MSFT: You are right, but since this is moved code, I think we should do that as a separate PR rather than complicate this one

Issue #327, opened 2024-02-24
Signing resource manifests


SteveL-MSFT:
### Summary of the new feature / enhancement

Need a way to sign a resource manifest to ensure it hasn't been tampered.  The manifest should include a hash of the executable or the thumbprint of it's signature otherwise the resource manifest is trusted, but uses a bogus executable.

michaeltlombardi:
Also worth noting that we should consider as _at least **related**_ SBOMs. If I care about the manifest being signed, I also probably care about the SBOM for a resource.

Relatedly, I discovered the [GitHub uses cosign][gh-cosign] for artifact and sbom attestations - and that the [rust library for cosign/sigstore][rs-sigstore] is under active development. This could make for an easier path towards having a built-in model for this domain when we approach signed resources, especially in consideration with publishing them to an OCI Registry (see #92).

[gh-cosign]: https://docs.github.com/en/actions/security-for-github-actions/using-artifact-attestations/using-artifact-attestations-to-establish-provenance-for-builds
[rs-sigstore]: https://github.com/sigstore/sigstore-rs

Issue #210, opened 2023-09-27
Signing configurations


SteveL-MSFT:
### Summary of the new feature / enhancement

Need a way to enable only executing signed/trusted configurations

michaeltlombardi:
Tools like [cosign][01], [ansible-sign][02], and [minisign][03] are what come to mind initially.

I've been following cosign for a while. It's initial selling point was signing container images (or ORAS artifacts, which might be worth keeping in mind re #92), but it also supports signing arbitrary blobs.

Relatedly, though possibly a fully separate issue, is whether a sort of sbom for configuration documents makes sense (unwinding the declared resource types and giving a more "full" snapshot of what the configuration document intended to convey).

Also worth keeping in mind on the approach whether/how to sign and verify manifests, and whether/how to indicate that DSC should only use signed and trusted resources _and_ manifests.

[01]: https://docs.sigstore.dev/signing/signing_with_blobs/
[02]: https://github.com/ansible/ansible-sign
[03]: https://jedisct1.github.io/minisign/

ThomasNieto:
Should DSC be validating signed configurations or should that be at a higher layer like an agent?

It could be problematic that DSC is making these kinds of technology decisions as what DSC has been proposed is that is the lowest level tool, read configuration and apply using resources. You're getting more into the solution layer when deciding which plain text configuration is valid or not.

As an example, in my [thought experiment](https://github.com/opendsc/opendsc) of bringing back the LCM and pull server in DSCv3 I was going to use GPG as the tool as it was cross-platform and was able to validate signatures and encrypted at rest/transit between pull server and node before passing the configuration to DSC.

michaeltlombardi:
@ThomasNieto - that's a fair point. I think there's still value in signing and sboms for _publishing_ DSC Resources (or reusable configurations, potentially) to a registry, but I see the sense in what you're saying around higher order tools being responsible for "should my users trust and execute this configuration or resource?"

ThomasNieto:
@michaeltlombardi I agree that there should be signing for DSC resources. The problem I see is that DSC supports resources in different languages and each use different signing schemes. Should DSC enforce its own signing policy or should that be left up to each language? If DSC enforces its own signing then the author may have to support multiple signing  technologies. For example, if I write my DSC resource in PowerShell or C# I can use Authenticode to sign the files. If I decide to write in a Linux paradigm I may use GPG to detach sign my files and publish the public key on the website for user validation.

[...]

PR #1130, opened 2025-09-23
Add warning on Windows if files aren't authenticode signed

SteveL-MSFT:
# PR Summary

If a configuration file, resources manifest, or the exe used by the resource manifest is not authenticode signed on Windows, you will get a warning message.

Also, when listing extensions and resources, there is a new `Trust` column/property indicating the trust level.  On Linux/macOS, it's currently always `unknown`

Future work will make it configurable if the warning is an error.  Catalog signed files are currently out-of-scope.

The Win32 calls replicate the example from the docs https://learn.microsoft.com/en-us/windows/win32/seccrypto/example-c-program--verifying-the-signature-of-a-pe-file

Because of the new warning message, many existing tests were breaking expecting no messages so added setting `DSC_TRACE_LEVEL='error'` to many tests and removing it at end.  Any explicit setting of trace level will override this.

## PR Context

Fix https://github.com/PowerShell/DSC/issues/210
Fix https://github.com/PowerShell/DSC/issues/327

### DIFF
--- a/dsc_lib/src/security/mod.rs
+++ b/dsc_lib/src/security/mod.rs
@@ -109,6 +109,6 @@ pub fn check_file_security(file_path: &Path) -> Result<TrustLevel, DscError> {
 /// # Errors
 /// This function does not return any errors on non-Windows platforms.
 #[cfg(not(windows))]
-pub fn check_file_security(_file_path: &Path) -> Result<(), DscError> {
-    Ok(())
+pub fn check_file_security(_file_path: &Path) -> Result<TrustLevel, DscError> {
+    Ok(TrustLevel::Unknown)
 }


TravisEz13:
Please elaborate on why you are verifying the file

ThomasNieto:
Can we have an issue or rfc on this PR? I have concerns about the current design and how cross platform differences.

SteveL-MSFT:
> Can we have an issue or rfc on this PR? I have concerns about the current design and how cross platform differences.

This is linked to existing issues.  Cross platform is not part of this PR as we are targeting Windows first.

SteveL-MSFT:
Thinking about this, I think it would be better to change this a bit so that whether a manifest or the target exe is signed should be part of the DscResource struct and shows up under `dsc resource list` in addition to the warning message when used.

SteveL-MSFT:
Since `powershell.exe` is used for some resources, will need to add catalog signing to cover Windows inbox files later.

SteveL-MSFT:
Found a sample here https://github.com/microsoft/Windows-classic-samples/blob/main/Samples/Security/CodeSigning/cpp/codesigning.cpp I can use for catalog signed verification, so will update this PR since it doesn't make sense currently for it to claim `powershell.exe` is untrusted

PR #1132, opened 2025-09-25
Add static memory cache for discovered resources

SteveL-MSFT:
# PR Summary

When using adapted resources, the engine will dynamically generate a nested configuration.  Each configuration (outer and inner) would incur the cost of performing discovery.  This change makes command discovery retain an in-memory static cache so that any multiple discoveries will just use the previous one.


dsc_lib/src/discovery/command_discovery_cache.rs
```diff
@@ -0,0 +1,79 @@
...
+    None
+}
+
+pub fn get_resources() -> BTreeMap<String, Vec<DscResource>> {
+    RESOURCES.lock().unwrap().clone()
+}
+
+pub fn resources_is_empty() -> bool {
+    RESOURCES.lock().unwrap().is_empty()
+}
```
andyleejordan: While I think the code is fine as-is, just for fun I'm sure curious if there's a way to make a template function that essentially applies `.lock().unwrap().clone()` as the name `get()` (or similar) etc. that would transform all these to just `RESOURCES.get()` etc. Like, this works, but there's clearly some repetition (which is fine, and arguably useful in terms of readability) that just makes me curious about a functional approach to it. I'd have to dig into Rust...oh no 🤭

dsc_lib/src/discovery/command_discovery.rs
```diff
@@ -307,7 +305,7 @@ impl ResourceDiscovery for CommandDiscovery {
         match kind {
             DiscoveryKind::Resource => {
                 // Now we need to call discover extensions and add those resource to the list of resources
-                for extension in self.extensions.values() {
+                for extension in get_extensions().values() {
```
andyleejordan: I'll just note that with this change, we're intentionally trading off for memory (which is ample nowadays) and possibly some amount of speed due to the `clone()` happening on the globals, as long as I remember how Rust's clone works anyway. That seems more than reasonable to avoid the issue this PR is addressing in the first place (them not being cached).

dsc_lib/src/discovery/command_discovery.rs
```diff
@@ -320,26 +318,27 @@ impl ResourceDiscovery for CommandDiscovery {
...
+        if resources_is_empty() && adapters_is_empty() {
             self.discover(&DiscoveryKind::Resource, "*")?;
         }
 
-        if self.adapters.is_empty() {
+        if adapters_is_empty() {
             return Ok(());
         }
 
+        let adapters = get_adapters();
```
andyleejordan: Is there any possible race condition between this (where it's cloned) and where it's used a few times below?

andyleejordan (commented):
Some questions, and some Rust I need to go look up.

dsc_lib/src/discovery/command_discovery_cache.rs
SteveL-MSFT: The helper functions abstract the unlocking of the mutex and unwrapping the result to get access to the actual hash table.  The function calls to get the lock, unwrapping, and then accessing functions of the hashtable have to be somewhere.

dsc_lib/src/discovery/command_discovery.rs
SteveL-MSFT: Yes, this is a trade-off to `clone()` which consumes memory vs cpu/disk to perform discovery.

dsc_lib/src/discovery/command_discovery.rs
SteveL-MSFT: All the helper functions that return are clones so the race condition (which doesn't exist today since it's single threaded essentially) is that the function is working off a clone and the actual source might have changed.  The mutex, currently, isn't required, but then the Rust compiler requires `unsafe {}` code to access globals due to the risk of a race condition.

dsc_lib/src/discovery/command_discovery_cache.rs
andyleejordan: I don't think I explained it well. I understand what it's doing, but was trying to suggest a pattern like this using Rust macros (metaprogramming, kinda like a partial, https://doc.rust-lang.org/rust-by-example/macros/dry.html). These thirteen functions could be replaced with the following four macros and used directly:

```rust
#[macro_export]
macro_rules! locked_is_empty {
    ($lockable:expr) => {{
        $lockable.lock().unwrap().is_empty()
    }};
}

#[macro_export]
macro_rules! locked_extend {
    ($lockable:expr, $items:expr) => {{
        $lockable.lock().unwrap().extend($items);
    }};
}

#[macro_export]
macro_rules! locked_clone {
    ($lockable:expr) => {{
        $lockable.lock().unwrap().clone()
    }};
}

#[macro_export]
macro_rules! locked_get {
    ($lockable:expr, $key:expr) => {{
        if let Some(v) = $lockable.lock().unwrap().get($key) {
            Some(v.clone())
        } else {
            None
        }
    }};
}

// Usage in command_discovery.rs would look like:
...
for extension in locked_clone!(EXTENSIONS).values() {
...
locked_extend!(ADAPTERS, adapters);
locked_extend!(RESOURCES, resources);
...
if let Some(resources) = locked_get!(RESOURCES, filter.resource_type()) {
...
```

Completely untested, and just a programming style change (one macro per thing we're trying to do so O(n), rather than a function for each thing we do to each static variable or O(n^2)). End result is the same, but it's reusable.

andyleejordan: You _could_ also do it as traits for `LazyLock<...>` and then be able to chain it instead of passing as the variables as args, but the implementation is more annoying.

SteveL-MSFT: I see what you're going for, but it doesn't seem like it saves anything and adds macros everywhere?

andyleejordan: With a slight refactoring the whole new file disappears: https://github.com/andyleejordan/DSC/commit/a53e56c76d5632538a3cada84b62938948a8ec9e

It's just a functional programming way of doing it instead of imperative, but hey that's a lot of why I like Rust 😆

SteveL-MSFT: @andyleejordan since those helper functions aren't public, I suppose your change is fine.  Do you want to create a PR into my branch for this PR?

andyleejordan: Sure!

andyleejordan: https://github.com/SteveL-MSFT/DSC/pull/1

SteveL-MSFT:
@andyleejordan FYI - the mutex with macros was resulting in a deadlock, changing to using `RwLock` which is better anyways since don't need to mutate every time fixes this

andyleejordan:
> @andyleejordan FYI - the mutex with macros was resulting in a deadlock, changing to using `RwLock` which is better anyways since don't need to mutate every time fixes this

I'm curious where exactly the deadlock was. If it didn't deadlock in the same way with the wrappers, that indicates we're acquiring the lock in the same scope in some place where we meant not to hold the lock (and so were implicitly depending on the lock expiring inside the wrapper)...ah, probably where we're reading to check `is_empty()` or `clone()` it and then writing to it with `extend()`...so was it here?

```
fn discover()
...
        match kind {
            DiscoveryKind::Resource => {
...
// clone and then extend deadlocking?
```

If that's the case, then I agree, `RwLock` is the correct type to use given the usage.

PR #1134, opened 2025-09-26
Add context() function reference documentation

Gijsreyn:
# PR Summary

This pull request adds reference documentation for the `context()` function. Also included it in the overview.


michaeltlombardi:
This PR has me thinking that we need to define a schema for the output of this function.

Probably something like:

```tree
.
├── docs/
│   └── reference/
│       └── schema/
│           └── config/
│               └── functions/
│                   └── context/
│                       ├── index.md (function, overview)
│                       └── output.md (detailed schema reference for the output object)
└── schemas/
    └── src/
        └── config/
            └── functions/
                └── output/
                    └── context.yaml
```

We should get the schema definition in first, then this reference documentation.

I'm having a think about whether and how to surface expressions more coherently in the schemas, but that's out of scope for now. We at least need to define the schema for this object and document it separately from (but related to) the configuration expression itself.

michaeltlombardi:
Superceded by #1402

PR #1135, opened 2025-09-26
Add user-defined function reference documentation

Gijsreyn:
# PR Summary

This pull request adds reference documentation on how to use user-defined functions.

### DIFF
--- a/docs/reference/schemas/config/functions/overview.md
+++ b/docs/reference/schemas/config/functions/overview.md
@@ -623,6 +623,13 @@ The following list of functions create or convert values of a given type:
   same type.
 - [int()][int] - Convert a string or number with a fractional part into an integer.
 
+### User-defined functions
+
+DSC also supports creating custom functions within configuration documents:
+
+- [User-defined functions][user-defined] - Define and invoke custom functions
+  with parameters, type validation, and reusable logic.
+
 <!-- Link references -->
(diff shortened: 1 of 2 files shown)


michaeltlombardi:
Thanks for another great submission, @Gijsreyn!

Thinking on this though, I believe we probably need to get the schema source defined _first_ and then define reference documentation. I've been thinking about the overall structure and _probably_ we need to do something like this:

```tree
.
├── docs/
│   ├── concepts/
│   │   └── config/
│   │       └── custom-functions/
│   │           ├── index.md (overview)
│   │           ├── defining.md
│   │           └── using.md
│   └── reference/
│       └── schema/
│           └── config/
│               └── functions/
│                   └── custom/
│                       ├── index.md (overview, usage, array)
│                       ├── definition.md (member object definition)
│                       ├── parameters.md (parameter definition)
│                       └── output.md (output definition)
└── schemas/
    └── src/
        └── config/
            └── document/
                └── document.functions.yaml
```

And prioritize (maybe across multiple PRs) schema -> reference -> concepts.

This structure would also allow us to break out the definition a bit, which would enable us to define the schema types and so on for every subproperty under an appropriate section heading (instead of nested lists without metadata, like we have here). That's also very helpful for deeplinking to relevant sections.

PR #1137, opened 2025-09-26
Add DSC MCP server concept

Gijsreyn:
# PR Summary

This pull request adds a new conceptual document on how to get started with the DSC MCP server. I kept the experimental and preview out of the text, as I assume a release 3.2.0 will be coming soon.


docs/concepts/dsc-mcp-server.md
```diff
@@ -0,0 +1,246 @@
...
+  enhances the authoring experience in supported editors like VS Code.
+ms.date: 09/26/2025
+title: Using DSC with Model Context Protocol (MCP) Server
+---
+
+# Using DSC with Model Context Protocol (MCP) Server
+
+Microsoft's Desired State Configuration (DSC) platform includes a built-in Model Context Protocol
+(MCP) server that enables AI agents and development tools to interact with DSC resources and
+functions through a standardized interface. This integration provides intelligent assistance for
```
SteveL-MSFT: Although config support doesn't exist yet, I think it's better to call it out than functions which is a subset of configuration as this is the direction we're heading anyways

```suggestion
configurations through a standardized interface. This integration provides intelligent assistance for
```

docs/concepts/dsc-mcp-server.md
```diff
@@ -0,0 +1,246 @@
...
+
+### `list_dsc_resources`
+
+Lists summary information for all DSC resources available on the local machine, including:
+
+- Resource type name and kind
+- Description and capabilities
+- Adapter requirements (for adapted resources)
+
+You can optionally filter to show only adapted resources that require a specific adapter type.
```
SteveL-MSFT: We want to make sure the reader doesn't think these are tools THEY directly call, but they are tools for agents to call

```suggestion
Agents can optionally provide a filter to show only adapted resources that require a specific adapter type.
```

docs/concepts/dsc-mcp-server.md
```diff
@@ -0,0 +1,246 @@
...
+external data sources and tools. To learn more about MCP and how it works with AI agents, see
+[Use MCP servers in VS Code][00].
+
+## Available DSC MCP tools
+
+The DSC MCP server provides three core tools for AI agents to interact with your DSC environment:
+
+### `list_dsc_resources`
+
+Lists summary information for all DSC resources available on the local machine, including:
```
SteveL-MSFT: I think it might be better to not have too many technical details here since the details are for the agent and not the user, but instead use these sections to describe why an agent might use this tool.  In this case, the user may be asking to change something that happens to be exposed as a DSC resource so the agent can discover this and use it to fulfill the user request.

SteveL-MSFT (changes requested):

docs/concepts/dsc-mcp-server.md
Gijsreyn: @SteveL-MSFT  - I hope I understood it correctly. I have now rephrased the sentences to focus less on technical details and provided a couple of examples. Please review it again and let me know your thoughts. Thanks!

Gijsreyn:
@SteveL-MSFT - with all the comments resolved, is this something that can be merged into main?

PR #1138, opened 2025-09-26
Add `intersection()` function

Gijsreyn:
# PR Summary

This pull request adds the `intersection()` function, including reference documentation.

## PR Context

Partially addresses #57.

### DIFF
--- a/dsc/tests/dsc_functions.tests.ps1
+++ b/dsc/tests/dsc_functions.tests.ps1
@@ -111,6 +111,76 @@ Describe 'tests for function expressions' {
     }
   }
 
+  It 'intersection function works for: <expression>' -TestCases @(
+    @{ expression = "[intersection(parameters('firstArray'), parameters('secondArray'))]"; expected = @('cd') }
+    @{ expression = "[intersection(parameters('firstObject'), parameters('secondObject'))]"; expected = [pscustomobject]@{ two = 'b' } }
+    @{ expression = "[intersection(parameters('thirdArray'), parameters('fourthArray'))]"; expected = @('ef', 'gh') }
+    @{ expression = "[intersection(parameters('thirdObject'), parameters('fourthObject'))]"; expected = [pscustomobject]@{ three = 'd' } }
+    @{ expression = "[intersection(parameters('firstArray'), parameters('thirdArray'))]"; expected = @() }
+    @{ expression = "[intersection(parameters('firstObject'), parameters('firstArray'))]"; isError = $true }
+  ) {
+    param($expression, $expected, $isError)
+
+    $config_yaml = @"
+            `$schema: https://aka.ms/dsc/schemas/v3/bundled/config/document.json
+            parameters:
+              firstObject:
+                type: object
+                defaultValue:
+                  one: a
+                  two: b
+              secondObject:
+                type: object
+                defaultValue:
+                  two: b
+                  three: d
+              thirdObject:
+                type: object
+                defaultValue:
+                  two: c
+                  three: d
+              fourthObject:
+                type: object
+                defaultValue:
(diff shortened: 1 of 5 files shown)


dsc/tests/dsc_functions.tests.ps1
```diff
@@ -111,6 +111,76 @@ Describe 'tests for function expressions' {
...
+    @{ expression = "[intersection(parameters('thirdObject'), parameters('fourthObject'))]"; expected = [pscustomobject]@{ three = 'd' } }
+    @{ expression = "[intersection(parameters('firstArray'), parameters('thirdArray'))]"; expected = @() }
+    @{ expression = "[intersection(parameters('firstObject'), parameters('firstArray'))]"; isError = $true }
+  ) {
+    param($expression, $expected, $isError)
+
+    $config_yaml = @"
+            `$schema: https://aka.ms/dsc/schemas/v3/bundled/config/document.json
+            parameters:
+              firstObject:
```
SteveL-MSFT, line 135: Please add a deep object test, at least one nested object level

dsc/tests/dsc_functions.tests.ps1
```diff
@@ -111,6 +111,76 @@ Describe 'tests for function expressions' {
     }
   }
 
+  It 'intersection function works for: <expression>' -TestCases @(
+    @{ expression = "[intersection(parameters('firstArray'), parameters('secondArray'))]"; expected = @('cd') }
+    @{ expression = "[intersection(parameters('firstObject'), parameters('secondObject'))]"; expected = [pscustomobject]@{ two = 'b' } }
+    @{ expression = "[intersection(parameters('thirdArray'), parameters('fourthArray'))]"; expected = @('ef', 'gh') }
+    @{ expression = "[intersection(parameters('thirdObject'), parameters('fourthObject'))]"; expected = [pscustomobject]@{ three = 'd' } }
+    @{ expression = "[intersection(parameters('firstArray'), parameters('thirdArray'))]"; expected = @() }
+    @{ expression = "[intersection(parameters('firstObject'), parameters('firstArray'))]"; isError = $true }
```
SteveL-MSFT, line 120: Should have a test with at least 3 parameters for both array and object

dsc_lib/src/functions/intersection.rs
```diff
@@ -0,0 +1,155 @@
...
+            remaining_arg_accepted_types: Some(vec![FunctionArgKind::Array, FunctionArgKind::Object]),
+            return_types: vec![FunctionArgKind::Array, FunctionArgKind::Object],
+        }
+    }
+
+    fn invoke(&self, args: &[Value], _context: &Context) -> Result<Value, DscError> {
+        debug!("{}", t!("functions.intersection.invoked"));
+        
+        if args[0].is_array() {
+            let first_array = args[0].as_array().unwrap();
```
SteveL-MSFT: Would be better to write as:

```rust
if let Some(first_array) = args[0].as_array() {
...
}
```

dsc_lib/src/functions/intersection.rs
```diff
@@ -0,0 +1,155 @@
...
+                
+                if found_in_all && !result.contains(item) {
+                    result.push(item.clone());
+                }
+            }
+            
+            return Ok(Value::Array(result));
+        }
+
+        if args[0].is_object() {
```
SteveL-MSFT: Same here:

```rust
if let Some(first_object) = args[0].as_object() {
...
}
```

SteveL-MSFT (changes requested):
I really like your examples in the ref docs as it shows how this can be used in the real world!

SteveL-MSFT (approved):

Issue #1136, opened 2025-09-26
Parser: Found error node parsing function while using user-defined functions


Gijsreyn:
### Prerequisites

- [x] Write a descriptive title.
- [x] Make sure you are able to repro it on the latest version
- [x] Search the existing issues.

### Summary

While writing the reference documentation for #1135, the following example throws an error:

```yaml
$in = @'
$schema: https://aka.ms/dsc/schemas/v3/bundled/config/document.json
functions:
- namespace: Environment
  members:
    getEnvironmentConfig:
      parameters:
      - name: isProduction
        type: bool
      output:
        type: object
        value: "[if(parameters('isProduction'), 
                   createObject('logLevel', 'error', 'debugMode', false()),
                   createObject('logLevel', 'debug', 'debugMode', true()))]"

resources:
- name: Echo production config
  type: Microsoft.DSC.Debug/Echo
  properties:
    output: "[Environment.getEnvironmentConfig(true())]"
- name: Echo development config
  type: Microsoft.DSC.Debug/Echo
  properties:
    output: "[Environment.getEnvironmentConfig(false())]"
'@
dsc config get --input $in
```

### Steps to reproduce

Run the above code snippet from the description

### Expected behavior

```console
results:
- name: Echo production config
  type: Microsoft.DSC.Debug/Echo
  result:
    actualState:
      output:
        logLevel: error
        debugMode: false
- name: Echo development config
  type: Microsoft.DSC.Debug/Echo
  result:
    actualState:
      output:
        logLevel: debug
        debugMode: true
messages: []
hadErrors: false
```

### Actual behavior

```console
2025-09-26T04:52:27.274014Z TRACE dsc_lib::configure: 997: Invoke property expression for output: "[Environment.getEnvironmentConfig(true())]"
2025-09-26T04:52:27.274361Z DEBUG dsc_lib::parser: 48: Parsing statement: [Environment.getEnvironmentConfig(true())]
2025-09-26T04:52:27.274706Z DEBUG dsc_lib::parser: 89: Parsing expression
2025-09-26T04:52:27.275003Z DEBUG dsc_lib::parser::expressions: 43: Parsing function '{Node function (0, 1) - (0, 40)}'
2025-09-26T04:52:27.275529Z ERROR dsc::subcommand: 76: Parser: Found error node parsing function
```

### Error details

```console

```

### Environment data

```PowerShell
Name                           Value
----                           -----
PSVersion                      7.5.3
PSEdition                      Core
[...]

PR #1144, opened 2025-09-27
Fix expression grammar to work with boolean functions

SteveL-MSFT:
# PR Summary

The grammar didn't allow for functions `true()` and `false()` and only expected boolean arguments `true` and `false` so this resulted in a parsing error.  Fix is to update the grammar with precedence for functions and booleans (in that order) so that `true()` and `false()` will be accepted as function names and not treated as booleans and thus invalidating the rest of the expression.

Some of the invalid tests needed to be updated, but all still result in an error as expected.

## PR Context

Fix https://github.com/PowerShell/DSC/issues/1136


SteveL-MSFT:
> I think this LGTM though I'm not a tree-sitter expert and can't explain exactly how this automatically resolves `true()` and `false()` to `true` and `false`. Per the tests, it does. My guess is that it's a recursive evaluation of the `$._booleanLiteral` when it's the `functionName`.

The main change in the grammar to fix this is to ensure that `true` and `false` are both booleans as parameters to functions, but also as `true()` and `false()` functions.  Without this change (even if you set precedence), it would match to the boolean first because it's an exact match compared to the regex and thus never match as a function.  In this case, we have explicitly added `true` and `false` as possible function names (in addition to the regex) so that it can be matched and the precedence (higher number is higher precedence) means it'll be evaluated first as a potential function name before identifying it as a boolean value.  This behavior is specific to tree-sitter generated parser.  Thanks for the extra set of eyes!

PR #1145, opened 2025-09-27
Add `substring()` function

Gijsreyn:
# PR Summary

This pull request adds the `substring()` function, including reference documentation.

## PR Context

Partially addresses #57.

### DIFF
--- a/dsc_lib/src/functions/substring.rs
+++ b/dsc_lib/src/functions/substring.rs
@@ -1,0 +1,225 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+use super::Function;
+use crate::configure::context::Context;
+use crate::functions::{FunctionArgKind, FunctionCategory, FunctionMetadata};
+use crate::DscError;
+use rust_i18n::t;
+use serde_json::Value;
+use tracing::debug;
+
+#[derive(Debug, Default)]
+pub struct Substring {}
+
+impl Function for Substring {
+    fn get_metadata(&self) -> FunctionMetadata {
+        FunctionMetadata {
+            name: "substring".to_string(),
+            description: t!("functions.substring.description").to_string(),
+            category: vec![FunctionCategory::String],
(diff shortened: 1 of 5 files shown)


dsc_lib/src/functions/substring.rs
```diff
@@ -0,0 +1,233 @@
...
+            ));
+        }
+
+        let start_index = usize::try_from(start_index_value).map_err(|_| {
+            DscError::FunctionArg(
+                "substring".to_string(),
+                t!("functions.substring.startIndexValueTooLarge").to_string(),
+            )
+        })?;
+        let string_length = string_to_parse.chars().count();
```
SteveL-MSFT, line 52: Shouldn't `len()` just work here?

```suggestion
        let string_length = string_to_parse.len();
```

dsc_lib/src/functions/substring.rs
```diff
@@ -0,0 +1,233 @@
...
+            }
+
+            length
+        };
+
+        let result: String = string_to_parse
+            .chars()
+            .skip(start_index)
+            .take(length)
+            .collect();
```
SteveL-MSFT, line 94: I think it's simpler to do something like:

```suggestion
        let result: String = string_to_parse[start_index..length].to_string();
```

dsc_lib/src/functions/substring.rs
Gijsreyn: It probably does, but if `len()` was used, Unicode characters are not preserved. 

```rust
let s = "héllo";
println!("len(): {}", s.len());           // 6 bytes (é takes 2 bytes)
println!("chars().count(): {}", s.chars().count()); // 5 characters
```

dsc_lib/src/functions/substring.rs
Gijsreyn: I think that if I were to change to this, it would have the same implication as my earlier comment.

SteveL-MSFT (approved):

PR #1112, opened 2025-09-17
Add copy() and copyIndex() function reference documentation

Gijsreyn:
# PR Summary

This pull request adds the `copy()` and `copyIndex()` reference documentation.

PR #1147, opened 2025-09-30
Reopen add `copy()` and `copyIndex()` function reference documentation

Gijsreyn:
# PR Summary

As PR #1112 was closed by the bot, this has been reopened.

### DIFF
--- a/docs/reference/schemas/config/functions/copy.md
+++ b/docs/reference/schemas/config/functions/copy.md
@@ -1,0 +1,220 @@
+---
+description: Reference for the 'copy' DSC configuration document resource loop
+ms.date:     02/28/2025
+ms.topic:    reference
+title:       copy
+---
+
+# copy
+
+## Synopsis
+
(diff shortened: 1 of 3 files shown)


docs/reference/schemas/config/functions/copy.md
```diff
@@ -0,0 +1,220 @@
...
+```
+
+## Description
+
+The `copy` property enables you to create multiple instances of a resource in a
+DSC configuration. This is the equivalent implementation of the copy
+functionality from Azure Resource Manager (ARM) templates, but without support
+for variables and properties, which will be added in future releases.
+
+When you use `copy` on a resource, DSC creates multiple instances of that
```
SteveL-MSFT, line 29: We should make it clear that `count` can take an expression with an example (probably taken from a parameter is most common use case)

SteveL-MSFT (changes requested):

Issue #1155, opened 2025-10-03
Standized code formatting via `rustfmt`


Gijsreyn:
### Summary of the new feature / enhancement

> As a contributor to the DSC repository,
> I want to have a clear and consistent Rust code style across the project,
> So that:
> - Pull request focuses on behavior changes rather than style arguments.
> - All code "looks like it came from the same author."
> - CI can automatically detect style violations and reject PRs that don't conform
> - New contributors have less friction

Currently, formatting style is ad hoc. This causes friction in reviews, occasional style comment debates, and some churn when refactoring.

I propose we adopt `rustfmt` with a small, curated `rustfmt.toml` config file and enforce it in the CI (separate or Pester tests) with `cargo fmt --check`.

## Proposed formatting policy

1. Use `rustfmt` as the canonical formatter:
   - Developers should run cargo fmt locally (or via IDE) before committing.
   - The project includes a `rustfmt.toml` file with agreed-upon overrides of defaults.
2. Minimal deviations from default style: 
   - Keep most defaults. Only override settings that are meaningful 
   - Example options to override: `max_width`, `imports_granularity`, `reorder_imports`, `wrap_comments`.
 3. Check in CI and locally
    - Either a CI pipeline or Pester test should run:
   
    ```bash
    cargo fmt --all -- --check -- --files-with-diff
    ```
    - If formatting is incorrect, the build should fail and show the diff.
 
4. One-time bulk reformat
   - Upon acceptance, run `cargo fmt` over the existing codebase.
   - After that, all future PRs must remain formatted.
5. Developer tooling / integration
   - Configure popular tooling (VS Code) to auto-run formatting on save using rustfmt
 
6. Versioning and toolchain alignment
   - We pin the `rustfmt` version implicitly by pinning the Rust toolchain version.

## Example `rustfmt.toml`

Below is a suggestion for a minimal config that tweaks a few settings to get started (whilst retaining most of the default style):

```toml
# rustfmt.toml for DSC

# Basic layout
max_width = 100
hard_tabs = false
tab_spaces = 4

# Imports
reorder_imports = true
imports_granularity = "Crate"
group_imports = "StdExternalCrate"

# Comments & docs
wrap_comments = true
format_code_in_doc_comments = true
normalize_doc_attributes = true

# Macro bodies, etc.
format_macro_bodies = true

# Edition (for parsing behavior)
edition = "2021"
```

Rationale for some choices:

[...]

PR #1165, opened 2025-10-07
Enable rustfmt in repo and fix current formatting

SteveL-MSFT:
# PR Summary

Add `rustfmt` under `-clippy`.  There is a new `-FmtFix` switch for `build.ps1` to automatically make formatting fixes, otherwise, CI will fail if rustfmt checks fail.

Used a slightly modified version of the proposed rustfmt.toml, however, many of the options are only available in the nightly channel and not the stable one.  I decided to keep them in the toml but not to use the nightly channel for now (due to being unstable by definition).

All the code changes are just formatting changes.  However, due to the formatting changes, some functions became too long so added `#[allow(clippy::too_many_lines)]` to some functions for now.

Making this a draft for now, the features we need in rustfmt are only available in the nightly toolchain and we don't want to take a dependency on that.  Will revisit when the nightly features show up in stable.

## PR Context

Fix https://github.com/PowerShell/DSC/issues/1155

### DIFF
--- a/rustfmt.toml
+++ b/rustfmt.toml
@@ -17,7 +17,7 @@ format_code_in_doc_comments = true
 normalize_doc_attributes = true
 
 # Macro bodies, etc.
-#format_macro_bodies = true
+format_macro_bodies = false
 
 # Edition (for parsing behavior)
 edition = "2021"


SteveL-MSFT:
It also seems within the Rust community that rustfmt isn't being actively maintained

PR #1161, opened 2025-10-06
(GH-1152, GH-1153) Define cargo workspace and reorganize projects

michaeltlombardi:
# PR Summary

Prior to this change, the project didn't use a cargo workspace. Instead, every crate independently managed its dependencies and we associated crates with Rust Analyzer in VS Code through the configuration settings.

This change:

- Defines a virtual workspace at the repository root, including every crate as a workspace member, defining the release profile, and defining every crate dependency.
- Updates the cargo manifest for every crate to inherit their dependencies from the workspace.
- Replaces the per-crate cargo lock files with the workspace-level lock file. Now dependency management happens with a single manifest and lock file.
- Makes minor changes to the build script to ensure functionality continues as previously defined.

Note that this change on its own doesn't drastically improve build times. For build times to improve substantially, we need to:

1. Restructure the build to compile all crates with a single command. When we use `cargo build` at the repository root or the `--workspace` flag, cargo builds every crate and its dependencies together. Cargo uses the dependency graph to only compile each dependency crate once during the build, instead of recompiling every dependency when we compile on a crate-by-crate basis.
1. Enable the use of caching, like with `sccache`. When we enable cached builds, only crates that have been modified since the last build will be recompiled. This _substantially_ improves build times for local development and when we invoke tests after initial compilation.

However, both of these changes require more substantial modifications to the build script.

To do:

- [x] Verify CI passes for initial commit
- [x] Add alternate build script (need to ensure functionality before removing existing build script)
- [x] Reorganize repository and update build scripts to match

Other changes required as part of this PR:

- Fix a bug in the Windows PowerShell adapter to remove the builtin `PSModulePaths` instead of blanket removing any module path with `powershell` as an ancestor folder.

## PR Context

- #1152
- #1153

PR #1166, opened 2025-10-08
(MAINT) Minimally Parallelize pester tests

michaeltlombardi:
# PR Summary

Prior to this change, the pester tests made up the majority of the build time in CI. This change attempts to speed up the pester tests by parallelizing them roughly by folder:

- The tests in the `dsc` folder run in their own cell.
- The tests in the `adapters` folder run in their own cell.
- The tests in the `extensions` folder run in their own cell.
- The tests in the `resources` folder run in their own cell.

This should substantially reduce the time spent on tests, particularly on Windows, where the CLI tests take much longer.

This change makes minor changes to the new build script and helpers to support targeted test execution.

## PR Context

Depends on #1161

### DIFF
--- a/.github/workflows/rust.new.yml
+++ b/.github/workflows/rust.new.yml
@@ -14,45 +14,133 @@ on:
 env:
   CARGO_TERM_COLOR: always
 
+defaults:
+  run:
+    shell: pwsh
+
 jobs:
-  build-linux-new:
+  linux-build:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v5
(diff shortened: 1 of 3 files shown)


michaeltlombardi:
To see the run times: https://github.com/PowerShell/DSC/actions/runs/18358507549

image

This is a substantial improvement and shows that the majority of the remaining performance to squeeze from CI right now is parallelizing the windows tests for the `dsc` folder. We can probably further decompose the Windows suite, but I'm leaving it as-is for now.

Issue #538, opened 2024-09-05
Canonicalize `dsc schema`  command 


michaeltlombardi:
## User Stories

> As a user, I want the schemas returned from the `dsc schema` command to be canonical, returning the appropriate and fully realized schema for the requested type, so that regardless of whether I obtained the schema from the command or the schema URI, the schema describes the data in the same way.

> As a documentarian, I want to keep the schema definitions more closely aligned to the code without having to do research/update work asynchronously to changes that affect the schema.

> As a developer, I want to ensure that my data types are correctly represented in the canonical schema so I don't have to debug behaviors caused by differences between the published schemas and those used internally.

> As a schema maintainer, I want to generate canonical schemas as close to the source code as possible to ensure they are testable and correctly integrated.

> As an integrating developer, I want to be able to rely on canonical schemas so my own code can use them for processing and validation.

## Current context

Currently, we use [serde (`v1.0.0`)][serde], [schemars (`v0.8.12`)][schemars] and [jsonschema (`v0.18.0`)][jsonschema-rs] for serializing, deserializing, generating, and validating JSON schemas and the associated data. The code currently derives the schemas for the data types with default settings and without additional schema validation. This generates Draft 07 schemas with minimal validation beyond the basic types.

### Publishing

The current published JSON Schemas (as static files served from this repository) are hand-crafted, include documentation, examples, full validation, and VS Code keywords to enable an improved author experience. These schemas are published in three formats:

| Format | Description |
|:------:|:------------|
| Canonical | the schema does not bundle references or include VS Code keywords. |
| Bundled   | The schema bundles references into the `$defs` keyword by ID and doesn't include VS Code keywords. |
| VS Code | The schema bundles references into the `definitions` keyword by ID and includes VS Code keywords. |

The _Canonical_ schemas define discrete schemas for the data models in DSC. They may use the `$ref` keyword to reference other canonical schemas by their `$id`, which is always the URI to the canonical schema file. These schemas are strictly draft 2020-12 compatible and don't include the VS Code keywords.

The _Bundled_ schemas are a convenience. Each of them represents a top-level schema with references, and includes those referenced schemas in the `$defs` keyword, reusing their `$id` as the definition name. Per the JSON Schema spec, this enables consumers to retrieve and process a single document containing that schema without having to fetch the references. This enables us to define smaller discrete canonical schemas while still ensuring users can operate on a single JSON document for validation and inspection.

[...]

PR #1171, opened 2025-10-09
(GH-538) Add `dsc-lib-jsonschema` crate

michaeltlombardi:
# PR Summary

This change begins the schema canonicalization process by adding a new library crate to contain the helpers for defining canonical and enhanced schemas ergonomically.

In this initial phase, the new crate defines:

- The `VSCODE_KEYWORDS` constant, enumerating the keywords that VS Code's JSON language server recognizes.
- The `idiomaticize_string_enum` transformer for munging the generated schemas for string enums with annotation keywords.
- The `idiomaticize_externally_tagged_enum` transformer for munging the generated schemas for externally tagged enums to a more idiomatic and readable representation.

This change uses the new `idiomaticize_*` transformers from the `dsc-lib-jsonschema` crate in the `dsc-lib` crate to ensure that the generated schemas are idiomatic.

## PR Context

- #538

### DIFF
--- a/dsc/tests/dsc_i18n.tests.ps1
+++ b/dsc/tests/dsc_i18n.tests.ps1
@@ -26,23 +26,25 @@ Describe 'Internationalization tests' {
             }
         }
 
+        $patterns = @{
+            t = '(?s)\bt\!\(\s*"(?<key>.*?)".*?\)'
+            panic_t = '(?s)\bpanic_t\!\(\s*"(?<key>.*?)".*?\)'
+            assert_t = '(?s)\bassert_t\!\(\s*.*?,\s*"(?<key>.*?)".*?\)'
+        }
+
         $missing = @()
         Get-ChildItem -Recurse -Path $project -Include *.rs -File | ForEach-Object {
-            # Write-Verbose -Verbose "File: $_"
-            $line = 0
-            Get-Content -Path $_ | ForEach-Object {
-                $line++
-                ($_ | Select-String -Pattern '[^\w]t\!\("(?<key>.*?)".*?\)' -AllMatches).Matches | ForEach-Object {
+            $content = Get-Content -Path $_ -Raw
+            foreach ($pattern in $patterns.keys) {
+                ($content | Select-String -Pattern $patterns[$pattern] -AllMatches).Matches | ForEach-Object {
                     # write-verbose -verbose "Line: $_"
                     if ($null -ne $_) {
                         $key = $_.Groups['key'].Value
                         if ($i18n.ContainsKey($key)) {
                             $i18n[$key] = 1
-                            # write-verbose -verbose "Found on line $line : $key"
                         }
                         else {
                             $missing += $key
-                            # write-verbose -verbose "Missing: $key"
                         }
                     }
                 }


lib/dsc-lib-jsonschema/src/tests/transforms/mod.rs
```diff
@@ -0,0 +1,4 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+//! Unit tests for [`dsc-lib-jsonschema::transforms`]
```
SteveL-MSFT, line 4: Doc examples are run as part of `cargo test` so do we need this placeholder for unittests?

lib/dsc-lib-jsonschema/tests/integration/transforms/idiomaticizing/enums/string_variants.rs
```diff
@@ -0,0 +1,374 @@
...
+//! [`idiomaticize_string_enum`]: crate::transforms::idiomaticize_string_enum
+
+use pretty_assertions::assert_eq as assert_pretty_eq;
+use schemars::{schema_for, JsonSchema, json_schema};
+
+use dsc_lib_jsonschema::transforms::idiomaticize_string_enum;
+
+/// Defines an enum where each variant maps to a string value. This enum includes every
+/// supported documentation keyword for the enum and each variant.
+#[allow(dead_code)]
```
SteveL-MSFT, line 17: It's probably better to have this in a scope that is `#[cfg(test)]` so it's conditionally compiled

lib/dsc-lib-jsonschema/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+serde = { workspace = true }
+serde_json = { workspace = true }
+tracing = { workspace = true }
+
+[dev-dependencies]
+# Helps review complex comparisons, like schemas
+pretty_assertions = { workspace = true }
+
+[lints.clippy]
+pedantic ={ level = "deny" }
```
SteveL-MSFT: Is there a specific pedantic rule(s) you are suppressing?

```suggestion
pedantic = { level = "deny" }
```

lib/dsc-lib-jsonschema/src/tests/transforms/mod.rs
michaeltlombardi: It's not strictly required. The intent here is to add unit tests as they make sense, but these two transforms really only make sense to test at an integration level, which is why I left this file empty.

lib/dsc-lib-jsonschema/Cargo.toml
michaeltlombardi: This is functionally hoisting the deny flag from the CLI to the configuration file. I was trying to get rust analyzer to flag pedantic lints for me without having to separately run clippy.

lib/dsc-lib-jsonschema/tests/integration/transforms/idiomaticizing/enums/string_variants.rs
michaeltlombardi: All of the code in `lib/dsc-lib-jsonschema/tests` is `#[cfg(test)]` scoped. This folder is _only_ compiled for cargo's integration tests, but I also applied the attribute on every module declaration, e.g.

```rust
#[cfg(test)] mod string_variants;
```

SteveL-MSFT (approved):

SteveL-MSFT:
@michaeltlombardi can you address the merge conflicts?

michaeltlombardi:
@SteveL-MSFT it looks like the changes for this PR are already in place because ##1197, which included these changes and built on them, is already merged. Closing this PR as it's no longer required.

PR #1174, opened 2025-10-10
Add `invoke_dsc_config()` MCP tool

Gijsreyn:
# PR Summary

This tool allows an agent to invoke a dsc config directly, including parameters.

## PR Context

Partially addresses #1093.


dsc/src/mcp/invoke_dsc_config.rs
```diff
@@ -0,0 +1,216 @@
...
+    pub async fn invoke_dsc_config(
+        &self,
+        Parameters(InvokeDscConfigRequest {
+            operation,
+            configuration,
+            parameters,
+        }): Parameters<InvokeDscConfigRequest>,
+    ) -> Result<Json<InvokeDscConfigResponse>, McpError> {
+        let result = task::spawn_blocking(move || {
+            let config: Configuration = match serde_json::from_str(&configuration) {
```
SteveL-MSFT: For MCP, I think it might be better to only support YAML so that if it generates a config, it's more human readable

Gijsreyn: Got it. Changed it now to YAML.

SteveL-MSFT (approved):

PR #1176, opened 2025-10-10
Add `uriComponent()` and `uriComponentToString()` functions

Gijsreyn:
# PR Summary

This pull request adds the `uriComponent()` and `uriComponentToString()` functions, along with their reference documentation.

## PR Context

Partially addresses #57.

### DIFF
--- a/dsc/tests/dsc_functions.tests.ps1
+++ b/dsc/tests/dsc_functions.tests.ps1
@@ -903,6 +903,35 @@ Describe 'tests for function expressions' {
     $out.results[0].result.actualState.output | Should -Be $expected
   }
 
+  It 'tryGet() function works for: <expression>' -TestCases @(
+    @{ expression = "[tryGet(createObject('a', 1, 'b', 2), 'a')]"; expected = 1 }
+    @{ expression = "[tryGet(createObject('a', 1, 'b', 2), 'c')]"; expected = $null }
+    @{ expression = "[tryGet(createObject('key', 'value'), 'key')]"; expected = 'value' }
+    @{ expression = "[tryGet(createObject('nested', createObject('x', 10)), 'nested')]"; expected = [pscustomobject]@{ x = 10 } }
+    @{ expression = "[tryGet(createObject('nested', createObject('x', 10)), 'missing')]"; expected = $null }
+    @{ expression = "[tryGet(createArray(1,2,3), 0)]"; expected = 1 }
+    @{ expression = "[tryGet(createArray(1,2,3), 3)]"; expected = $null }
+    @{ expression = "[tryGet(createArray(1,2,3), -3)]"; expected = $null }
+  ) {
+    param($expression, $expected)
+
+    $escapedExpression = $expression -replace "'", "''"
+    $config_yaml = @"
+            `$schema: https://aka.ms/dsc/schemas/v3/bundled/config/document.json
+            resources:
+            - name: Echo
+              type: Microsoft.DSC.Debug/Echo
+              properties:
+                output: '$escapedExpression'
+"@
+    $out = $config_yaml | dsc config get -f - | ConvertFrom-Json
+    if ($expected -is [pscustomobject]) {
+      ($out.results[0].result.actualState.output | Out-String) | Should -BeExactly ($expected | Out-String)
+    } else {
+      $out.results[0].result.actualState.output | Should -BeExactly $expected
+    }
+  }
+
   It 'uriComponent function works for: <testInput>' -TestCases @(
     @{ testInput = 'hello world' }
     @{ testInput = 'hello@example.com' }
@@ -1030,34 +1059,5 @@ Describe 'tests for function expressions' {
 "@
     $out = $config_yaml | dsc config get -f - | ConvertFrom-Json
     $out.results[0].result.actualState.output | Should -BeExactly $expected
-  }
-
-    It 'tryGet() function works for: <expression>' -TestCases @(
-    @{ expression = "[tryGet(createObject('a', 1, 'b', 2), 'a')]"; expected = 1 }
-    @{ expression = "[tryGet(createObject('a', 1, 'b', 2), 'c')]"; expected = $null }
-    @{ expression = "[tryGet(createObject('key', 'value'), 'key')]"; expected = 'value' }
-    @{ expression = "[tryGet(createObject('nested', createObject('x', 10)), 'nested')]"; expected = [pscustomobject]@{ x = 10 } }
-    @{ expression = "[tryGet(createObject('nested', createObject('x', 10)), 'missing')]"; expected = $null }
-    @{ expression = "[tryGet(createArray(1,2,3), 0)]"; expected = 1 }
-    @{ expression = "[tryGet(createArray(1,2,3), 3)]"; expected = $null }
-    @{ expression = "[tryGet(createArray(1,2,3), -3)]"; expected = $null }
-  ) {
-    param($expression, $expected)
-
-    $escapedExpression = $expression -replace "'", "''"
-    $config_yaml = @"
-            `$schema: https://aka.ms/dsc/schemas/v3/bundled/config/document.json
-            resources:
-            - name: Echo
-              type: Microsoft.DSC.Debug/Echo
-              properties:
-                output: '$escapedExpression'
-"@
-    $out = $config_yaml | dsc config get -f - | ConvertFrom-Json
-    if ($expected -is [pscustomobject]) {
-      ($out.results[0].result.actualState.output | Out-String) | Should -BeExactly ($expected | Out-String)
-    } else {
-      $out.results[0].result.actualState.output | Should -BeExactly $expected
-    }
   }
 }


Gijsreyn:
Not sure what totally went wrong here, but the older [build failed](https://github.com/PowerShell/DSC/actions/runs/18405350611), while the new one succeeded at first.

lib/dsc-lib/src/functions/uri.rs
```diff
@@ -0,0 +1,193 @@
...
+
+        let result = combine_uri(base_uri, relative_uri);
+        Ok(Value::String(result))
+    }
+}
+
+fn combine_uri(base_uri: &str, relative_uri: &str) -> String {
+    if base_uri.is_empty() {
+        return relative_uri.to_string();
+    }
```
SteveL-MSFT: Is this behavior correct?  `uri()` is supposed to return an absolute URI so a relative URI wouldn't be absolute.  I think this is an error case.  The .NET `Uri` class also won't work with an empty base Uri.

lib/dsc-lib/src/functions/uri.rs
```diff
@@ -0,0 +1,193 @@
...
+
+    let base_ends_with_slash = base_uri.ends_with('/');
+    let relative_starts_with_slash = relative_uri.starts_with('/');
+
+    // Case 1: baseUri ends with trailing slash
+    if base_ends_with_slash {
+        if relative_starts_with_slash {
+            // Combine trailing and leading slash into one
+            return format!("{base_uri}{}", &relative_uri[1..]);
+        }
```
SteveL-MSFT: What if the relative_uri is `///foo`?  The ARM code just uses the .NET `[Uri]::TryCreate()` method, so perhaps in the tests, you can just use that for the expected result.

lib/dsc-lib/src/functions/uri_component_to_string.rs
```diff
@@ -0,0 +1,171 @@
...
+        let result = percent_decode_uri_component(uri_encoded_string)?;
+        Ok(Value::String(result))
+    }
+}
+
+/// Decodes a percent-encoded URI component string.
+/// 
+/// Decodes percent-encoded sequences (e.g., %20 to space, %40 to @).
+/// This is the inverse operation of `percent_encode_uri_component`.
+fn percent_decode_uri_component(input: &str) -> Result<String, DscError> {
```
SteveL-MSFT: May be better to use https://docs.rs/urlencoding/latest/urlencoding/ crate for encoding and decoding

Gijsreyn: I hadn't thought about that one. Added!

lib/dsc-lib/src/functions/uri.rs
Gijsreyn: Added test cases and updated the docs.

dsc/tests/dsc_functions.tests.ps1
```diff
@@ -858,4 +859,116 @@ Describe 'tests for function expressions' {
     $out = $config_yaml | dsc config get -f - | ConvertFrom-Json
     $out.results[0].result.actualState.output | Should -Be $expected
   }
+
+  It 'uri function works for: <expression>' -TestCases @(
+    @{ expression = "[uri('https://example.com/', 'path/file.html')]"; expected = 'https://example.com/path/file.html' }
```
SteveL-MSFT: I think for `uri` specific tests, you should just have it like:

```powershell
@{ base = 'https://example.com/'; relative = 'path/file.html' }
```

then to validate, you'd just:

```powershell
$expected = ""
$success = [uri]::tryfrom([uri]$base, [uri]$relative)
$success | Should -BeTrue
$out.results[0].result.actualState.output | Should -BeExactly $expected
```

dsc/tests/dsc_functions.tests.ps1
```diff
@@ -858,4 +859,116 @@ Describe 'tests for function expressions' {
...
+              properties:
+                output: '$escapedExpression'
+"@
+    $out = $config_yaml | dsc config get -f - | ConvertFrom-Json
+    $out.results[0].result.actualState.output | Should -Be $expected
+  }
+
+  It 'uri function error handling: <expression>' -TestCases @(
+    @{ expression = "[uri('', 'path')]" ; expectedError = 'baseUri parameter cannot be empty' }
+    @{ expression = "[uri('https://example.com/', '///foo')]" ; expectedError = 'Invalid URI|invalid sequence' }
```
SteveL-MSFT: Should include also where base is not an absolute URI like: `example.com` which should fail.

dsc/tests/dsc_functions.tests.ps1
```diff
@@ -860,25 +860,29 @@ Describe 'tests for function expressions' {
...
   ) {
-    param($expression, $expected)
+    param($base, $relative)
+    
+    $expected = ''
+    $success = [uri]::TryCreate([uri]$base, [uri]$relative, [ref]$expected)
+    $success | Should -BeTrue
+    
+    $expression = "[uri('$base', '$relative')]"
     $escapedExpression = $expression -replace "'", "''"
```
SteveL-MSFT, line 873: This line isn't needed I think

lib/dsc-lib/locales/en-us.toml
```diff
@@ -531,6 +531,7 @@ invoked = "uniqueString function"
 [functions.uri]
 description = "Creates an absolute URI by combining the baseUri and the relativeUri string"
 emptyBaseUri = "The baseUri parameter cannot be empty. An absolute URI requires a valid base URI."
+notAbsoluteUri = "The baseUri must be an absolute URI (must include a scheme such as http:// or https://)."
```
SteveL-MSFT: I think it would be better to use `file://` and `https://` as examples

dsc/tests/dsc_functions.tests.ps1
Gijsreyn: Didn't totally get you, but if you were talking about the escapedExpression one, it is:

image

SteveL-MSFT: I see the misunderstanding here.  You should remove this line and change 936 to use double quotes.  It's functionally equivalent, but easier for future folks to understand what's happening.

lib/dsc-lib/src/functions/uri.rs
```diff
@@ -0,0 +1,275 @@
...
+    
+    if !base_uri.contains("://") && !base_uri.starts_with("//") {
+        return Err(DscError::Parser(t!("functions.uri.notAbsoluteUri").to_string()));
+    }
+    
+    if relative_uri.is_empty() {
+        return Ok(base_uri.to_string());
+    }
+
+    if relative_uri.starts_with("///") {
```
SteveL-MSFT: This isn't correct if we're mimicking .NET behavior:

```powershell
[uri]::TryCreate([uri]'file://1/2/3/4/5', [uri]'///baz', [ref]$out)
```

Even this is allowed:

```powershell
[uri]::TryCreate([uri]'file://1/2/3/4/5', [uri]'//////////baz', [ref]$out)
```

The [.NET Code](https://github.com/dotnet/runtime/blob/b28b882d2cfa23452fab461b88e0e4e87b52f263/src/libraries/System.Private.Uri/src/System/UriExt.cs#L312) looks non-trivial in terms of the rules.  Perhaps there's a URL crate that handles this already?

Gijsreyn: You're correct on the second one, but I don't think the first one is correct. If I'm running:

```powershell
$out = $null
$success = [uri]::TryCreate([uri]'https://example.com/', [uri]'///foo', [ref]$out)
Write-Host "  Success: $success"
Write-Host "  Result: $($out.AbsoluteUri)"
```

The result is:

```plaintext
Success: False
  Result:
```

Or did I misunderstand you?

dsc/tests/dsc_functions.tests.ps1
```diff
@@ -901,4 +902,64 @@ Describe 'tests for function expressions' {
     $out = $config_yaml | dsc config get -f - | ConvertFrom-Json
     $out.results[0].result.actualState.output | Should -Be $expected
   }
-}
+
+  It 'uriComponent function works for: <expression>' -TestCases @(
+    @{ expression = "[uriComponent('hello world')]"; expected = 'hello%20world' }
```
SteveL-MSFT: I think these tests should just check against `[System.Web.HttpUtility]::UrlEncode()`

dsc/tests/dsc_functions.tests.ps1
```diff
@@ -901,4 +902,64 @@ Describe 'tests for function expressions' {
...
+              type: Microsoft.DSC.Debug/Echo
+              properties:
+                output: '$escapedExpression'
+"@
+    $out = $config_yaml | dsc config get -f - | ConvertFrom-Json
+    $out.results[0].result.actualState.output | Should -Be $expected
+  }
+
+  It 'uriComponentToString function works for: <expression>' -TestCases @(
+    @{ expression = "[uriComponentToString('hello%20world')]"; expected = 'hello world' }
```
SteveL-MSFT: Same here but `[System.Web.HttpUtility]::UrlDecode()`

dsc/tests/dsc_functions.tests.ps1
```diff
@@ -901,4 +902,64 @@ Describe 'tests for function expressions' {
...
+  ) {
+    param($expression, $expected)
+    $escapedExpression = $expression -replace "'", "''"
+    $config_yaml = @"
+            `$schema: https://aka.ms/dsc/schemas/v3/bundled/config/document.json
+            resources:
+            - name: Echo
+              type: Microsoft.DSC.Debug/Echo
+              properties:
+                output: '$escapedExpression'
```
SteveL-MSFT, line 925:
```suggestion
                output: "$escapedExpression"
```

dsc/tests/dsc_functions.tests.ps1
```diff
@@ -901,4 +902,64 @@ Describe 'tests for function expressions' {
...
+    @{ expression = "[uriComponent('name=John Doe&age=30')]"; expected = 'name%3DJohn%20Doe%26age%3D30' }
+    @{ expression = "[uriComponent('/path/to/my file.txt')]"; expected = '%2Fpath%2Fto%2Fmy%20file.txt' }
+    @{ expression = "[uriComponent(concat('hello', ' ', 'world'))]"; expected = 'hello%20world' }
+    @{ expression = "[uriComponent('user+tag@example.com')]"; expected = 'user%2Btag%40example.com' }
+    @{ expression = "[uriComponent('1234567890')]"; expected = '1234567890' }
+    @{ expression = "[uriComponent('100%')]"; expected = '100%25' }
+    @{ expression = "[uriComponent(' ')]"; expected = '%20' }
+  ) {
+    param($expression, $expected)
+    $escapedExpression = $expression -replace "'", "''"
```
SteveL-MSFT, line 918:
```suggestion
```

dsc/tests/dsc_functions.tests.ps1
Gijsreyn: I think we shouldn't fully rely on that. See results in the image:

image

The code used in the PowerShell script:

```powershell
# Test if System.Web assembly is needed
Add-Type -AssemblyName System.Web

Write-Host "Test 1: UrlEncode"
$input = "hello world"
$result = [System.Web.HttpUtility]::UrlEncode($input)
Write-Host "  Input: '$input'"
Write-Host "  Result: '$result'"

Write-Host "`nTest 2: UrlDecode"
$input = "hello%20world"
$result = [System.Web.HttpUtility]::UrlDecode($input)
Write-Host "  Input: '$input'"
Write-Host "  Result: '$result'"

Write-Host "`nTest 3: Special characters"
$input = "café"
$result = [System.Web.HttpUtility]::UrlEncode($input)
Write-Host "  Input: '$input'"
Write-Host "  Result: '$result'"
```

For Bicep:

```bicep
// Basic hello world example
output encodedHelloWorld string = uriComponent('hello world')

// More examples with different strings
output encodedCafe string = uriComponent('café')
output encodedEmail string = uriComponent('hello@example.com')
output encodedSpecialChars string = uriComponent(':/?#[]@!$&()*+,;=')
output encodedPath string = uriComponent('/path/to/my file.txt')

// Decoding examples using uriComponentToString()
output decodedHelloWorld string = uriComponentToString('hello%20world')
output decodedCafe string = uriComponentToString('caf%C3%A9')
output decodedSpecialChars string = uriComponentToString('%3A%2F%3F%23%5B%5D%40%21%24%26%28%29%2A%2B%2C%3B%3D')
```

This means the current implementation is "okay".

dsc/tests/dsc_functions.tests.ps1
Gijsreyn: See comment above.

dsc/tests/dsc_functions.tests.ps1
Gijsreyn: Removed!

dsc/tests/dsc_functions.tests.ps1
SteveL-MSFT: I checked the ARM code and they actually use `[uri]::EscapeDataString()` and `UnescapeDataString()`

Gijsreyn: Thanks, just what I needed. Changed the tests.

dsc/tests/dsc_functions.tests.ps1
```diff
@@ -2,903 +2,1033 @@
...
-  It 'function works: <text>' -TestCases @(
-    @{ text = "[concat('a', 'b')]"; expected = 'ab' }
-    @{ text = "[concat('a', 'b', 'c')]"; expected = 'abc' }
-    @{ text = "[concat('a', concat('b', 'c'))]"; expected = 'abc' }
-    @{ text = "[base64('ab')]"; expected = 'YWI=' }
-    @{ text = "[base64(concat('a','b'))]"; expected = 'YWI=' }
-    @{ text = "[base64(base64(concat('a','b')))]"; expected = 'WVdJPQ==' }
-  ) {
-    param($text, $expected)
+#   It 'function works: <text>' -TestCases @(
```
SteveL-MSFT: Why are these commented out?

Gijsreyn:
@SteveL-MSFT - I might have known only the method on `-TagFilter`. What I know mostly do is comment out the `launch.json`, comment out the test, and put a breakpoint. I'll take a look at your suggestion next time.

SteveL-MSFT:
> @SteveL-MSFT - I might have known only the method on `-TagFilter`. What I know mostly do is comment out the `launch.json`, comment out the test, and put a breakpoint. I'll take a look at your suggestion next time.

Here's an example:

```powershell
invoke-pester -Output Detailed ./dsc/tests/dsc_expressions.tests.ps1 -FullNameFilter '*Accessors work*'
```

Issue #1177, opened 2025-10-10
Copy (loop) function array/object and conditionals


ThomasNieto:
### Summary of the new feature / enhancement

In [BICEP](https://learn.microsoft.com/en-us/azure/azure-resource-manager/bicep/loops) users can loop through a collection or object (dictionary). The [loop condition](https://learn.microsoft.com/en-us/azure/azure-resource-manager/bicep/loops#loop-with-condition) would also be very useful.

These kind of loops are a very common pattern to loop through a collection or object (dictionary) in order to set the same value.

SteveL-MSFT:
@ThomasNieto can you show an example of what the Bicep looks like and the resulting JSON template?  The doc link only shows the Bicep side.

Gijsreyn:
Just chipping in my two cents here. If this is what you meant Thomas:

```bicep
targetScope = 'desiredStateConfiguration'

// Parameter object containing registry values to manage
param registryValues object = {
  items: [
    {
      name: 'ExampleString'
      keyPath: 'HKCU\\DscExamples'
      valueName: 'StringValue'
      valueData: {
        String: 'Hello from DSC'
      }
    }
    {
      name: 'ExampleDWord'
      keyPath: 'HKCU\\DscExamples'
      valueName: 'DWordValue'
      valueData: {
        DWord: 42
      }
    }
    {
      name: 'ExampleExpandString'
      keyPath: 'HKCU\\DscExamples\\Config'
      valueName: 'PathValue'
      valueData: {
        ExpandString: '%USERPROFILE%\\Documents'
      }
    }
    {
      name: 'ExampleMultiString'
      keyPath: 'HKCU\\DscExamples\\Config'
      valueName: 'ListValue'
      valueData: {
        MultiString: [
          'Item1'
          'Item2'
          'Item3'
        ]
      }
    }
  ]
}

// Loop over the registry values and create a resource for each
resource registryConfigs 'Microsoft.Windows/Registry@2025-08-27' = [for item in registryValues.items: {
  name: item.name
  properties: {
    keyPath: item.keyPath
    valueName: item.valueName
    valueData: item.valueData
    _exist: true
  }
}]
```

It converts it to JSON like this:

```json
{
  "$schema": "https://aka.ms/dsc/schemas/v3/bundled/config/document.json",
  "contentVersion": "1.0.0.0",
  "metadata": {
    "_generator": {
      "name": "bicep",
      "version": "0.38.3.11034",
[...]

PR #1178, opened 2025-10-11
Add `items()` function

Gijsreyn:
# PR Summary

As preparation for #1177, this pull request adds the `items()` function, along with its reference documentation.

## PR Context

Partially addresses #57.

> [!NOTE]
> Already included example in the docs to use with loops (even properties are not supported yet).

### DIFF
--- a/lib/dsc-lib/locales/en-us.toml
+++ b/lib/dsc-lib/locales/en-us.toml
@@ -368,7 +368,7 @@ invoked = "intersection function"
 invalidArgType = "All arguments must either be arrays or objects"
 
 [functions.items]
-description = "Converts a dictionary object to an array"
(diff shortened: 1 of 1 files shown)


lib/dsc-lib/locales/en-us.toml
```diff
@@ -367,6 +367,10 @@ description = "Returns a single array or object with the common elements from th
 invoked = "intersection function"
 invalidArgType = "All arguments must either be arrays or objects"
 
+[functions.items]
+description = "Converts a dictionary object to an array"
```
SteveL-MSFT: I think the description should be expanded for non-technical users:

```suggestion
description = "Converts a dictionary object to an array of objects with key and value properties"
```

PR #1182, opened 2025-10-13
(MAINT) Define tasks for build and test

michaeltlombardi:
# PR Summary

This change adds VS Code tasks for building and testing the projects for convenience.

## PR Context

Wanted to make it a little more convenient for contributors to build and test the project locally. This change doesn't replace invoking the build script from the terminal, but it does simplify invoking the build script as an editor task.

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -1,0 +1,84 @@
+{
+    // See https://go.microsoft.com/fwlink/?LinkId=733558
+    // for the documentation about the tasks.json format
+    "version": "2.0.0",
+    "tasks": [
+        {
+            "label": "DSC: Build",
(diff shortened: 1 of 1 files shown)


SteveL-MSFT:
@michaeltlombardi at what point do we have just a single `build.ps1` again?

michaeltlombardi:
@SteveL-MSFT once we confirm that the new build script works well enough for packaging. I'm confident in its use for building and testing, but I lack the local setup to confirm packaging aside from zip and tgz. I can double check and fix those up today, but the MSIX and macOS packages are something I can't currently validate. After we've confirmed that, we can switch to the new script entirely for GHA and Azure Pipelines.

Issue #1160, opened 2025-10-04
Enable multiple resources in single resource manifest


SteveL-MSFT:
### Summary of the new feature / enhancement

Been a common ask by partner teams. 

### Proposed technical implementation details (optional)

I think the simplest solution is to allow a manifest to contain an array of resource manifests

michaeltlombardi:
Thinking about this, I would prefer that we have a (lightweight) wrapper object, like:

```jsonc
{
    "$schema": "https://aka.ms/dsc/schemas/v3/resource/manifest.json",
    "resources": [
        {
            "type": "Example/Foo"
            // elided fields
        },
        {
            "type": "Example/Bar"
            // elided fields
        },
        {
            "type": "Example/Baz"
            // elided fields
        }
    ]
}
```

Where we update the resource schema to use the `oneOf` keyword pointing to the correct schema. Primarily, my thinking is:

1. I don't think it makes sense to support different resources in the same manifest with different schema versions.
2. Getting the enhanced authoring schema to deal nicely with the array-of-resources model will be painful compared to the wrapper.
3. We could _optionally_ consider (later, not required for initial implementation) supporting field definitions at the top level defining defaults for lower-level, which would reduce duplication (such as setting author, version, etc).

SteveL-MSFT:
Current proposal, have a new `.dsc.resourcelist.json` (or yaml) that is an array of manifests.  A `resourcelist` manifest will have a single property `resources` which is an array of nested manifest objects.

PR #1187, opened 2025-10-16
Enable a single '.dsc.manifests.json' file to contain many resource/extension manifests

SteveL-MSFT:
# PR Summary

This change allows for a single `.dsc.manifests.json` (or yaml) file to contain multiple resource or extension manifests so you no longer need a separate file per resource/extension.

No new test added as the `dsctest` resource now has a single manifest for all of its resources.

This turned out to be more complicated than I expected due to a known limitation in `serde` for handling `HashMap<i32, String>` which was also embedded in an `enum` which is now the case.  To workaround this, the schema now has `exitCodes` be `HashMap<String, String>` which is what it actually is in the JSON and at runtime, we parse the key to be a i32 otherwise return an error.

Also some general code cleanup in the modified areas putting things in alpha order and modifying some code so that you get a specific error if there's invalid JSON/YAML instead of a generic error.

```json
{
  "resources": [ ],
  "extensions": [ ]
}
```

## PR Context

FIx https://github.com/PowerShell/DSC/issues/1160

### DIFF
--- a/tools/dsctest/.project.data.json
+++ b/tools/dsctest/.project.data.json
@@ -8,7 +8,7 @@
     ],
     "CopyFiles": {
         "All": [
-            "dsctrest.dsc.manifests.json"
+            "dsctest.dsc.manifests.json"
         ]
     }
 }


lib/dsc-lib/src/discovery/command_discovery.rs
```diff
@@ -31,6 +29,7 @@ use crate::util::get_setting;
 use crate::util::get_exe_path;
 
 const DSC_RESOURCE_EXTENSIONS: [&str; 3] = [".dsc.resource.json", ".dsc.resource.yaml", ".dsc.resource.yml"];
+const DSC_RESOURCE_LIST_EXTENSIONS: [&str; 3] = [".dsc.resourcelist.json", ".dsc.resourcelist.yaml", ".dsc.resourcelist.yml"];
```
michaeltlombardi: I would prefer `.dsc.resource.list.*` for readability, if possible.

lib/dsc-lib/src/dscresources/resource_manifest.rs
```diff
@@ -20,6 +20,12 @@ pub enum Kind {
     Resource,
 }
 
+#[derive(Debug, Clone, PartialEq, Deserialize, Serialize, JsonSchema)]
+pub struct ResourceManifestList {
+    /// The list of resource manifests.
+    pub resources: Vec<ResourceManifest>,
+}
```
michaeltlombardi: I think we should _probably_ hoist the `$schema` from the individual resource manifests to this level, so we can appropriately apply the schema to the document. Ideally, we shouldn't _allow_ the `$schema` for individual resources in this document - otherwise they could be validating against different schemas and that gets messy quickly. That _would_ gunk up the simplicity of this implementation though.

lib/dsc-lib/src/discovery/command_discovery.rs
SteveL-MSFT: My preference is that we have too many "extensions" for the extension already.  Only resource authors and the engine cares about these particular ones.

lib/dsc-lib/src/dscresources/resource_manifest.rs
SteveL-MSFT: Hypothetically, each individual manifest COULD be from a different schema particularly if a package contains multiple resources and the older ones don't get updated, but newer ones get added.

michaeltlombardi: That still leaves the problem of validating the top-level document - we _can_ just validate it against "latest" and so long as this document _never changes_ and every contained resource defines `$schema` we're okay.

lib/dsc-lib/src/discovery/command_discovery.rs
michaeltlombardi: I was mostly thinking that the `<kind>.list.<extension>` is more readable and will be useful if we eventually support extension list manifest and the hypothetical adapted resource list manifests.

SteveL-MSFT: Was thinking that there's no reason to have separate lists for adapters, resources, extensions, etc... we could just have a single format that's an array of manifests.  Perhaps `.dsc.manifests.json`?

lib/dsc-lib/src/dscresources/resource_manifest.rs
SteveL-MSFT: My thinking is that if this is ever just a container schema, then it should never change.

SteveL-MSFT:
I'm going to update this PR based on recent discussion:

- a single container format allowing any type of valid DSC manifest in the array
- file to be called `.dsc.manifestlist.json` that contains a property `manifests` which is an array of manifests

ThomasNieto:
Do we need to have list in the name of the file?

What about just having it be plural? `dsc.manifests.json`

Actually looks like that was the name of the file in the comment thread not sure why the new comment has a different name.

> Was thinking that there's no reason to have separate lists for adapters, resources, extensions, etc... we could just have a single format that's an array of manifests. Perhaps .dsc.manifests.json?

SteveL-MSFT:
@ThomasNieto the original addition for `list` was to make it more obvious between `resource` and `resources`, but since we've moved on from that, I think just `manifests` is fine.  Will change.

lib/dsc-lib/src/discovery/command_discovery.rs
```diff
@@ -30,15 +28,28 @@ use which::which;
...
+#[serde(untagged)]
+pub enum Manifest {
+    Resource(Box<ResourceManifest>),
+    Extension(Box<ExtensionManifest>),
+}
+
+#[derive(Debug, Clone, PartialEq, Deserialize, Serialize, JsonSchema)]
+pub struct ManifestList {
+    pub manifests: Vec<Manifest>,
+}
```
michaeltlombardi: Can we rework this to have separate properties in the top-level document for resource and extension manifests to enable simpler consumption by higher order tools which may not care about subsets of valid manifests (like ignoring non-resource extensions for Ansible)?

That would still enable us to include new manifests in the future but make reviewing and working with the multi-manifest document a little easier and more ergonomic.

lib/dsc-lib/src/discovery/command_discovery.rs
```diff
@@ -31,15 +29,22 @@ use which::which;
...
 static ADAPTERS: LazyLock<RwLock<BTreeMap<String, Vec<DscResource>>>> = LazyLock::new(|| RwLock::new(BTreeMap::new()));
 static RESOURCES: LazyLock<RwLock<BTreeMap<String, Vec<DscResource>>>> = LazyLock::new(|| RwLock::new(BTreeMap::new()));
 static EXTENSIONS: LazyLock<RwLock<BTreeMap<String, DscExtension>>> = LazyLock::new(|| RwLock::new(BTreeMap::new()));
 static ADAPTED_RESOURCES: LazyLock<RwLock<BTreeMap<String, Vec<DscResource>>>> = LazyLock::new(|| RwLock::new(BTreeMap::new()));
 
+#[derive(Debug, Clone, PartialEq, Deserialize, Serialize, JsonSchema)]
+pub struct ManifestList {
+    pub resources: Option<Vec<ResourceManifest>>,
+    pub extensions: Option<Vec<ExtensionManifest>>,
+}
```
michaeltlombardi: Should we skip serializing when the value isn't defined? Can we simplify accessing these values by defining it as:

```suggestion
#[derive(Debug, Clone, PartialEq, Deserialize, Serialize, JsonSchema)]
pub struct ManifestList {
    #[serde(default, skip_serializing_if = "Vec::is_empty")]
    pub resources: Vec<ResourceManifest>,
    #[serde(default, skip_serializing_if = "Vec::is_empty")]
    pub extensions: Vec<ExtensionManifest>,
}

impl Default for ManifestList {
    fn default() -> Self {
        Self {
            resources: vec![],
            extensions: vec![],
        }
    }
}
```

SteveL-MSFT: I don't think we would actually ever need to serialize this.  So perhaps we should remove that from this struct.  Removing the `Option` here doesn't really save anything since Rust has `if let Some()` syntax.  We should separately review all the macro usage and remove ones that aren't actually being used as it's generating code that isn't used and just bloating the exe.

michaeltlombardi: I can see a use case for serializing this from the rust development kit (create a manifest list for your project), but we can always add that code in if and when it becomes useful.

Issue #1186, opened 2025-10-14
For `resourceId()` function, urlencode the resource name


SteveL-MSFT:
### Summary of the new feature / enhancement

Currently, to determine if a resource is unique, the `resourceId()` concatenates the type and name with a colon in between. 

### Proposed technical implementation details (optional)

It would be better to urlencode the name so that special characters don't present a problem and don't require restrictions on the resource name.

This is something that only happens inside the engine (unless the user explicitly retrieves the `resourceId()`, but it's only intended to be a unique identifier, so isn't a breaking change.

PR #1188, opened 2025-10-16
Encode name part of `resourceId()` function to allow name to accept any characters

SteveL-MSFT:
# PR Summary

For DSC, the `name` property for resources is intended to be a friendly name and not an identity, so we don't have the same restrictions as ARM on what is allowed.  Change is to have `resourceId()` url encode the name so that arbitrary characters will be encoded and thus allow any character.

## PR Context

Fix https://github.com/PowerShell/DSC/issues/1186

### DIFF
--- a/lib/dsc-lib/locales/en-us.toml
+++ b/lib/dsc-lib/locales/en-us.toml
@@ -476,8 +476,6 @@ unavailableInUserFunction = "The 'reference()' function is not available in user
 [functions.resourceId]
 description = "Constructs a resource ID from the given type and name"
 incorrectTypeFormat = "Type argument must contain exactly one slash"
-invalidFirstArgType = "Invalid argument type for first parameter"
-invalidSecondArgType = "Invalid argument type for second parameter"
 
 [functions.secret]
 description = "Retrieves a secret from a vault"


lib/dsc-lib/src/functions/resource_id.rs
```diff
@@ -28,32 +29,15 @@ impl Function for ResourceId {
...
-        let resource_name = &args[1];
-        if let Some(value) = resource_name.as_str() {
-            if value.contains('/') {
-                return Err(DscError::Function("resourceId".to_string(), t!("functions.resourceId.incorrectNameFormat").to_string()));
-            }
-
-            result.push_str(value);
-        } else {
-            return Err(DscError::Parser(t!("functions.resourceId.invalidSecondArgType").to_string()));
+        let resource_type = &args[0].as_str().unwrap();
```
tgauth, line 33: is `unwrap()` sufficient here in place of the previous error handling? same comment for line 39 below

SteveL-MSFT: Yes, current recommendation for functions is to just `unwrap()`.  It should never panic since DSC engine has already performed arg validation before the function is invoked.  Basically having all the explicit checks in the function results in errors that never get raised.

PR #1190, opened 2025-10-17
Add `json()` function

Gijsreyn:
# PR Summary

This pull request adds the `json()` function, including reference documentation.

## PR Context

Partially addresses #57.

### DIFF
--- a/lib/dsc-lib/src/functions/json.rs
+++ b/lib/dsc-lib/src/functions/json.rs
@@ -27,9 +27,7 @@ impl Function for Json {
     }
 
     fn invoke(&self, args: &[Value], _context: &Context) -> Result<Value, DscError> {
-        let Some(json_str) = args[0].as_str() else {
-            return Err(DscError::Parser(t!("functions.json.notString").to_string()));
-        };
+        let json_str = args[0].as_str().unwrap();
 
         match serde_json::from_str(json_str) {
             Ok(value) => Ok(value),
--- a/lib/dsc-lib/locales/en-us.toml
+++ b/lib/dsc-lib/locales/en-us.toml
@@ -386,7 +386,6 @@ invalidObjectElement = "Array elements cannot be objects"
 
 [functions.json]
 description = "Converts a valid JSON string into a JSON data type"
-notString = "Argument must be a string"
 invalidJson = "Invalid JSON string"
 
 [functions.lastIndexOf]


dsc/tests/dsc_functions.tests.ps1
```diff
@@ -1060,4 +1060,58 @@ Describe 'tests for function expressions' {
...
+    @{ testInput = "('$(ConvertTo-Json -Compress -InputObject @(1,2,3))')[2]"; expected = 3 }
+    @{ testInput = "('$(ConvertTo-Json -Compress -InputObject 'hello')')"; expected = 'hello' }
+    @{ testInput = "('$(ConvertTo-Json -Compress -InputObject 42)')"; expected = 42 }
+    @{ testInput = "('$(ConvertTo-Json -Compress -InputObject $true)')"; expected = $true }
+    @{ testInput = "('$(ConvertTo-Json -Compress -InputObject $false)')"; expected = $false }
+    @{ testInput = "('$(ConvertTo-Json -Compress -InputObject $null)')"; expected = $null }
+    @{ testInput = "('$(ConvertTo-Json -Compress -InputObject @{ users = @( @{ name = 'Alice' }, @{ name = 'Bob' } ) })').users[0].name"; expected = 'Alice' }
+    @{ testInput = "('$(ConvertTo-Json -Compress -InputObject @{ users = @( @{ name = 'Alice' }, @{ name = 'Bob' } ) })').users[1].name"; expected = 'Bob' }
+    @{ testInput = "('$(ConvertTo-Json -Compress -InputObject @{ key = 'value' })').key"; expected = 'value' }
+    @{ testInput = "('$(ConvertTo-Json -Compress -InputObject @{ nested = @{ value = 123 } })').nested.value"; expected = 123 }
```
SteveL-MSFT: I think it'll be simpler and easier to read/update if you just have the hashtable in the testcases and convert them to JSON after line 1079

dsc/tests/dsc_functions.tests.ps1
```diff
@@ -1060,4 +1060,58 @@ Describe 'tests for function expressions' {
...
+    $out = $config_yaml | dsc config get -f - | ConvertFrom-Json
+    $out.results[0].result.actualState.output | Should -Be $expected
+  }
+
+  It 'json() error handling: <expression>' -TestCases @(
+    @{ expression = "[json('not valid json')]"; expectedError = 'Invalid JSON string' }
+    @{ expression = "[json('{""key"":""value""')]"; expectedError = 'Invalid JSON string' }
+    @{ expression = "[json('')]"; expectedError = 'Invalid JSON string' }
+    @{ expression = "[json('{incomplete')]"; expectedError = 'Invalid JSON string' }
+    @{ expression = "[json('[1,2,')]"; expectedError = 'Invalid JSON string' }
```
SteveL-MSFT: If they all have the same error, you don't need to have `expectedError` in the test case

lib/dsc-lib/src/functions/json.rs
```diff
@@ -0,0 +1,154 @@
...
+            ],
+            remaining_arg_accepted_types: None,
+            return_types: vec![FunctionArgKind::Object, FunctionArgKind::Array, FunctionArgKind::String, FunctionArgKind::Number, FunctionArgKind::Boolean],
+        }
+    }
+
+    fn invoke(&self, args: &[Value], _context: &Context) -> Result<Value, DscError> {
+        let Some(json_str) = args[0].as_str() else {
+            return Err(DscError::Parser(t!("functions.json.notString").to_string()));
+        };
```
SteveL-MSFT: Since this function ONLY accepts a string and the input param type is pre-validated before calling this implementation, that error message should never be actually used.  I think it's simpler to just use:

```rust
let json_str = args[0].as_str().unwrap();
```

Since validation already happened, this should never panic.

dsc/tests/dsc_functions.tests.ps1
Gijsreyn: I did it slightly differently now, also with the accessor.

PR #1191, opened 2025-10-17
Add `loadTextContent()` and `loadFileAsBase64()` functions

### LINKED ISSUE #1189: Functions for filesystem operations

Looks like Bicep has added a bunch of new functions to work with the filesystem which would be useful for DSC.  https://github.com/Azure/bicep/pull/2501

_No response_

Gijsreyn:
# PR Summary

This pull request adds the `loadTextContent()` and `loadFileAsBase64()` functions, along with their corresponding reference documentation.

## PR Context

Fix #1189.

### DIFF
--- a/lib/dsc-lib/src/functions/load_text_content.rs
+++ b/lib/dsc-lib/src/functions/load_text_content.rs
@@ -15,9 +15,6 @@ const MAX_CONTENT_SIZE: usize = 131_072; // 131072 characters
 #[derive(Debug, Default)]
 pub struct LoadTextContent {}
 
-#[derive(Debug, Default)]
-pub struct LoadTextContent {}
-
 impl super::Function for LoadTextContent {
     fn get_metadata(&self) -> FunctionMetadata {
         FunctionMetadata {


docs/reference/schemas/config/functions/loadFileAsBase64.md
```diff
@@ -0,0 +1,234 @@
...
+
+- The file path can be absolute or relative to the configuration document.
+- The maximum allowed file size is 96 KB (98,304 bytes).
+- Both text and binary files are supported.
+- The encoding is suitable for embedding binary data in text-based formats.
+
+> [!IMPORTANT]
+> This function loads file content at compile time. The file must be accessible
+> when the configuration is compiled and executed. For runtime file operations, use an
+> appropriate resource.
```
SteveL-MSFT, line 31: This is true for bicep, but not true for DSC as we don't have a "compile" step and this function will be used at runtime.  However, this does bring up an interesting problem.  If this function only exists in bicep and not ARM, then a DSC config using this function would be executed at bicep compile time and thus DSC will never see the function call.

I wonder if it makes sense then to call this something else to avoid the confusion as the DSC function would execute at runtime.  This is a good topic to discuss at next WG meeting.

SteveL-MSFT (changes requested):

docs/reference/schemas/config/functions/loadFileAsBase64.md
Gijsreyn: Let's keep the PR open and put the other functions on hold as mentioned in the issue.

theJasonHelmick:
Thank you @Gijsreyn for the contribution and the thoughtful implementation! After discussion in DSC Working Group , we’ve decided to close this PR as declined.

- The proposed functions mirror Bicep’s loadTextContent / loadFileAsBase64, which are compile-time conveniences in Bicep. They don’t exist at ARM runtime, and in DSC they would execute at deployment/runtime, which changes the security and validation model.
- Allowing file content to be pulled in at runtime introduces ambiguity and risk (e.g., path tampering, non-determinism) unless we design clear guarantees and constraints for DSC’s local/remote file and scheme support.
- Name collisions with Bicep semantics would be confusing. If DSC offers similar capabilities in the future, we should use DSC-specific names and semantics and ship them as part of an explicit design.

Thank you for your time and continued contributions to the project!

PR #1192, opened 2025-10-18
Re-open add `uri()` function

Gijsreyn:
# PR Summary

This pull request adds the `uri()` function with the url crate. It also contains the reference documentation.

There's one rough edge case that didn't work, which was on IPv6. See image below.

image

If this needs to be addressed, I need to debug it further. The script used to compare .NET vs `dsc` is attached.

[test_uri_comparison.txt](https://github.com/user-attachments/files/22981606/test_uri_comparison.txt)

## PR Context

Partially addresses #57.

### DIFF
--- a/lib/dsc-lib/src/functions/uri.rs
+++ b/lib/dsc-lib/src/functions/uri.rs
@@ -8,7 +8,6 @@ use rust_i18n::t;
 use serde_json::Value;
 use super::Function;
 use url::Url;
-use std::net::Ipv6Addr;
 
 #[derive(Debug, Default)]
 pub struct Uri {}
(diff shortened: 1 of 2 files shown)


SteveL-MSFT:
@Gijsreyn so in your examples, those ipv6 addresses are the same, just one is long form and one is short form.  In this case, you could use `[IPAddress]::parse().tostring()` to get the short form and compare that.

lib/dsc-lib/src/functions/uri.rs
```diff
@@ -0,0 +1,220 @@
...
+        }
+
+        if let Some(uri_without_slashes) = relative_uri.strip_prefix("//") {
+            let scheme = base.scheme();
+            
+            if uri_without_slashes.contains('/') {
+                return Ok(Value::String(format!("{scheme}:{relative_uri}")));
+            }
+            return Ok(Value::String(format!("{scheme}:{relative_uri}/")));
+        }
```
SteveL-MSFT: Does `join()` not already cover this?

Gijsreyn: Great catch! Modified it.

lib/dsc-lib/src/functions/uri.rs
```diff
@@ -48,19 +49,24 @@ impl Function for Uri {
...
+        let final_result = if let Some(url::Host::Ipv6(ipv6_addr)) = result.host() {
+            let segments = ipv6_addr.segments();
+            let expanded = format!("{:04X}:{:04X}:{:04X}:{:04X}:{:04X}:{:04X}:{:04X}:{:04X}",
+                segments[0], segments[1], segments[2], segments[3],
+                segments[4], segments[5], segments[6], segments[7]
+            );
+            
+            let addr = Ipv6Addr::from(segments);
+            let compressed = addr.to_string();
+            result.to_string().replace(&compressed, &expanded)
```
SteveL-MSFT: I don't think you need to have ipv6 expanded, I think compressed is more desirable.  In this case, I think you could handle it in the tests and it's ok for DSC and .NET to have different representations as the functionality is the same

SteveL-MSFT (approved):

PR #1197, opened 2025-10-21
(GH-538) Define schema extension methods

michaeltlombardi:
# PR Summary

This change defines the `SchemaUtilityExtensions` trait in the `dsc-lib-jsonschema` crate to provide useful shorthand methods for the `schemars::Schema` type when munging and transforming schemas.

This change:

- Defines and implements the trait, providing documentation for every new method.
- Adds tests for the behaviors of every extension method.
- Depends on #1171

## PR Context

While the trait includes some methods I expect to be useful, it doesn't try to fully map methods for retrieving every keyword. Instead, it primarily simplifies working with the following keywords:

- `$id` to retrieve and set the identifier for a schema.
- `$defs` to retrieve and munge referenced subschemas.
- `properties` to retrieve and munge property definitions.

The trait _does_ define generic methods for retrieving keywords with an expected type, like `get_keyword_as_object()` and `get_keyword_as_str`. These are convenience methods to replace always calling code like the following when you want to retrieve a keyword with a known type:

```rust
let ref schema = json_schema!({
    "title": "Schema title"
});
if let Some(title) = schema.get("title").and_then(|v| v.as_str()) {
    println!("Schema title is {title}");
}
```

So you can instead do:

```rust
let ref schema = json_schema!({
    "title": "Schema title"
});
if let Some(title) = schema.get_keyword_as_str("title") {
    println!("Schema title is {title}");
}
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -699,6 +699,7 @@ dependencies = [
  "serde",
  "serde_json",
  "tracing",
+ "url",
 ]
 
 [[package]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -198,7 +198,7 @@ tree-sitter-rust = { version = "0.24" }
 utfx = { version = "0.1" }
 # dsc-lib
 uuid = { version = "1.18", features = ["v4"] }
-# dsc-lib
+# dsc-lib, dsc-lib-jsonschema
 url = { version = "2.5" }
 # dsc-lib
 urlencoding = { version = "2.1" }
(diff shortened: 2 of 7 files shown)


michaeltlombardi:
After #1171 is merged, I'll rebase this PR prior to final review and merge.

SteveL-MSFT:
@michaeltlombardi can you rebase off main so I can see just the delta?

michaeltlombardi:
@SteveL-MSFT - I don't think so, because these changes build on the `dsc-lib-jsonschema` crate.

But all the changes for _this_ PR (on top of #1171) are in this commit:

https://github.com/PowerShell/DSC/pull/1197/commits/970dd4efb98067875ed980166e7af82c675235d7

PR #1210, opened 2025-10-25
Add support for resources to be represented as hash map as part of ARM Language 2.0

### LINKED ISSUE #1193: Add support for Bicep language version 2.0

`languageVersion` needs to be added to the config as a string
`resources` needs to support symbolic names

_No response_

andyleejordan:
Resolves #1193 (at least what we know is required experimentally).

Okay so I tried a bunch of different ways and settled on this being the simplest approach. In short, a custom serde deserializer function let's us take `resources` as either an array or an object and produce what the existing implementation expects: a `Vec<Resource>` such that after the initial deserialization, nothing else needs to change. This means that later ordering should work (though when I tested `dependsOn` in Bicep, I got a different error).

I can at least now pass through this Bicep directly to DSC and it gets to the point that it attempts to execute the resource:

```bicep
extension dsc
targetScope = 'desiredStateConfiguration'

resource myEcho 'Microsoft.DSC.Debug/Echo@1.0.0' = {output: 'Hello world!'}
```

which produces this ARM/JSON:

```json
{
  "languageVersion": "2.0",
  "contentVersion": "1.0.0.0",
  "metadata": {
    "_generator": {
      "name": "bicep",
      "version": "0.38.33.27573",
      "templateHash": "9007474339958309780"
    }
  },
  "imports": {
    "dsc": {
      "provider": "DesiredStateConfiguration",
      "version": "0.1.0"
    }
  },
  "resources": {
    "myEcho": {
      "import": "dsc",
      "type": "Microsoft.DSC.Debug/Echo@1.0.0",
      "properties": {
        "output": "Hello world!"
      }
    }
  }
}
```

Note that there is also a slightly different variation which uses `extension(s)` instead of `import(s)` with a particular experimental feature flag that, as of right now, is necessary for the end-to-end demo to work. Those are also added and ignored. See https://github.com/Azure/bicep/pull/18365

Perhaps we should gate this behind a feature-flag, but between this and publishing a basic version of the Bicep extension to an MCR, we can resume demos but with fancy IntelliSense and working versions!

For the fields noted as irrelevant, I sure would like to use `#[serde(skip)]` but it is [incompatible](https://github.com/serde-rs/serde/issues/2121) with `deny_unknown_fields`. Perhaps we should revisit that Serde flag, possibly using the linked workaround with `IgnoredAny` and warnings instead of failures.

### DIFF
--- a/lib/dsc-lib/src/configure/config_doc.rs
+++ b/lib/dsc-lib/src/configure/config_doc.rs
@@ -528,6 +528,66 @@ mod test {
     }
 
     #[test]
+    fn test_invalid_resource_field_in_array() {
+        let config_json = r#"{
+            "resources": [
+                {
+                    "invalidField": "someValue"
+                }
+            ]
+        }"#;
+
+        let result: Result<Configuration, _> = serde_json::from_str(config_json);
+        assert!(result.is_err());
+        let err = result.unwrap_err().to_string();
+        assert!(err.starts_with("unknown field `invalidField`, expected one of `condition`, `type`,"));
+    }
+
+    #[test]
+    fn test_invalid_resource_field_in_object() {
+        let config_json = r#"{
+            "resources": {
+                "someResource": {
+                    "invalidField": "someValue"
+                }
+            }
+        }"#;
+
+        let result: Result<Configuration, _> = serde_json::from_str(config_json);
+        assert!(result.is_err());
+        let err = result.unwrap_err().to_string();
+        assert!(err.starts_with("unknown field `invalidField`, expected one of `condition`, `type`,"));
+    }
+
+    #[test]
+    fn test_invalid_resource_type_in_array() {
+        let config_json = r#"{
+            "resources": [
+                "invalidType"
+            ]
+        }"#;
+
+        let result: Result<Configuration, _> = serde_json::from_str(config_json);
+        assert!(result.is_err());
+        let err = result.unwrap_err().to_string();
+        assert!(err.contains("expected struct Resource"));
+    }
+
+    #[test]
+    fn test_invalid_resource_type_in_object() {
+        let config_json = r#"{
+            "resources": {
+                "someResource": "invalidType"
+            }
+        }"#;
+
+        let result: Result<Configuration, _> = serde_json::from_str(config_json);
+        assert!(result.is_err());
+        let err = result.unwrap_err().to_string();
+        assert!(err.contains("expected struct Resource"));
+    }
+
+    #[test]
     fn test_resources_as_array() {
         let config_json = r#"{
             "$schema": "https://aka.ms/dsc/schemas/v3/bundled/config/document.json",


lib/dsc-lib/src/configure/config_doc.rs
```diff
@@ -466,4 +511,69 @@ mod test {
 
         assert!(result.is_ok());
     }
+
+    #[test]
+    fn test_deserialize_resources_array() {
+        let config_json = r#"{
+            "$schema": "https://raw.githubusercontent.com/PowerShell/DSC/main/schemas/2024/04/config/document.json",
```
Gijsreyn: Shouldn't the schema be: `https://aka.ms/dsc/schemas/v3/bundled/config/document.json`?

lib/dsc-lib/src/configure/config_doc.rs
```diff
@@ -466,4 +511,69 @@ mod test {
...
+        let config: Configuration = serde_json::from_str(config_json).unwrap();
+        assert_eq!(config.resources.len(), 2);
+        assert_eq!(config.resources[0].name, "resource1");
+        assert_eq!(config.resources[1].name, "resource2");
+    }
+
+    #[test]
+    fn test_deserialize_resources_object() {
+        let config_json = r#"{
+            "$schema": "https://raw.githubusercontent.com/PowerShell/DSC/main/schemas/2024/04/config/document.json",
```
Gijsreyn: Same here?

lib/dsc-lib/src/configure/config_doc.rs
```diff
@@ -169,9 +174,42 @@ pub struct Configuration {
     pub outputs: Option<HashMap<String, Output>>,
     #[serde(skip_serializing_if = "Option::is_none")]
     pub parameters: Option<HashMap<String, Parameter>>,
+    #[serde(deserialize_with = "deserialize_resources")]
     pub resources: Vec<Resource>,
     #[serde(skip_serializing_if = "Option::is_none")]
     pub variables: Option<Map<String, Value>>,
+    /// Irrelevant Bicep metadata from using the extension
```
SteveL-MSFT, line 181: Is this documented somewhere?

lib/dsc-lib/src/configure/config_doc.rs
```diff
@@ -169,9 +174,42 @@ pub struct Configuration {
...
+/// of resources (where the keys are symbolic names) as found in ARMv2 back to a
+/// vector, so the rest of this codebase can remain untouched.
+fn deserialize_resources<'de, D>(deserializer: D) -> Result<Vec<Resource>, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    let value = Value::deserialize(deserializer)?;
+
+    match value {
+        Value::Array(resources) => {
```
SteveL-MSFT, line 198: Do you need to validate that the items in the array are of type `Resource`?

lib/dsc-lib/src/configure/config_doc.rs
```diff
@@ -169,9 +174,42 @@ pub struct Configuration {
...
+            resources.into_iter()
+                .map(|resource| serde_json::from_value(resource).map_err(serde::de::Error::custom))
+                .collect()
+        }
+        Value::Object(resources) => {
+            resources.into_iter()
+                .map(|(name, resource)| {
+                    let mut resource: Resource = serde_json::from_value(resource).map_err(serde::de::Error::custom)?;
+                    // TODO: Decide what to do if resource.name is already set.
+                    resource.name = name;
```
SteveL-MSFT: Looking at the ARM docs https://learn.microsoft.com/en-us/azure/azure-resource-manager/templates/resource-declaration#use-symbolic-name, I don't think you should just clobber the name with the symbolic name.  Instead, looking at https://learn.microsoft.com/en-us/azure/azure-resource-manager/templates/resource-dependency, it appears that symbolic names can be used in `dependsOn` in addition to `resourceId()` in the classical sense.  So I think what we may need to eventually (this can be a follow-up PR) do is maintain a list of symbolic names and for resolving dependencies allow for either symbolic name or resourceid().

For this PR, I would just change the `TODO` comment to reflect what I said above.

lib/dsc-lib/src/configure/config_doc.rs
```diff
@@ -466,4 +511,69 @@ mod test {
 
         assert!(result.is_ok());
     }
+
+    #[test]
+    fn test_deserialize_resources_array() {
```
SteveL-MSFT: It's ok to keep a few unit tests here, but would prefer to also add some Pester tests.

lib/dsc-lib/src/configure/config_doc.rs
andyleejordan: Yes, these are generated and I'm probably going to delete them now that I've been pointed at the Pester tests.

lib/dsc-lib/src/configure/config_doc.rs
SteveL-MSFT: Seems like it might be coming from https://learn.microsoft.com/en-us/azure/azure-resource-manager/bicep/bicep-import, but they don't show what the resulting ARM JSON looks like

lib/dsc-lib/src/configure/config_doc.rs
andyleejordan: I don't think so. If they're not, we'd be in the exact same situation as the existing code, the deserialization will fail and pass up the error.

lib/dsc-lib/src/configure/config_doc.rs
andyleejordan: I put an example of the resulting ARM in the PR description. Since using the DSC extension for Bicep results in it importing the extension, this metadata gets added to the output (along with the other fields).

lib/dsc-lib/src/configure/config_doc.rs
```diff
@@ -316,6 +354,7 @@ pub struct Resource {
     #[serde(skip_serializing_if = "Option::is_none", rename = "apiVersion")]
     pub api_version: Option<String>,
     /// A friendly name for the resource instance
+    #[serde(default)]
     pub name: String, // friendly unique instance name
```
andyleejordan, line 369: By the way, this is saying that if the `name` field doesn't exist (such as in the case of ARMv2 where symbolic names are used instead) that it just be an empty string.

lib/dsc-lib/src/configure/config_doc.rs
SteveL-MSFT: I missed that :)

Seems unfortunate Bicep created a new property for this rather than sticking it into metadata

lib/dsc-lib/src/configure/config_doc.rs
SteveL-MSFT: It would be good to just have a Pester test case cover this then

lib/dsc-lib/src/configure/config_doc.rs
SteveL-MSFT: In ARM, keep in mind that `name` is an identity of a ARM resource, so it's not the friendly name usage we have in DSC.  The symbolic name is more like the friendly name (although with character restrictions).

lib/dsc-lib/src/configure/config_doc.rs
andyleejordan: A couple more showed up during further testing with an experimental feature turned on that we might want to use later, and that for now works around https://github.com/Azure/bicep/pull/18365

lib/dsc-lib/src/configure/config_doc.rs
andyleejordan: Done.

lib/dsc-lib/src/configure/config_doc.rs
andyleejordan: I'm not entirely sure I'm following. In the way that DSC expects to have its `Vec<Resource>` it only has support for `name` (from the name property). Before the ARMv2 update, this came from `resource <the name> <the type@ the version>`, and mapped directly to `name` in the generated JSON. The deserialization happening above implements the exact same behavior. There is not (yet) any other support for symbolic names in DSC.

lib/dsc-lib/src/configure/config_doc.rs
andyleejordan: Whoops, comment on wrong thread. Replying to this thread: I'm not following at all. What would we be testing? That the user hand-crafted a bad JSON file with data that doesn't deserialize to a `Resource` in the array? The behavior of DSC has not changed in this case.

andyleejordan:
@SteveL-MSFT Pester test added for the latest ARM generated by Bicep. Note that https://github.com/Azure/bicep/pull/18365 will be required to use "just" the `desiredStateConfiguration` Bicep experimental feature. In which case the extensions/extension fields are replaced by imports/import.

lib/dsc-lib/src/configure/config_doc.rs
andyleejordan: Done.

andyleejordan:
Just a heads up that as noted in the PR, Bicep has and will likely continue to add new fields (especially when using experimental features) and since we're setting `#[serde(deny_unknown_fields)]` we're allowing that to **break** us rather than warn us. I highly recommend we switch from that to an ignore and warn approach instead. /cc @michaeltlombardi

lib/dsc-lib/src/configure/config_doc.rs
SteveL-MSFT: Validating the error message we expect and not that it would fail at deployment time

lib/dsc-lib/src/configure/config_doc.rs
SteveL-MSFT: I'm saying that "symbolic name" != "name".  So the two can be different.  In DSC, if someone hand crafted the YAML w/ 2.0, the symbolic name could be short hand while the name is verbose, for example.

SteveL-MSFT:
@andyleejordan I'm in agreement that we should relax the strict validation, but then we'd need to warn about unknown new properties.  That can be a separate PR, can you open an issue for this?

lib/dsc-lib/src/configure/config_doc.rs
andyleejordan: I get that, but in DSC this field is documented as "/// A friendly name for the resource instance."

I guess I wasn't expecting users to start writing YAML/JSON using the ARMv2 way of structs instead of a list of resources, since they're writing _DSC_ not ARM, so I was focusing on just Bicep output. It should work, but yeah if both these fields are defined, right now the symbolic name will be lost. I added a note that when we fix `dependsOn` we may need to track both if that's a scenario we want to support.

lib/dsc-lib/src/configure/config_doc.rs
andyleejordan: I added a suite of tests validating that the error messages for invalid data where a `Resource` is expected is the same, and manually tested that on `main` to verify that behavior hasn't changed. (This is kind of just testing that Serde is still working the same way, but now we know experimentally.)

andyleejordan:
CI is failing because the PowerShell Gallery is down due to the Azure Front Door outage.

SteveL-MSFT (approved):

PR #1212, opened 2025-10-28
Add support for invoking property expressions in resource configuration

Gijsreyn:
# PR Summary

This pull request implements support for evaluating expressions in resource properties during copy operations. Previously, while `copy` count could use expressions, properties within copied resources were not evaluated per iteration, preventing users from using functions like `copyIndex()` within property values.

## PR Context

Fix #1179.

Users can now use:

```yaml
resources:
- name: "[format('Server-{0}', copyIndex())]"
  copy:
    name: testLoop
    count: 3
  properties:
    output: "[format('Port-{0}', copyIndex())]"
```

### DIFF
--- a/dsc/tests/dsc_copy.tests.ps1
+++ b/dsc/tests/dsc_copy.tests.ps1
@@ -263,4 +263,75 @@ resources:
         $out.results[3].name | Should -Be 'Server-3'
         $out.results[3].result.actualState.output | Should -Be 'Hello'
     }
+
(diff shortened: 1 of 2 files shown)


emelybeaute-ops (commented):
[Plan_Meta_Ads_Ciluxe.csv](https://github.com/user-attachments/files/23191947/Plan_Meta_Ads_Ciluxe.csv)

PR #1214, opened 2025-10-28
Fix typo Select-Obect -> Select-Object from commit 91a6293…

danellisuk:
I had the error:

_"The term 'Select-Obect' is not recognized as a name of a cmdlet, function, script file, or executable program."_

So this is just a fix for what appears to be a standard typo.


emelybeaute-ops (approved):

SteveL-MSFT:
Thanks for the PR, but I didn't see this was already opened and fixed this via https://github.com/PowerShell/DSC/pull/1216/files#diff-5536a551448f8fd773c3cb53fb0b9034d4ee374fde5a18725a2f61d235ba72e4R53

PR #1215, opened 2025-10-28
Add parameter merging functionality with inline precedence

### LINKED ISSUE #1213: Inline support with parameter file

> As a configuration author,
> I want to provide both inline parameters and parameters file together,
> So during a run, I can dynamically pass in values.

Currently, when ever you try to pass inline parameters including a parameter file, it throws an error:

error: the argument '--parameters-file <PARAMETERS_FILE>' cannot be used with '--parameters <PARAMETERS>'

This is not helpful were you dynamically fetch values during runtime and then pass it along to the configuration document.

Allow both inline values and parameters file together. Taking a look at Bicep's parameter precedence:

_You can use inline parameters and a local parameters file in the same deployment operation. For example, you can specify some values in the local parameters file and add other values inline during deployment. If you provide values for a parameter in both the local parameters file and inline, the inline value takes precedence._

Gijsreyn:
# PR Summary

This PR implements support for using `--parameters` and `--parameters-file` together, with inline parameters taking precedence over file parameters, matching Azure ARM template behavior.

## PR Context

Fix #1213


dsc/src/main.rs
```diff
@@ -54,16 +54,63 @@ fn main() {
...
+                                        error!("{}: {err}", t!("util.invalidUtf8"));
+                                        exit(EXIT_INVALID_INPUT);
+                                    }
+                                }
+                            },
+                            Err(err) => {
+                                error!("{}: {err}", t!("util.failedToReadStdin"));
+                                exit(EXIT_INVALID_INPUT);
+                            }
+                        }
```
SteveL-MSFT: Reading from STDIN is duplicated here and lines below.  I think instead of this approach, you can always assume to merge and sometimes file will be empty or inline will be empty.

dsc/src/util.rs
```diff
@@ -582,3 +578,87 @@ pub fn in_desired_state(test_result: &ResourceTestResult) -> bool {
...
+        }
+    };
+
+    // Both must be objects to merge
+    let Some(mut file_obj) = file_value.as_object().cloned() else {
+        return Err(DscError::Parser("File parameters must be a JSON object".to_string()));
+    };
+    
+    let Some(inline_obj) = inline_value.as_object() else {
+        return Err(DscError::Parser("Inline parameters must be a JSON object".to_string()));
```
SteveL-MSFT: Error messages should be in the locale file, but also this error should just say `must be an object` since it could be YAML

dsc/src/util.rs
```diff
@@ -582,3 +578,87 @@ pub fn in_desired_state(test_result: &ResourceTestResult) -> bool {
...
+        Ok(json) => json,
+        Err(_) => {
+            // YAML
+            match serde_yaml::from_str::<serde_yaml::Value>(inline_params) {
+                Ok(yaml) => serde_json::to_value(yaml)?,
+                Err(err) => {
+                    return Err(DscError::Parser(format!("Failed to parse inline parameters: {err}")));
+                }
+            }
+        }
```
SteveL-MSFT: This sequence is often repeated, in dsc_lib we have `parse_input_to_json()` although it returns a String and not a Value.  You could add the same util function in this file and have it return a Value.

dsc/src/util.rs
```diff
@@ -582,3 +578,87 @@ pub fn in_desired_state(test_result: &ResourceTestResult) -> bool {
...
+            }
+            file_obj.insert("parameters".to_string(), Value::Object(file_params_obj));
+        } else {
+            // If one is not an object, inline takes precedence
+            file_obj.insert("parameters".to_string(), inline_params_value.clone());
+        }
+    } else if let Some(inline_params_value) = inline_obj.get("parameters") {
+        // Only inline has parameters
+        file_obj.insert("parameters".to_string(), inline_params_value.clone());
+    }
```
SteveL-MSFT: A bit worried how this will work when we fix the other issue to support ARM style parameters files.  I think it would be better to simply convert the file_params to a hashtable, the inline_params to a hashtable and layer the inline over the file one.  Then if we support different formats it's simply a function to convert it to a hashtable.

Gijsreyn:
@SteveL-MSFT - thanks for the feedback. I have (tried) incorperating your feedback points. Can you please take another stab at it?

PR #1224, opened 2025-10-31
Enable resource manifests to have relative paths

SteveL-MSFT:
# PR Summary

Hit a partner scenario where they don't want to have all of their assets in `PATH`, so this change enables a resource manifest to specify the executable with a relative path so that only the resource manifest is found in `PATH`.  This change modifies both the verification of the exe and also the invocation of the exe.  If the specified executable has a relative path, then the path is resolved with the directory of the location of the resource manifest.

The major change here is that the `verify_executable()` helper needs the current directory to use a new helper `canonicalize_which()` that checks if the specified exe in the manifest exists or is a relative path.  If relative, it canonicalizes the joined path with the current directory and also checks on Windows that if the exe doesn't have an extension (since it could be `.cmd`, `.bat`, etc...), then it adds `.exe` which is required for the `canonicalize()` function which fails if the resulting path (with extension) doesn't exist.

Some minor code cleanup with `std::fs` API usage to make it consistent.


DHowett:
Tested with `winget configure --processor-path` and directly through `dsc resource {list,get}`, found to be passing!

lib/dsc-lib/src/util.rs
```diff
@@ -232,6 +233,29 @@ pub fn resource_id(type_name: &str, name: &str) -> String {
     result
 }
 
+pub fn canonicalize_which(executable: &str, cwd: Option<&str>) -> Result<String, DscError> {
+    // Use PathBuf to handle path separators robustly
+    let mut executable_path = PathBuf::from(executable);
+    if cfg!(target_os = "windows") && executable_path.extension().is_none() {
+        executable_path.set_extension("exe");
```
andyleejordan, line 240: We should document that we're only going to check for '.exe' if no extension was supplied. This is different than what I believe to be Windows' standard behavior that checks for all of '.cmd', '.bat', variations etc.

lib/dsc-lib/src/util.rs
```diff
@@ -232,6 +233,29 @@ pub fn resource_id(type_name: &str, name: &str) -> String {
     result
 }
 
+pub fn canonicalize_which(executable: &str, cwd: Option<&str>) -> Result<String, DscError> {
+    // Use PathBuf to handle path separators robustly
```
andyleejordan, line 237: 👍

lib/dsc-lib/src/discovery/command_discovery.rs
```diff
@@ -711,38 +710,38 @@ fn load_resource_manifest(path: &Path, manifest: &ResourceManifest) -> Result<Ds
 
     let mut capabilities: Vec<Capability> = vec![];
     if let Some(get) = &manifest.get {
```
andyleejordan, line 712: Unrelated but I wonder if we could write a macro to de-duplicate all this code.

lib/dsc-lib/src/util.rs
```diff
@@ -232,6 +233,29 @@ pub fn resource_id(type_name: &str, name: &str) -> String {
...
+    let mut executable_path = PathBuf::from(executable);
+    if cfg!(target_os = "windows") && executable_path.extension().is_none() {
+        executable_path.set_extension("exe");
+    }
+    let mut executable = executable_path.to_string_lossy().to_string();
```
andyleejordan: What's the necessity of `to_string_lossy().to_string()` here (and in `verify_executable`)? It's going to drop Unicode characters, which are usually valid in paths on modern systems.

lib/dsc-lib/src/util.rs
```diff
@@ -232,6 +233,29 @@ pub fn resource_id(type_name: &str, name: &str) -> String {
...
+    if which(&executable).is_err() && !Path::new(&executable).is_absolute() {
```
andyleejordan: Since `PathBuf` derefs to `Path`, did we need to go through the string conversion? Actually, since it has `is_absolute` did we need to ditech `executable_path` at all? It must have to do with the lossy conversion. I'll go look at `which`.

andyleejordan: The `which` crate takes a `PathBuf` directly so I don't understand why we're intentionally doing the lossy conversion.

lib/dsc-lib/src/util.rs
```diff
@@ -232,6 +233,29 @@ pub fn resource_id(type_name: &str, name: &str) -> String {
...
+        if let Some(cwd) = cwd {
+            let cwd_path = Path::new(cwd);
+            match canonicalize(cwd_path.join(&executable)) {
```
andyleejordan: Wait, why aren't we just using `which_in`? https://docs.rs/which/latest/which/fn.which_in.html

> Find binary_name in the path list paths, using cwd to resolve relative paths.

That's exactly what this PR sets out to do, no? If so, let's use it if there's no reason not to!

andyleejordan (changes requested):
I'm on vacation but can check back on Tuesday to review again 👍

lib/dsc-lib/src/util.rs
andyleejordan: Because we're _passing_ in a folder to use as `cwd`...but yeah no, that's the third argument, it can be passed:

```rust
pub fn which_in<T, U, V>(
    binary_name: T,
    paths: Option<U>,
    cwd: V,
) -> Result<PathBuf>
where
    T: AsRef<OsStr>,
    U: AsRef<OsStr>,
    V: AsRef<Path>,
```

It is 11 pm and I was traveling today, I could definitely be wrong!

lib/dsc-lib/src/util.rs
SteveL-MSFT: This was detailed in the PR description

lib/dsc-lib/src/discovery/command_discovery.rs
SteveL-MSFT: Open a new issue :)

lib/dsc-lib/src/util.rs
SteveL-MSFT: That's a good question.  I guess I do it out of habit when rendering to console.  Let me try to see if it works with it removed.

lib/dsc-lib/src/util.rs
SteveL-MSFT: Now that you brought this up, the `is_absolute()` call isn't needed here since: `c:\foo` is absolute, but is missing the extension.  So that check can just be removed.  I'll check if we can skip some of the calls.

lib/dsc-lib/src/util.rs
SteveL-MSFT: Wasn't aware of that one, let me see if it does what we need.

SteveL-MSFT: `which_in()` needs to separate out the exe, rel path, and cwd, so I don't think it saves anything but probably adds more code than using `canonical()` to resolve relative path.

lib/dsc-lib/src/util.rs
SteveL-MSFT: So `PathBuf` doesn't implement `Display` which means `to_string()` doesn't work which is why `to_string_lossy()` needs to be called first since paths are being passed around as strings.  `to_string_lossy()` returns a `Cow` (clone-on-write) which is why `to_string()` needs to be called on that to get an actual string.

It probably makes sense eventually to update the code so that paths are always `PathBuf`, but that's a big change for another PR.

andyleejordan: > It probably makes sense eventually to update the code so that paths are always PathBuf, but that's a big change for another PR.

Agreed, because passing them as strings going through lossy conversions is introducing bugs if we get more complicated paths. Works in 90%+ use cases, but we just need one username with a Unicode character that gets dropped to break everything.

lib/dsc-lib/src/util.rs
andyleejordan: It is, but PRs aren't permanent nor are they documentation; just calling it out for our docs in the next stable release.

lib/dsc-lib/src/discovery/command_discovery.rs
andyleejordan: No worries, it's more a curiosity than an issue!

lib/dsc-lib/src/util.rs
```diff
@@ -232,6 +233,25 @@ pub fn resource_id(type_name: &str, name: &str) -> String {
...
+    // Use PathBuf to handle path separators robustly
+    let mut executable_path = PathBuf::from(executable);
+    if cfg!(target_os = "windows") && executable_path.extension().is_none() {
+        executable_path.set_extension("exe");
+    }
+    if which(executable).is_err() {
+        if let Some(cwd) = cwd {
+            let cwd_path = Path::new(cwd);
+            if let Ok(canonical_path) = canonicalize(cwd_path.join(&executable_path)) {
+                return Ok(canonical_path.to_string_lossy().to_string());
```
andyleejordan, line 246: I double checked that this is checking for existence, which it is, but have we considered the ramifications of the conversion to extended length paths it performs on Windows:

> On Windows, this converts the path to use [extended length path](https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file) syntax, which allows your program to use longer path names, but means you can only join backslash-delimited paths to it, and it may be incompatible with other applications (if passed to the application on the command-line, or written to a file another application may read).

The docs aren't entirely clear on this but I _think_ they're implying a conversion to UNC, meaning always prepended with `\\` (we had to do this with the port of Mesos' stout C lib to Windows, and it's annoying to say the least):

> A UNC name of any format, which always start with two backslash characters ("\\"). For more information, see the next section.

lib/dsc-lib/src/util.rs
```diff
@@ -232,6 +233,25 @@ pub fn resource_id(type_name: &str, name: &str) -> String {
     result
 }
 
+pub fn canonicalize_which(executable: &str, cwd: Option<&str>) -> Result<String, DscError> {
+    // Use PathBuf to handle path separators robustly
+    let mut executable_path = PathBuf::from(executable);
+    if cfg!(target_os = "windows") && executable_path.extension().is_none() {
+        executable_path.set_extension("exe");
+    }
+    if which(executable).is_err() {
```
andyleejordan, line 242: I did pull some examples and I think it may still be cleaner, and could avoid the UNC issue?

> `which_in(argv0, Some(self.dirs().path_env()), env::current_dir()?)`

> `if let (Some(path), Some(exe_filename)) = (env::var_os("PATH"), exe.file_name()) {`
> `   match which_in(exe_filename, Some(&path), &cwd) {`

andyleejordan (approved):
I think I'd want to double check using `canonicalize` over `which_in` given the potential UNC issues, but up to your discretion. This otherwise looks good to me, though please file an issue for the Unicode path bug we probably have.

SteveL-MSFT:
@andyleejordan created https://github.com/PowerShell/DSC/issues/1233 to track using `PathBuf` throughout the code in the future

I appreciate you digging into the details here.  In this case, the result is only ever used by `CreateProcess()` so not worried about the UNC long file path.  Will take this as-is for now.  Thanks!

PR #1226, opened 2025-10-31
(GH-538) Enable retrieving keywords as `Schema` objects

michaeltlombardi:
# PR Summary

This change:

- Adds two helper methods to retrieve a keyword as a `schemars::Schema` instance (borrowed and mutably borrowed) to make working with subschemas more ergonomic.
- Updates the following functions to return instances of `schemars::Schema` instead of `Map<String, Value>`, since the returned data is _always_ a subschema, if it exists:

  - `get_defs_subschema_from_id()`
  - `get_defs_subschema_from_id_mut()`
  - `get_defs_subschema_from_reference()`
  - `get_defs_subschema_from_reference_mut()`
  - `get_property_subschema()`
  - `get_property_subschema_mut()`
- Removes the type aliases `Object` (for `Map<String, Value>`) and `Array` (for `Vec<Value>`), as these conveniences weren't saving much typing and Rust Analyzer wasn't always plumbing them through for IntelliSense. The uses of these aliases now revert to calling the underlying types.
- Updates documentation and tests for the modified functions.

This change also refactors the `dsc-lib-jsonschema` library for consistency and ease of navigation:

- Splits the functions in the `transforms` module out into submodules and re-exports them from `transforms` - this keeps referencing the functions the way it was before but makes it easier to navigate the files, given their length.
- Makes the unit tests for `schema_utility_extensions` mirror the structure from the source code.
- Makes the integration tests for `transform` mirror the structure from the source code.

## PR Context

Prior to this change, the `dsc-lib-jsonschema` crate defined extension methods for schemas to retrieve keywords as various underlying data types when the caller knows the data type they need. However, the extension methods didn't include a way to retrieve values as _schemas_, so any use of those values requires the caller to reimplement the extension methods logic or to manually convert the value to a schema.


SteveL-MSFT:
@michaeltlombardi can you fix the merge conflict, I think it's due to a change I made to revert a workaround you put in since schemars was updated and fixed the issue

michaeltlombardi:
@SteveL-MSFT - rebased and resolved, should be ready for review.

SteveL-MSFT (approved):

PR #1229, opened 2025-11-02
Add parameter validation checks for length, allowed values, and number limits on defaultValue

### LINKED ISSUE #1228: defaultValue on constraints are not being evaluated
### Prerequisites

### Summary

Parameter constraints (`minValue`, `maxValue`, `minLength`, `maxLength`, `allowedValues`) are not being validated when parameters use their default values. Constraints are correctly validated when parameters are provided at runtime via input, but the validation is skipped entirely when a parameter falls back to its `defaultValue`.

### Steps to reproduce

1. Create a configuration with a parameter that has an invalid default value:

```yaml
$schema: https://raw.githubusercontent.com/PowerShell/DSC/main/schemas/2024/04/config/document.json
parameters:
  port:
    type: int
    minValue: 1
    maxValue: 65535
    defaultValue: 99999  # Violates maxValue constraint
resources:
  - name: Echo Test
    properties:
      output: "[parameters('port')]"
```

2. Run the configuration WITHOUT providing parameter input:

```powershell
dsc config get --file config.yaml
```

### Expected behavior

```console
DSC should validate the `defaultValue` (99999) against the `maxValue` constraint (65535) and raise an error:


ERROR Parameter input failure: Validation: Parameter 'port' has maximum value constraint of 65535 but is 99999
```

### Actual behavior

```console
The configuration runs successfully without any validation error, using the invalid default value of 99999.
```

### Error details

```console

```

### Environment data

```PowerShell
PSVersion                      7.5.4
GitCommitId                    7.5.4
```

### Version

dsc 3.2.0-preview.7

### Visuals

_No response_

Gijsreyn:
# PR Summary

This PR fixes a bug where parameter constraints (`minValue`, `maxValue`, `minLength`, `maxLength`, `allowedValues`) were not being validated when parameters used their `defaultValue`. Constraints were only validated for runtime input parameters, allowing invalid default values to pass through unchecked.

## PR Context

Fix #1228.

### DIFF
--- a/dsc/tests/dsc_parameters.tests.ps1
+++ b/dsc/tests/dsc_parameters.tests.ps1
@@ -590,15 +590,16 @@ Describe 'Parameters tests' {
             $value = "'$value'"
         }
 
+        $minConstraint = if ($type -eq 'string') { "minLength: $min" } else { "minValue: $min" }
+        $maxConstraint = if ($type -eq 'string') { "maxLength: $max" } else { "maxValue: $max" }
+
         $config_yaml = @"
             `$schema: https://aka.ms/dsc/schemas/v3/bundled/config/document.json
             parameters:
               param1:
                 type: $type
-                minLength: $min
-                maxLength: $max
-                minValue: $min
-                maxValue: $max
+                $minConstraint
+                $maxConstraint
                 defaultValue: $value
             resources:
             - name: Echo
@@ -651,19 +652,16 @@ Describe 'Parameters tests' {
     ) {
         param($type, $value, $min, $max)
 
-        if ($type -eq 'string') {
-            $value = "'$value'"
-        }
+        $minConstraint = if ($type -eq 'string') { "minLength: $min" } else { "minValue: $min" }
+        $maxConstraint = if ($type -eq 'string') { "maxLength: $max" } else { "maxValue: $max" }
 
         $config_yaml = @"
             `$schema: https://aka.ms/dsc/schemas/v3/bundled/config/document.json
             parameters:
               param1:
                 type: $type
-                minLength: $min
-                maxLength: $max
-                minValue: $min
-                maxValue: $max
+                $minConstraint
+                $maxConstraint
                 defaultValue: $value
             resources:
             - name: Echo
@@ -684,10 +682,6 @@ Describe 'Parameters tests' {
         @{ type = 'int'; value = 10; allowed = @(1, 5, 10) }
     ) {
         param($type, $value, $allowed)
-
-        if ($type -eq 'string') {
-            $value = "'$value'"
-        }
 
         $config_yaml = @"
             `$schema: https://aka.ms/dsc/schemas/v3/bundled/config/document.json


dsc/tests/dsc_parameters.tests.ps1
```diff
@@ -577,4 +577,164 @@ Describe 'Parameters tests' {
...
+"@
+
+        $testError = & {$config_yaml | dsc config get -f - 2>&1}
+        $LASTEXITCODE | Should -Be 4
+        $testError | Should -Match 'allowed values'
+    }
+
+    It 'Default values pass constraint validation for <type>' -TestCases @(
+        @{ type = 'string'; value = 'admin'; min = 3; max = 20 }
+        @{ type = 'int'; value = 8080; min = 1; max = 65535 }
```
SteveL-MSFT, line 649: Can you add tests for boundary conditions, exactly min and exactly max?

dsc/tests/dsc_parameters.tests.ps1
```diff
@@ -577,4 +577,164 @@ Describe 'Parameters tests' {
...
+                type: $type
+                minLength: $min
+                maxLength: $max
+                defaultValue: $value
+            resources:
+            - name: Echo
+              type: Microsoft.DSC.Debug/Echo
+              properties:
+                output: '[parameters(''param1'')]'
+"@
```
SteveL-MSFT: Since the two parts if the `if` statement are essentially the same, I think you should just simplify this to:

```
if ($type -eq 'string') {
  $value = "'$value'"
}
$config_yaml = ...
```

dsc/tests/dsc_parameters.tests.ps1
```diff
@@ -577,4 +577,164 @@ Describe 'Parameters tests' {
...
+        }
+
+        $out = $config_yaml | dsc config get -f - | ConvertFrom-Json
+        $LASTEXITCODE | Should -Be 0
+        if ($type -eq 'string') {
+            $out.results[0].result.actualState.output | Should -BeExactly 'admin'
```
SteveL-MSFT: Also here. I don't think you need to differentiate `string` here.  I believe `Should -BeEaxctly $value` would work.

dsc/tests/dsc_parameters.tests.ps1
```diff
@@ -577,4 +577,164 @@ Describe 'Parameters tests' {
...
+        } else {
+            $out.results[0].result.actualState.output | Should -BeExactly 8080
+        }
+    }
+
+    It 'Default values with allowedValues pass validation for <type>' -TestCases @(
+        @{ type = 'string'; value = 'dev'; allowed = @('dev', 'test', 'prod') }
+        @{ type = 'int'; value = 5; allowed = @(1, 5, 10) }
```
SteveL-MSFT, line 681: Maybe add one more test variation covering the last allowed value, doesn't matter if string or int.

SteveL-MSFT (changes requested):

Gijsreyn:
Thanks for the quick check @SteveL-MSFT. Resolved the remarks. Slightly rewritten the constraints tests.

PR #1230, opened 2025-11-03
Add `parseCidr()`, `cidrHost()`, and `cidrSubnet()` functions

Gijsreyn:
# PR Summary

The following pull request adds the cidr functions (`parseCidr()`, `cidrHost()`, `cidrSubnet()`) functions including their reference documentation.

I've tried to re-use logic in the existing crates, but they either didn't provide the functionality I was looking for, or they were quite old.

## PR Context

Partially addresses #57.


docs/reference/schemas/config/functions/cidrHost.md
```diff
@@ -0,0 +1,226 @@
...
+- name: Gateway address
+  type: Microsoft.DSC.Debug/Echo
+  result:
+    actualState:
+      output: 10.0.1.1
+messages: []
+hadErrors: false
+```
+
+### Example 2 - Assign multiple host addresses
```
SteveL-MSFT, line 64: All the doc examples use ipv4 which implies ipv6 isn't supported.  Can you make sure each doc has one ipv6 example?  You can either convert an existing ipv4 one to ipv6 or add a new one.

dsc/tests/dsc_functions.tests.ps1
```diff
@@ -1293,4 +1293,273 @@ Describe 'tests for function expressions' {
...
+"@
+    $out = $config_yaml | dsc config get -f - | ConvertFrom-Json
+    $LASTEXITCODE | Should -Be 0
+    $result = $out.results[0].result.actualState.output
+    $result.network | Should -BeExactly '192.168.1.0'
+    $result.broadcast | Should -BeExactly '192.168.1.255'
+  }
+
+  It 'parseCidr fails with invalid CIDR: <cidr>' -TestCases @(
+    @{ cidr = 'invalid'; errorMatch = 'Invalid CIDR notation' }
```
SteveL-MSFT: I think for invalid, it may be better to have something that looks like cidr

dsc/tests/dsc_functions.tests.ps1
```diff
@@ -1293,4 +1293,273 @@ Describe 'tests for function expressions' {
...
+            resources:
+            - name: Echo
+              type: Microsoft.DSC.Debug/Echo
+              properties:
+                output: "[cidrHost('192.168.1.0/31', 1)]"
+"@
+    $out = $config_yaml | dsc config get -f - | ConvertFrom-Json
+    $LASTEXITCODE | Should -Be 0
+    $out.results[0].result.actualState.output | Should -BeExactly '192.168.1.1'
+  }
```
SteveL-MSFT: Is there a reason these two variants are just in the TestCases above?

lib/dsc-lib/src/functions/cidr_host.rs
```diff
@@ -0,0 +1,329 @@
...
+    }
+
+    fn invoke(&self, args: &[Value], _context: &Context) -> Result<Value, DscError> {
+        debug!("{}", t!("functions.cidrHost.invoked"));
+
+        let cidr_string = args[0].as_str().ok_or_else(|| {
+            DscError::FunctionArg(
+                "cidrHost".to_string(),
+                t!("functions.cidrHost.invalidNetwork").to_string(),
+            )
```
SteveL-MSFT: Since parameter types are pre-validated, this error will never get returned so no need to have this string stored in the locales.  You should just `as_str().unwrap();` as it should never panic.

lib/dsc-lib/src/functions/cidr_host.rs
```diff
@@ -0,0 +1,329 @@
...
+        })?;
+
+        let host_index = args[1].as_i64().ok_or_else(|| {
+            DscError::FunctionArg(
+                "cidrHost".to_string(),
+                t!("functions.cidrHost.invalidHostIndex").to_string(),
+            )
```
SteveL-MSFT: Same here

lib/dsc-lib/src/functions/cidr_host.rs
```diff
@@ -0,0 +1,329 @@
...
+        })?;
+
+        if host_index < 0 {
+            return Err(DscError::FunctionArg(
+                "cidrHost".to_string(),
+                t!("functions.cidrHost.negativeHostIndex").to_string(),
+            ));
```
SteveL-MSFT, line 43: Wouldn't `parse()` below fail already if negative?

lib/dsc-lib/src/functions/cidr_host.rs
```diff
@@ -0,0 +1,329 @@
...
+        let result = match network {
+            IpNetwork::V4(net) => calculate_ipv4_host(net, host_index as u32)?,
+            IpNetwork::V6(net) => calculate_ipv6_host(net, host_index as u128)?,
+        };
+
+        Ok(Value::String(result))
+    }
+}
+
+fn calculate_ipv4_host(net: Ipv4Network, host_index: u32) -> Result<String, DscError> {
```
SteveL-MSFT:
```suggestion
fn calculate_ipv4_host(net: &Ipv4Network, host_index: u32) -> Result<String, DscError> {
```

Should use a reference for structs so a copy isn't implicitly made

lib/dsc-lib/src/functions/cidr_subnet.rs
```diff
@@ -0,0 +1,342 @@
...
+
+#[derive(Debug, Default)]
+pub struct CidrSubnet {}
+
+impl Function for CidrSubnet {
+    fn get_metadata(&self) -> FunctionMetadata {
+        FunctionMetadata {
+            name: "cidrSubnet".to_string(),
+            description: t!("functions.cidrSubnet.description").to_string(),
+            category: vec![FunctionCategory::String],
```
SteveL-MSFT: Looks like ARM put these under their own category `CIDR` so you should add a new category for this instead of under String

lib/dsc-lib/src/functions/cidr_subnet.rs
```diff
@@ -0,0 +1,342 @@
...
+    }
+
+    fn invoke(&self, args: &[Value], _context: &Context) -> Result<Value, DscError> {
+        debug!("{}", t!("functions.cidrSubnet.invoked"));
+
+        let cidr_string = args[0].as_str().ok_or_else(|| {
+            DscError::FunctionArg(
+                "cidrSubnet".to_string(),
+                t!("functions.cidrSubnet.invalidNetwork").to_string(),
+            )
```
SteveL-MSFT: same as before

lib/dsc-lib/src/functions/parse_cidr.rs
```diff
@@ -0,0 +1,274 @@
...
+                })
+            }
+            IpNetwork::V6(net) => {
+                let network_addr = net.network();
+                let broadcast_addr = net.broadcast();
+
+                json!({
+                    "network": network_addr.to_string(),
+                    "netmask": net.mask().to_string(),
+                    "broadcast": broadcast_addr.to_string(),
```
SteveL-MSFT: Based on this example: https://learn.microsoft.com/en-us/azure/azure-resource-manager/templates/template-functions-cidr#examples, ipv6 doesn't return `broadcast` property

dsc/tests/dsc_functions.tests.ps1
Gijsreyn: It showcases different behaviour, so I thought it deserved its own test to highlight this edge case.

lib/dsc-lib/src/functions/cidr_host.rs
Gijsreyn: I think negative i64 wraps to huge u32 when cast. The check gives users a clearer error.

lib/dsc-lib/src/functions/mod.rs
```diff
@@ -362,6 +363,7 @@ impl Display for FunctionCategory {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         match self {
             FunctionCategory::Array => write!(f, "Array"),
+            FunctionCategory::Cidr => write!(f, "Cidr"),
```
SteveL-MSFT: Since `CIDR` is an acronym, should be all in uppercase.  The enum value being Pascal case is fine, just the Display implementation.

lib/dsc-lib/src/functions/mod.rs
```diff
@@ -363,7 +363,7 @@ impl Display for FunctionCategory {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         match self {
             FunctionCategory::Array => write!(f, "Array"),
-            FunctionCategory::Cidr => write!(f, "Cidr"),
+            FunctionCategory::CIDR => write!(f, "Cidr"),
```
SteveL-MSFT: Sorry, you reversed it.  Ok to keep the enum value Pascal casing, but the `write!` macro should use CIDR.

Gijsreyn: Oops, got you wrong. Done!

SteveL-MSFT (approved):

PR #1238, opened 2025-11-05
Add lambda expression and `map()` + `filter()` functions with ARM syntax

Gijsreyn:
# PR Summary

This PR implements lambda expressions and the `map()` function for DSC configuration documents, enabling array transformations. The implementation uses ARM compatible syntax e.g., (`lambda('param', body)` and `lambdaVariables('param')`)

**Example usage:**

```yaml
# Transform array elements
map(createArray(1, 2, 3), lambda('x', mul(lambdaVariables('x'), 2)))
# Returns: [2, 4, 6]
```

Seperated tests in a different file.

## PR Context

Partially addresses #57.

### DIFF
--- a/lib/dsc-lib/src/functions/lambda_variables.rs
+++ b/lib/dsc-lib/src/functions/lambda_variables.rs
@@ -22,12 +22,12 @@ impl Function for LambdaVariables {
             accepted_arg_ordered_types: vec![vec![FunctionArgKind::String]],
             remaining_arg_accepted_types: None,
             return_types: vec![
-                FunctionArgKind::String,
-                FunctionArgKind::Number,
-                FunctionArgKind::Boolean,
                 FunctionArgKind::Array,
-                FunctionArgKind::Object,
+                FunctionArgKind::Boolean,
                 FunctionArgKind::Null,
+                FunctionArgKind::Number,
+                FunctionArgKind::Object,
+                FunctionArgKind::String,
             ],
         }
     }


lib/dsc-lib/src/configure/context.rs
```diff
@@ -25,6 +25,8 @@ pub struct Context {
     pub dsc_version: Option<String>,
     pub execution_type: ExecutionKind,
     pub extensions: Vec<DscExtension>,
+    pub lambda_variables: HashMap<String, Value>,
```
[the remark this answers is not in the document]
Gijsreyn: N/A as other don't have it.

lib/dsc-lib/src/configure/context.rs
```diff
@@ -25,6 +25,8 @@ pub struct Context {
...
+    pub lambdas: std::cell::RefCell<HashMap<String, crate::parser::functions::Lambda>>,
```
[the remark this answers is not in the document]
Gijsreyn: Isolation is correct, as each iteration of `map()` should have its own isolated scope.

lib/dsc-lib/src/parser/functions.rs
```diff
@@ -23,6 +23,13 @@ pub struct Function {
 pub enum FunctionArg {
     Value(Value),
     Expression(Expression),
+    Lambda(Lambda),
+}
+
+#[derive(Clone)]
```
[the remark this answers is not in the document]
Gijsreyn: N/A

lib/dsc-lib/src/functions/mod.rs
```diff
@@ -277,6 +283,63 @@ impl FunctionDispatcher {
...
+            body: body_expr,
+        };
+        
+        let lambda_id = format!("__lambda_{}", Uuid::new_v4());
+        context.lambdas.borrow_mut().insert(lambda_id.clone(), lambda);
+
+        // Return the ID as a string value
+        Ok(Value::String(lambda_id))
+    }
+
```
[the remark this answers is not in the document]
Gijsreyn: False alarm

lib/dsc-lib/src/functions/map.rs
```diff
@@ -0,0 +1,99 @@
...
+    fn get_metadata(&self) -> FunctionMetadata {
+        FunctionMetadata {
+            name: "map".to_string(),
+            description: t!("functions.map.description").to_string(),
+            category: vec![FunctionCategory::Array, FunctionCategory::Lambda],
+            min_args: 2,
+            max_args: 2,
+            accepted_arg_ordered_types: vec![
+                vec![FunctionArgKind::Array],
+                vec![FunctionArgKind::String], // Lambda ID as string
```
SteveL-MSFT: Since `lambda` is special, did you consider defining it as a new `FunctionArgKind` since only some functions accept it?  This way validation of what accepts it will happen with other parameter validation.

Gijsreyn: I hadn't considered that, but I've added it now. Thanks Steve!

lib/dsc-lib/src/functions/lambda_variables.rs
```diff
@@ -0,0 +1,77 @@
...
+            ],
+        }
+    }
+
+    fn invoke(&self, args: &[Value], context: &Context) -> Result<Value, DscError> {
+        debug!("{}", t!("functions.lambdaVariables.invoked"));
+        
+        if args.len() != 1 {
+            return Err(DscError::Parser(t!("functions.invalidArgCount", name = "lambdaVariables", count = 1).to_string()));
+        }
```
SteveL-MSFT: This isn't needed as function validation ensures min_args and max_args is correct

lib/dsc-lib/src/functions/lambda_variables.rs
```diff
@@ -0,0 +1,77 @@
...
+
+        let Some(var_name) = args[0].as_str() else {
+            return Err(DscError::Parser(t!("functions.lambdaVariables.paramNameMustBeString").to_string()));
+        };
```
SteveL-MSFT: Same here, just unwrap.

lib/dsc-lib/src/functions/map.rs
```diff
@@ -0,0 +1,99 @@
...
+            return Err(DscError::Parser(t!("functions.invalidArgCount", name = "map", count = 2).to_string()));
+        }
+
+        let Some(array) = args[0].as_array() else {
+            return Err(DscError::Parser(t!("functions.map.firstArgMustBeArray").to_string()));
+        };
+
+        let Some(lambda_id) = args[1].as_str() else {
+            return Err(DscError::Parser(t!("functions.map.secondArgMustBeLambda").to_string()));
+        };
```
SteveL-MSFT: Same as above, this is all validated already so these strings won't be used

lib/dsc-lib/src/functions/mod.rs
```diff
@@ -289,18 +297,79 @@ impl FunctionDispatcher {
...
+            parameters,
+            body: body_expr,
+        };
+        
+        let lambda_id = format!("__lambda_{}", Uuid::new_v4());
+        context.lambdas.borrow_mut().insert(lambda_id.clone(), lambda);
+
+        // Return the ID as a string value
+        Ok(Value::String(lambda_id))
+    }
```
SteveL-MSFT: Instead of special handling it this way, you should have a new ProcessMode::Lambda and just have the `lambda()` function do this only when process mode is correct other return an error.  This keeps it consistent with modes.

Gijsreyn: I need help with this, as I see two options here: either use `RefCell<ProcessMode>`, which I guess is less invasive and doesn't require many file changes, or change the `invoke` signature to `&mut Context`, which requires a ton of file changes, but follows the Rust design. Do you see another option here?

SteveL-MSFT: I don't think making it `mut` everywhere is a good solution.  Instead, perhaps modify the `invoke()` on line 256 above and if the function takes any `Lambda` args, then clone the context and modify it setting ProcessMode::Lambda.  The only other consideration I was thinking about is whether there's now going to be overlap of ProcessModes such that it needs to be bit flags, but I haven't completely thought it through whether that's the case here.

Gijsreyn: I've tried implementing your suggestion. Can you please check again?

lib/dsc-lib/src/functions/mod.rs
```diff
@@ -284,27 +292,97 @@ impl FunctionDispatcher {
...
+    ///
+    /// # Arguments
+    ///
+    /// * `args` - Raw FunctionArg list (unevaluated)
+    /// * `context` - Context to store the lambda in
+    ///
+    /// # Errors
+    ///
+    /// This function will return an error if the lambda syntax is invalid.
+    pub fn invoke_lambda(&self, args: &Option<Vec<crate::parser::functions::FunctionArg>>, context: &Context) -> Result<Value, DscError> {
```
SteveL-MSFT: With the other change, I think you should move this code into the body of the `lambda()` function itself and just have it check that `ProcessMode == Lambda`.  That way it should be simpler for future lambda functions to be implemented I think.

lib/dsc-lib/src/functions/map.rs
```diff
@@ -0,0 +1,78 @@
...
+                lambda_context.lambda_variables.insert(
+                    lambda.parameters[1].clone(),
+                    Value::Number(serde_json::Number::from(index))
+                );
+            }
+
+            // Evaluate lambda body with bound variables
+            let result = lambda.body.invoke(&dispatcher, &lambda_context)?;
+            result_array.push(result);
+        }
```
SteveL-MSFT: I haven't thought this through, but instead of having each lambda function implement processing lambdas, it would seem like this could be in `lambda()` itself so the other lambda functions in https://learn.microsoft.com/en-us/azure/azure-resource-manager/templates/template-functions-lambda would just handle how to apply the lambda result.

Perhaps it would help to also implement `filter()` in this PR to see how it would work?

SteveL-MSFT (changes requested):

lib/dsc-lib/src/functions/mod.rs
Gijsreyn: Done!

lib/dsc-lib/src/functions/map.rs
Gijsreyn: I added the `filter()` function to compare, where I had introduced a new file (`lamba_helpers.rs`). Mind checking again?

lib/dsc-lib/src/functions/lambda_variables.rs
```diff
@@ -0,0 +1,71 @@
...
+            max_args: 1,
+            accepted_arg_ordered_types: vec![vec![FunctionArgKind::String]],
+            remaining_arg_accepted_types: None,
+            return_types: vec![
+                FunctionArgKind::String,
+                FunctionArgKind::Number,
+                FunctionArgKind::Boolean,
+                FunctionArgKind::Array,
+                FunctionArgKind::Object,
+                FunctionArgKind::Null,
```
SteveL-MSFT:
```suggestion
                FunctionArgKind::Array,
                FunctionArgKind::Boolean,
                FunctionArgKind::Null,
                FunctionArgKind::Number,
                FunctionArgKind::Object,
                FunctionArgKind::String,
```

SteveL-MSFT:
@Gijsreyn can you resolve the merge conflict and accept my suggestion to order then alpha?

SteveL-MSFT:
@Gijsreyn can you accept my suggestion to alpha order the ArgKind and I can merge?

Gijsreyn:
@SteveL-MSFT - my bad. I thought I had already pushed it. It's good to go!

PR #1258, opened 2025-11-14
Add `objectKeys()` function

Gijsreyn:
# PR Summary

This pull request adds the `objectKeys()` function including its reference documentation.

## PR Context

Partially addresses #57.

### DIFF
--- a/lib/dsc-lib/src/functions/object_keys.rs
+++ b/lib/dsc-lib/src/functions/object_keys.rs
@@ -27,9 +27,7 @@ impl Function for ObjectKeys {
     }
 
     fn invoke(&self, args: &[Value], _context: &Context) -> Result<Value, DscError> {
-        let Some(obj) = args[0].as_object() else {
-            return Err(DscError::Parser(t!("functions.objectKeys.notObject").to_string()));
-        };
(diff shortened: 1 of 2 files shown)


lib/dsc-lib/src/functions/object_keys.rs
```diff
@@ -0,0 +1,136 @@
...
+                vec![FunctionArgKind::Object],
+            ],
+            remaining_arg_accepted_types: None,
+            return_types: vec![FunctionArgKind::Array],
+        }
+    }
+
+    fn invoke(&self, args: &[Value], _context: &Context) -> Result<Value, DscError> {
+        let Some(obj) = args[0].as_object() else {
+            return Err(DscError::Parser(t!("functions.objectKeys.notObject").to_string()));
```
SteveL-MSFT: This error will never happen since arg types are prevalidated.  Just use:

```rust
let obj = args[0].as_object().unwrap();
```

unless the compiler or clippy is complaining, otherwise this is a string that never gets returned and takes up space

SteveL-MSFT (changes requested):

Issue #586, opened 2024-10-25
Python adapter


SteveL-MSFT:
### Summary of the new feature / enhancement

To make DSC successful on Linux, we probably need a Python adapter.  The key part is to keep it idiomatic Python.

### Proposed technical implementation details (optional)

We could use Python classes with the same type of model as PowerShell classes with individual `get()`, `set()`, `test()`, `export()` methods and the class defines the schema that the adapter would validate against.

michaeltlombardi:
It's probably a good idea to implement this as a python script that uses a module for defining and interacting with resources, so resource authors can leverage some shared code to define resources.

For example, we could define `DscResource` as an [abstract class](https://docs.python.org/3/library/abc.html) that authors could use - ensuring a common implementation surface that is documented and reusable. We could also provide the helpers for converting to/from JSON, emitting messages for DSC, retrieving manifest information, etc.

EmperorArthur:
Before finding this issue, I created the quick and dirty resource found in PR #1262.  It's nowhere near production ready, but may be of some use.

PR #1262, opened 2025-11-14
Quick and dirty Python Pip Resource

EmperorArthur:
# PR Summary

Quick and dirty implementation of #586

## PR Context

Wanting to use DSCV3 with Pip I came up with this simple script as my first resource. It's nowhere near perfect, but does work and should be helpful for the final implementation.

# ONLY EXAMPLE CODE (do not merge as is)
This should be a complete python module with a `pyproject.toml` file.  Since it is not, I have left the code in the "resources" folder as a clear indication this is not ready to be merged in its current form.

### DIFF
--- a/resources/pip.dsc.py
+++ b/resources/pip.dsc.py
@@ -1,0 +1,197 @@
+#!/bin/env python
+# Quick and Dirty Python Pip DSC V3 Resource
+# MIT LIcense
+# Copyright 2025 Arthur Moore & Higher Education Loan Authority of the State of Missouri (MOHELA)
+
+import subprocess
+import sys
+import json
+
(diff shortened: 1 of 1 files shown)


EmperorArthur:
@microsoft-github-policy-service agree company="MOHELA"

SteveL-MSFT:
@EmperorArthur thanks for the contribution!  However, we have in progress work already for a formal Python adapter.  I think we'll have something public to show end of this year.

EmperorArthur:
@SteveL-MSFT That's great to hear!  I was dreading having to figure out how to package and distribute this thing.

I hope the related feature requests show what issues a novice to the project experiences.

PR #1275, opened 2025-11-19
Add sshdconfig set for "regular" keywords & clobber = true

tgauth:
# PR Summary
- note: to minimize PR size this is only intended to support: `clobber: true` and regular keyword (no objects/arrays yet)
- general flow of `set` command: create temp file based on input, validate temp file using `sshd -T`, back-up any existing sshd_config file (only if it is not already managed by this resource), then update/create sshd_config file
- add Pester tests

### DIFF
--- a/resources/sshdconfig/locales/en-us.toml
+++ b/resources/sshdconfig/locales/en-us.toml
@@ -7,6 +7,7 @@ setInput = "input to set in sshd_config"
 
 [error]
 command = "Command"
+envVar = "Environment Variable"
 invalidInput = "Invalid Input"
 fmt = "Format"
 io = "IO"


resources/sshdconfig/src/metadata.rs
```diff
@@ -39,6 +39,12 @@ pub const REPEATABLE_KEYWORDS: [&str; 12] = [
     "subsystem"
 ];
 
+
+pub const SSHD_CONFIG_HEADER: &str = "# This file is managed by the Microsoft DSC sshdconfig resource.";
```
SteveL-MSFT: Should probably put the full resource type name here.  It may also make sense to include the version of the resource (version can be on a separate line so you don't need to handle comparing it).  Also suggest an additional line telling the user not to make modifications directly lin the file as it will be overwritten.

resources/sshdconfig/src/set.rs
```diff
@@ -86,3 +105,79 @@ fn remove_registry(name: &str) -> Result<(), SshdConfigError> {
...
+            filepath: Some(temp_path),
+            additional_args: None,
+        }
+    );
+
+    debug!("{}", t!("set.validatingTempConfig"));
+    let result = invoke_sshd_config_validation(args);
+    // Always cleanup temp file, regardless of result success or failure
+    if let Err(e) = std::fs::remove_file(&path) {
+        debug!("{}", t!("set.cleanupFailed", path = path.display(), error = e));
```
SteveL-MSFT: Since the user might want to know, this should probably be a warning

resources/sshdconfig/src/set.rs
```diff
@@ -86,3 +105,79 @@ fn remove_registry(name: &str) -> Result<(), SshdConfigError> {
...
+        } else {
+            return Err(SshdConfigError::CommandError(t!("set.sshdConfigReadFailed", path = sshd_config_path.display()).to_string()));
+        }
+        if !sshd_config_content.starts_with(SSHD_CONFIG_HEADER) {
+            // If config file is not already managed by this resource, create a backup of the existing file
+            debug!("{}", t!("set.backingUpConfig"));
+            let backup_path = format!("{}.bak", sshd_config_path.display());
+            std::fs::write(&backup_path, &sshd_config_content)
+                .map_err(|e| SshdConfigError::CommandError(e.to_string()))?;
+            debug!("{}", t!("set.backupCreated", path = backup_path));
```
SteveL-MSFT: This should be an info

resources/sshdconfig/src/util.rs
```diff
@@ -34,6 +35,22 @@ pub fn enable_tracing() {
...
+/// Get the `sshd_config` path
+/// Uses the input value, if provided.
+/// If input value not provided, get default path for the OS.
+/// On Windows, uses the `ProgramData` environment variable and standard path.
+/// On Unix-like systems, uses the standard path.
+pub fn get_default_sshd_config_path(input: Option<PathBuf>) -> PathBuf {
+    if let Some(path) = input {
+        path
+    } else if cfg!(windows) {
+        let program_data = std::env::var("ProgramData").unwrap_or_else(|_| "C:\\ProgramData".into());
```
SteveL-MSFT: You can't rely on system drive always being `C:`.  I think you can just fail if the env var isn't set.

PR #1276, opened 2025-11-19
(GH-538) Add VS Code keywords, vocabulary, and dialect to `dsc-lib-jsonschema`

michaeltlombardi:
# PR Summary

This change:

- Defines every custom keyword that the VS Code language server for JSON recognizes as custom keywords in the `jsonschema` crate so that we can correctly validate schemas that use those keywords programmatically.

  The implementation in `jsonschema` is primarily intended for custom validation keywords. Every custom keyword that VS Code recognizes is an annotation keyword. They don't affect validation for instances of data. This implementation ensures that we can parse those keywords for our own use and validate that our schema definitions correctly define values for those keywords.
- Defines the VS Code vocabulary, which includes the keywords. This enables us to define our own dialects which recognize the VS Code keywords.
- Defines a VS Code dialect, which includes the Draft 2020-12 vocabularies and the VS Code vocabulary.
- Defines the `VSCodeKeyword` enum for easier selection of keywords, given that you can't ergonomically or safely pass a _type_ in Rust functions.

  This replaces the previous `VSCODE_KEYWORDS` struct.
- Updates the `idiomaticize_externally_tagged_enum` transform function to use the `VSCodeKeyword` enum now that `VSCODE_KEYWORDS` is removed.
- Defines extension methods for the `ValidationOptions` struct from the `jsonschema` crate to make adding VS Code keywords, vocabulary, and dialect simpler and more ergonomic, participating in the builder pattern that `jsonschema` uses when creating a validator.
- Defines extension methods for the `Schema` struct from the `schemars` crate to simplify working with VS COde keywords in schemas, such as retrieving the value for a keyword or checking whether the schema uses the VS Code dialect.
- Adds three new extension methods to the `SchemaUtilityExtensions` trait:

  - `get_bundled_schema_resource_ids`, which retrieves the value of the `$id` keyword from entries in the `$defs` keyword, each of which represents a bundled schema resource.
  - `get_references`, which recursively retrieves the value for every `$ref` keyword in the schema.
  - `replace_references`, which recursively replaces the value of the `$ref` keyword from a given value to a new one specified by the caller.
  - `reference_is_for_bundled_resource`, which indicates whether a given value for a `$ref` keyword points to a bundled schema resource in the `$defs` keyword for the schema.

## PR Context

This change addresses the need for parsing, validating, and processing the custom JSON Schema keywords that VS Code recognizes for use in the DSC schemas and resource schemas. This change lays the groundwork for incorporating the VS Code keywords into the schemas generated from Rust code and for defining our own vocabulary for DSC as needed.

### DIFF
--- a/build.helpers.psm1
+++ b/build.helpers.psm1
@@ -644,6 +644,16 @@ function Install-PowerShellTestPrerequisite {
             Write-Verbose "Installing module 'Pester'"
             Install-PSResource Pester -WarningAction Ignore -Repository $repository -TrustRepository
         }
+
+        if (-not (Get-Module -ListAvailable -Name YaYaml)) {
+            Write-Verbose "Installing module 'YaYaml'"
+            Install-PSResource YaYaml -WarningAction Ignore -Repository $repository -TrustRepository
+        }
+
+        if (-not (Get-Module -ListAvailable -Name PSToml)) {
+            Write-Verbose "Installing module 'PSToml'"
+            Install-PSResource PSToml -WarningAction Ignore -Repository $repository -TrustRepository
+        }
     }
 }
 
--- a/dsc/tests/dsc_i18n.tests.ps1
+++ b/dsc/tests/dsc_i18n.tests.ps1
@@ -2,57 +2,252 @@
 # Licensed under the MIT License.
 
 BeforeDiscovery {
-    $tomls = Get-ChildItem $PSScriptRoot/../../en-us.toml -Recurse -File
-    $projects = @()
-    $tomls | ForEach-Object {
-        $projectName = (Split-Path $_ -Parent | Split-Path -Parent)
-        $projects += @{ project = $projectName; toml = $_ }
-    }
-}
+    #region    Rust i18n type definitions and functions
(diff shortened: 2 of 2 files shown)


build.helpers.psm1
```diff
@@ -1539,6 +1549,186 @@ function Export-RustDocs {
 }
 #endregion Documenting project functions
 
+#region    Rust i18n type definitions and functions
```
SteveL-MSFT: Is this used anywhere outside of the i18n tests?  If not, would prefer to keep that code together in the test ps1 rather than creating a dependency on the helper module for that test

michaeltlombardi: I put it here for easier access to checking translations during normal development workflow, since the i18n tests are in the `dsc` project - when working locally I was testing against the `dsc-lib-jsonschema` and `dsc-lib` crates independently.

Happy to move it though if that's preferable.

SteveL-MSFT: My preference would be to move it since someone reading this in the future wouldn't be aware of the separation of the code.  If it was used across multiple scripts, then it would make more sense to me to have it centralized here.

PR #1281, opened 2025-11-20
update sshd config grammar

tgauth:
# PR Summary
- update sshd config grammar to use aliasing and handle parsing the following strings: 
  - strings without spaces that are potentially comma-separated
  - strings potentially with spaces that are either single-quoted or double-quoted
- update sshd config grammar logic to support multiple `Match` criteria

### DIFF
--- a/grammars/tree-sitter-ssh-server-config/grammar.js
+++ b/grammars/tree-sitter-ssh-server-config/grammar.js
@@ -48,7 +48,7 @@ module.exports = grammar({
     boolean: $ => choice('yes', 'no'),
     number: $ => /\d+/,
     operator: $ => token(prec(PREC.OPERATOR, /[-+\^]/)),
-    string: $ => /[^\r\n,"'\s]+/, /* cannot contains spaces */
+    string: $ => /[^\r\n,"'\s]+/, /* cannot contain spaces */
 
     _quotedString: $ => /[^\r\n,"']+/, /* can contain spaces */
     _doublequotedString: $ => seq('"', alias($._quotedString, $.string), repeat(seq(',', alias($._quotedString, $.string))), '"'),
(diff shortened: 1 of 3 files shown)


grammars/tree-sitter-ssh-server-config/grammar.js
```diff
@@ -43,7 +43,7 @@ module.exports = grammar({
     string: $ => /[^\r\n,"]+/,
 
     _commaSeparatedString: $ => seq($.string, repeat(seq(',', $.string))),
-    _quotedString: $ => seq('\"', $.string, '\"'),
+    quotedString: $ => seq('\"', $.string, '\"'),
```
SteveL-MSFT: For the parser, you don't care if it's quoted or not, but simply the result is the contents between the quotes (if provided), right?  Can't you define a string as a choice between quoted and not quoted where the quoted is just the contents between the quotes?

SteveL-MSFT (changes requested):

grammars/tree-sitter-ssh-server-config/grammar.js
tgauth: The full scenario the Rust parser is running into:

The existing parser has a test for:
`allowgroups group1 "group with spaces"`

However, the existing parser would parse this scenario:
`allowgroups group1 group2`

as a single string, `"group1 group2"` instead of `"group1" and "group2"`

To support both, I plan to update the Rust parser to be able to split "string" nodes by spaces but not "quoted string nodes"

It may be easier to consolidate the grammar and parser changes because this has broken the parser anyway

tgauth: another option would probably be to update the grammar to not declare space in `extras` but that may have broader impact

SteveL-MSFT: I think space in `extras` is fine to allow arbitrary whitespace, but the Rust parsing code should just handle the tokens returned by the grammar and not have to worry about the enclosing quotes and just the contents which can contain whitespace.  I think this should be possible by defining as (pseudo-grammar):

[keyword] [parameter]+

where [parameter] is `choice([string], [number], [yes/no], etc...)` and `[string]` is either a `bareString` or `quotedString`, but the parser just gets it as a `string`.

Issue #1047, opened 2025-08-12
Support `.bicepparam` file


Gijsreyn:
### Summary of the new feature / enhancement

As a user, I would like the ability to pass in a `.bicepparam` file to the `--parameters-file` subcommand. With Bicep support now added, it makes sense to include it through an extension.

SteveL-MSFT:
This should probably be a new `import` arg type to pass the parameter file passed to DSC to `bicep` without interpretation

Gijsreyn:
After taking a closer look, I think this can rather fall under the `discover` extension type. The reason is the output and potential passing it to the correct parameter command:

<img width="1591" height="168" alt="Image" src="https://github.com/user-attachments/assets/bd8f689e-3a01-4587-8e8c-c4917fa8b424" />

When calling the Az CLI or using Azure PowerShell, a `.bicepparam` file can be added to `TemplateFile` or `TemplateParameterFile`, introducing additional complexity in determining on the DSC side what the user actually wants. Instead of doing it directly in the engine, a script can determine what the user intended. For example, when the `using` statement is empty, it should be added as a parameter file. But this is just a thought.

PR #1282, opened 2025-11-21
Add support for ARM params file and Bicep params file formats

SteveL-MSFT:
# PR Summary

Add import extension to support `.bicepparam` file format
Allow parameters to be in simple or ARM parameters format, however, `reference` support is not added at this time, only `value` is allowed.
Because the `.bicepparam` results in a wrapped object, instead of adding direct support of it, I added a new `output` property to the `import` extension manifest that takes an expression to allow simple transformation which is used in the new `bicecpparameter` extension to get the internal object string.

Added new `stdout()` function that can only be used currently with `output` property expression which just returns the stdout of the executed command.

Fixed an issue where `import` capability wasn't being displayed correct via `dsc extension list`.
Added a bunch of new parser tracing to help diagnose issues.

Consolidated reading input files for params and config into one helper function.  This also takes care of reading from STDIN.  This also uses any `import` extension.
Supports 3 parameters formats:

- current simplified format:

```json
{
  "parameters": {
    "name": "value"
  }
}
```

- ARM format:

```json
{
  "parameters": {
    "name": {
      "value": "paramValue"
    }
}
```

Added helper function to take care of all of these formats
Added new `Microsoft.DSC.Extension/BicepParameters` import extension supporting `.bicepparam` file extension

## PR Context

Fix https://github.com/PowerShell/DSC/issues/1047
Fix https://github.com/PowerShell/DSC/issues/867

### DIFF
--- a/lib/dsc-lib/src/configure/mod.rs
+++ b/lib/dsc-lib/src/configure/mod.rs
@@ -836,9 +836,7 @@ impl Configurator {
 
         // process input parameters first
         if let Some(parameters_input) = parameters_input {
-            trace!("parameters_input: {parameters_input}");
             let input_parameters: HashMap<String, Value> = import_parameters(parameters_input)?;
-
             for (name, value) in input_parameters {
                 if let Some(constraint) = parameters.get(&name) {
                     debug!("Validating parameter '{name}'");
--- a/lib/dsc-lib/src/extensions/import.rs
+++ b/lib/dsc-lib/src/extensions/import.rs
@@ -89,9 +89,7 @@ impl DscExtension {
             if stdout.is_empty() {
                 info!("{}", t!("extensions.dscextension.importNoResults", extension = self.type_name));
             } else {
-                debug!("got stdout: {}", stdout);
                 if let Some(output) = &import.output {
-                    debug!("processing output: {}", output);
                     debug!("{}", t!("extensions.dscextension.importProcessingOutput", extension = self.type_name));
                     let mut parser = Statement::new()?;
                     let mut context = Context::new();


dsc/tests/dsc_parameters.tests.ps1
```diff
@@ -394,6 +394,53 @@ Describe 'Parameters tests' {
...
+            type: array
+          myInt:
+            type: int
+          myBool:
+            type: bool
+        resources:
+        - name: echo
+          type: Microsoft.DSC.Debug/Echo
+          properties:
+            output: "[concat(parameters('myString'), '-', parameters('myObject').prop1, '-', parameters('myArray')[0], parameters('myArray')[1], '-', string(parameters('myInt')), '-', string(parameters('myBool'))]"
```
Gijsreyn: Nit: Can you use `>-` like:

```yaml
output: >-
  [concat(
    parameters('myString'),
    '-',
    parameters('myObject').prop1,
    '-',
    parameters('myArray')[0],
    parameters('myArray')[1],
    '-',
    string(parameters('myInt')),
    '-',
    string(parameters('myBool'))
  )]
  ```

File: dsc/examples/hello_world.dsc.json [hunk 2]
```diff
@@ -0,0 +1,9 @@
+{
+  "$schema": "https://schema.management.azure.com/schemas/2019-04-01/deploymentParameters.json#",
+  "contentVersion": "1.0.0.0",
+  "parameters": {
+    "text": {
+      "value": "This is a parameterized hello world!"
+    }
+  }
+}
```
tgauth: Nit: add newline at EOF
```suggestion
}

```

tgauth: Actually, is this file required? I see the `hello_world.dsc.bicepparam` and `hello_world.dsc.bicep` used in the tests, so not sure if this is a test artifact

SteveL-MSFT: Actually, this was generated by `bicep` during my testing, it should be checked in.  Good catch, will remove.

SteveL-MSFT:
@tgauth can you take another look, made a fundamental change with how to handle `bicepparam` format.  I didn't like how the engine hard hardcoded knowledge about it when it's just an object wrapper around the ARM param object so created a new `output` property with a new `stdout()` function to extract it

PR #1284, opened 2025-11-21
Expand sshdconfig parser

tgauth:
# PR Summary
- updates sshdconfig Rust parser to support `match` blocks with tweak to tree-sitter grammar for clarity
- updates sshdconfig Rust parser to support `operator` nodes - add (+), remove (-), insert (^)
- updates sshdconfig Rust parser to serialize `booleans` to `true`/`false` instead of `yes`/`no`

### DIFF
--- a/resources/sshdconfig/src/error.rs
+++ b/resources/sshdconfig/src/error.rs
@@ -9,10 +9,10 @@ use thiserror::Error;
 pub enum SshdConfigError {
     #[error("{t}: {0}", t = t!("error.command"))]
     CommandError(String),
-    #[error("{t}: {0}", t = t!("error.fmt"))]
-    FmtError(#[from] std::fmt::Error),
     #[error("{t}: {0}", t = t!("error.envVar"))]
     EnvVarError(#[from] std::env::VarError),
+    #[error("{t}: {0}", t = t!("error.fmt"))]
+    FmtError(#[from] std::fmt::Error),
     #[error("{t}: {0}", t = t!("error.invalidInput"))]
     InvalidInput(String),
     #[error("{t}: {0}", t = t!("error.io"))]


resources/sshdconfig/src/parser.rs
```diff
@@ -268,4 +395,204 @@ mod tests {
...
+        assert!(ciphers.get("value").unwrap().is_array());
+
+        let macs = result.get("macs").unwrap().as_object().unwrap();
+        assert_eq!(macs.get("operator").unwrap(), &Value::String("-".to_string()));
+        assert!(macs.get("value").unwrap().is_array());
+
+        let kex = result.get("kexalgorithms").unwrap().as_object().unwrap();
+        assert_eq!(kex.get("operator").unwrap(), &Value::String("^".to_string()));
+        assert!(kex.get("value").unwrap().is_array());
+    }
```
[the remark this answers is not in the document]
tgauth: it's not a valid scenario for operators to appear on multiple lines for the same keywords

resources/sshdconfig/src/parser.rs
```diff
@@ -59,67 +59,185 @@ impl SshdConfigParser {
...
+                return Err(SshdConfigError::ParserError(t!("parser.failedToParseNode", input = input).to_string()));
+            }
+
+            match child.kind() {
+                "alpha" => {
+                    let Ok(text) = child.utf8_text(input_bytes) else {
+                        return Err(SshdConfigError::ParserError(t!("parser.failedToParseNode", input = input).to_string()));
+                    };
+                    key = Some(text.to_string());
+                }
```
SteveL-MSFT: This seems a bit confusing to me to read.  I think the grammar should be updated to have a field called `key` or something that is distinct from `alpha`

resources/sshdconfig/src/parser.rs
```diff
@@ -59,67 +59,185 @@ impl SshdConfigParser {
...
+                "boolean" | "string" => {
```
SteveL-MSFT: I wonder if internally we should treat `yes`/`no` booleans as actual booleans instead of strings and just render them back to `yes`/`no` when serializing?

resources/sshdconfig/src/parser.rs
```diff
@@ -132,8 +250,15 @@ impl SshdConfigParser {
             } else {
                 return Err(SshdConfigError::ParserError(t!("parser.keyNotRepeatable", key = key).to_string()));
             }
+        } else if is_repeatable {
+            // Initialize repeatable keywords as arrays
+            if let Value::Array(_) = value {
```
SteveL-MSFT: would it be slightly more readable to use `.is_array()`?

SteveL-MSFT (changes requested):

resources/sshdconfig/src/parser.rs
tgauth: Yeah - this can simplified. Modified to use the existing keyword/arg parsing by referencing each by their field name

SteveL-MSFT (approved):

PR #1307, opened 2025-12-05
Expand sshdconfig set capability

tgauth:
# PR Summary
- Modify `get` behavior to `warn` instead of `error` if any input is provided for filtering so DSC's synthetic test can be utilized 
- Add trace level and formatting options for easier debugging
- Add helper functions for formatting different argument types, including `match` blocks
- Expand `set` Pester tests to cover various argument types and invocation from DSC

### DIFF
--- a/resources/sshdconfig/src/util.rs
+++ b/resources/sshdconfig/src/util.rs
@@ -15,7 +15,7 @@ use crate::metadata::{MULTI_ARG_KEYWORDS_COMMA_SEP, SSHD_CONFIG_DEFAULT_PATH_UNI
 use crate::parser::parse_text_to_map;
 
 #[derive(Debug, Deserialize)]
-struct MatchBlockFmt {
+struct MatchBlock {
     criteria: Map<String, Value>,
     #[serde(flatten)]
     contents: Map<String, Value>,
@@ -343,7 +343,7 @@ fn get_bool_or_default(map: &mut Map<String, Value>, key: &str, default: bool) -
 }
 
 fn format_match_block(match_obj: &Value) -> Result<String, SshdConfigError> {
-    let match_block = match serde_json::from_value::<MatchBlockFmt>(match_obj.clone()) {
+    let match_block = match serde_json::from_value::<MatchBlock>(match_obj.clone()) {
         Ok(result) => {
             result
         }


resources/sshdconfig/src/main.rs
```diff
@@ -29,10 +29,30 @@ const EXIT_SUCCESS: i32 = 0;
 const EXIT_FAILURE: i32 = 1;
 
 fn main() {
-    enable_tracing();
-
     let args = Args::parse();
 
+    let trace_level = match &args.trace_level {
```
SteveL-MSFT: Would be better to encapsulate this logic within `enable_tracing()`

resources/sshdconfig/src/util.rs
```diff
@@ -208,3 +309,46 @@ fn get_bool_or_default(map: &mut Map<String, Value>, key: &str, default: bool) -
         Ok(default)
     }
 }
+
+fn format_match_block(match_obj: &Map<String, Value>) -> Result<Option<String>, SshdConfigError> {
+    let mut lines = Vec::new();
+
+    let Some(criteria_value) = match_obj.get("criteria") else {
```
SteveL-MSFT: Since it looks like the `match` object has a specific structured syntax, did you consider a match struct that this Value gets deserialized into instead of manually parsing?

resources/sshdconfig/src/util.rs
```diff
@@ -23,7 +23,28 @@ use crate::parser::parse_text_to_map;
 /// # Errors
 ///
 /// This function will return an error if it fails to initialize tracing.
-pub fn enable_tracing(trace_level: &TraceLevel, trace_format: &TraceFormat) {
+pub fn enable_tracing(trace_level: Option<TraceLevel>, trace_format: &TraceFormat) {
```
SteveL-MSFT: You should have `trace_level` be `Option<&TraceLevel>` to avoid the clone().  Pass in as `&args.trace_level.as_deref()`

resources/sshdconfig/src/util.rs
```diff
@@ -13,6 +14,13 @@ use crate::inputs::{CommandInfo, Metadata, SshdCommandArgs};
 use crate::metadata::{MULTI_ARG_KEYWORDS_COMMA_SEP, SSHD_CONFIG_DEFAULT_PATH_UNIX, SSHD_CONFIG_DEFAULT_PATH_WINDOWS};
 use crate::parser::parse_text_to_map;
 
+#[derive(Debug, Deserialize)]
+struct MatchBlockFmt {
```
SteveL-MSFT: I don't think the `Fmt` suffix is needed?

PR #1309, opened 2025-12-08
(GH-538) Define derive macro for `DscRepoSchema`

michaeltlombardi:
# PR Summary

This change:

- Defines a new `dsc-lib-jsonschema-macros` crate to publish procedural macros related to JSON Schemas.
- Defines a derive macro for the `DscRepoSchema` trait.
- Provides integration testing for the new macro.

This change doesn't migrate the implementations for the `DscRepoSchema` trait from manual to macro-derived in `dsc-lib`. Those changes should follow separately.

## PR Context

Prior to this change, adding the `DscRepoSchema` trait to any type required manual implementation. This required a lot of boilerplate and wasn't ergonomic. With the creation of a derive macro, we can annotate the type definitions with an attribute and minimize the additional code we need to write.

### DIFF
--- a/build.helpers.psm1
+++ b/build.helpers.psm1
@@ -471,6 +471,19 @@ function Install-TreeSitter {
         [switch]$UseCFS
     )
 
+    begin {
+        $arguments = @(
+            'install',
+            'tree-sitter-cli',
+            '--version', '0.25.10'
+        )
+
+        if ($UseCFS) {
+            $arguments += '--config'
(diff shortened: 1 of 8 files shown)


SteveL-MSFT (approved):

michaeltlombardi:
CI failures are unrelated to this change - there seems to be some difference in the generated bindings for grammars in CI that I haven't been able to replicate locally yet.

PR #1314, opened 2025-12-11
Fix how clippy is used in build

SteveL-MSFT:
# PR Summary

- Looks like the build helper wasn't using pedantic mode, so added that
- Declare the generated code for grammars in the project data files to be pedantic unclean
- Tree-sitter auto-changed the grammar.js
- Removed unused use declaration in dsc lib

### DIFF
--- a/build.helpers.psm1
+++ b/build.helpers.psm1
@@ -1300,7 +1300,7 @@ function Export-GrammarBinding {
             } finally {
                 Pop-Location
             }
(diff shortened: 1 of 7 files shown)


michaeltlombardi (approved):
This change requires an update to the JSON Schema for the `.project.data` files, but I'll submit that separately after this merges. Otherwise, LGTM!

PR #1317, opened 2025-12-13
Implement routing to delete operation for resources with _exist set to false

### LINKED ISSUE #1268: Calling `set` with `_exist` is `false` does not run `delete`
### Prerequisites

### Summary

Resources that support both `set` and `delete` does not call the correct operation when `_exist` is set to `false`.

I know if I were to create a config doc it would be called correctly but with rapid testing, I just change the payload from creating to deleting by setting the `_exist` property instead of changing the `set` to `delete`.

### Steps to reproduce

1. Create resource with `set` and `delete` operations and `_exist` canonical property.
2. Call `dsc resource set` with `(@{ _exist = $false} | ConvertTo-Json)`

### Expected behavior

```console
The resource is deleted
```

### Actual behavior

```console
The resource is not deleted
```

### Error details

```console

```

### Environment data

```PowerShell
7.6.0-preview.5
```

### Version

dsc 3.2.0-preview.8

### Visuals

_No response_

Gijsreyn:
# PR Summary

Fixes `dsc resource set` to route to the delete operation when `_exist` is set to `false` for resources that have the `Delete` capability but not `SetHandlesExist`.

The implementation mirrors the existing logic from configuration documents and reuses the `get_diff` function for calculating the `changedProperties`. I've decided to manually construct the `SetResult` response since `delete()` returns `()` rather than state info as a design decision.

## PR Context

Fixes #1268.

### DIFF
--- a/dsc/tests/dsc_resource_set.tests.ps1
+++ b/dsc/tests/dsc_resource_set.tests.ps1
@@ -14,4 +14,10 @@ Describe 'Invoke a resource set directly' {
         $out.afterState.version | Should -BeExactly '1.1.2'
         $out.changedProperties | Should -BeNullOrEmpty
     }
+
+    It '_exist false routes to delete operation' {
+        $out = dsc -l trace resource set --resource Test/Delete --input '{"_exist": false}' 2>&1
+        $LASTEXITCODE | Should -Be 0
+        $out | Out-String | Should -Match 'Routing to delete operation because _exist is false'
+    }
 }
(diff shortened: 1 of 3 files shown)


dsc/tests/dsc_resource_set.tests.ps1
```diff
@@ -14,4 +14,10 @@ Describe 'Invoke a resource set directly' {
         $out.afterState.version | Should -BeExactly '1.1.2'
         $out.changedProperties | Should -BeNullOrEmpty
     }
+
+    It '_exist false routes to delete operation' {
```
SteveL-MSFT, line 18: Is there a test case to handle what-if for this scenario?

Gijsreyn: Is `--what-if` not only for `dsc config`? I can't seem to find them on `dsc resource` options.

SteveL-MSFT: I believe you are correct, so nevermind.

Issue #833, opened 2025-05-24
Output from a class-based resource interrupts the parsing process


johlju:
### Prerequisites

- [x] Write a descriptive title.
- [x] Make sure you are able to repro it on the latest version
- [x] Search the existing issues.

### Summary

I am experiencing a problem with how output (for example a warning from `Write-Warning`) are handled during module import. I think any output to output streams should be handled correctly and not interrupt the resource parsing process. But actually, such output (like warnings) are being interpreted as output and included in the parsing of results, causing DSC to fail parsing the result from Get operation.

This could be related to (or resolved by) issue https://github.com/PowerShell/DSC/issues/423.

### Steps to reproduce

1. Install latest module SqlServerDsc (that include the class-based resource) into the correct folder
2. Run the following:
    ```powershell
     $desiredParameters = @{
         InstanceName = 'PBIRS'
         AcceptLicensingTerms = $true
         Action = 'Install'
         MediaPath = '<Path to any existing executable on local machine>'
         InstallFolder = [System.IO.Path]::GetTempPath()
         Edition = 'Developer'
         SuppressRestart = $true
         VersionUpgrade = $true
     }

      dsc --trace-level trace resource get --resource SqlServerDsc/SqlRSSetup --output-format pretty-json --input ($desiredParameters | ConvertTo-Json -Compress)
      ```
3. Watch the error `407: Failed to parse resource` where it includes an output from `Write-Warning` that the SqlServerDsc module output during module import.

### Expected behavior

```console
DSC executable should parse the result from a Get operation regardless what is output on output streams.
```

### Actual behavior

```console
Fails to parse the result becuase it includes the warning stream.
```

### Error details

```console
2025-05-23T05:27:02.114426Z  WARN dsc_lib::discovery::command_discovery: 407: Failed to parse resource: WARNING: Failed to find a dependent module. Unable to run SQL Server commands or use SQL Server types. Please install  -> expected value at line 1 column 1
2025-05-23T05:27:02.114443Z  WARN dsc_lib::discovery::command_discovery: 407: Failed to parse resource: one of the preferred SMO modules or the SQLPS module, then try to import SqlServerDsc again. -> expected value at line 1 column 1
...
2025-05-23T05:27:43.375232Z ERROR dsc::resource_command: 44: Error: Operation: Failed to parse JSON from 'get': executable = 'powershell' stdout = 'WARNING: Failed to find a dependent module. Unable to run SQL Server commands or use SQL Server types. Please install 
[...]

PR #1319, opened 2025-12-15
RFC0001: Class-based PSDSC resource contract for DSC

theJasonHelmick:


gaelcolas:
I'm unsure about the Attributes and metadata, mostly that those should not be relied upon by the adapter, since it should be defined in the resource manifest.
@michaeltlombardi don't we think that should be done by the resource manifest anyway, and the attributes are only relevant to generate the manifest? (so they're out of scope for this RFC, but the manifest metadata should be in-scope and clarified here?)

Another thing I think is missing is how we could get the same outcome as the Machine Config "reasons".
I think the return object `[System.Tuple[bool, <ResourceClass>, String[]]]` is simplistic, or should at least explain how the `[string[]]` are used/displayed through the adapter and engine.

Gijsreyn:
After going through the RFC a couple of times, some small remarks from my understanding of it:

- In the skeletons, I still see `Ensure` is (still) the word. Even though many DSC resources will use this (also for backwards compatibility), I think it would be better if they followed the semantics of Microsoft DSC to start implementing `Exist`, knowing that the `DscResourceProperty()` attribute isn't going to be an optional one in this case. If "they" need to reshape the class anyway with `static` methods, why not include this one directly...
- In case of classes going through the adapter, the `Validate` needs to be properly implemented to validate the schema (I might be missing something here as I couldn't imagine the part on: _If the resource doesn't implement this method, DSC generates a JSON Schema by inspecting the resource class itself._) .
- I think the RFC also needs to include the logging part, meaning: how can resource authors hook into the logging semantics of DSC (especially because of issue: https://github.com/PowerShell/DSC/issues/833)

gaelcolas:
Todo: update with the what-if scenario #1361

PR #1323, opened 2025-12-16
(MAINT) Fix packaging for `zip` and `tgz`

michaeltlombardi:
# PR Summary

This change:

- Corrects the `Import-DscBuildData` function to retain the root project when refreshing the project data.
- Ensures the correct path is returned for `TgzTarget`.
- Ensures the `Build-RustProject` function selects the correct workspace member group when the user specifies an architecture.
- Ensures that `Build-DscPackage` calls `Build-DscZipPackage` with the required parameters when package type is `zip`.
- Fixes the `clean` section of the `build.ps1` script to only reset the Rust environment variables if the caching variable isn't null.

This change was tested locally to produce `.tar.gz` (WSL) and `.zip` (Windows 11) packages without errors.

## PR Context

Prior to this change, there were several issues that prevented the use of the `build.ps1` script to package builds for `zip` and `tgz`:

- The `Import-DscBuildData` function erroneously dropped the `root` project, which is used to add the `NOTICE.txt` file to the `bin` folder.
- The `Get-ArtifactDirectoryPath` failed to correctly define the `TgzTarget` property due to a mistaken key name (invalid `$` prefix).
- The `Build-RustProject` function had a bug in the implementation for specified (rather than `current`) architectures where incorrect casing caused the workspace member group to be an empty array.
- The `Build-DscPackage` function called `Build-DscTgzPackage` for the `zip` package type without passing any build parameters.
- The `clean` section of the `build.ps1` script always tried to reset the Rust environment variables, even when they hadn't been cached.

### DIFF
--- a/build.data.json
+++ b/build.data.json
@@ -218,10 +218,10 @@
       "IsRust": true
     },
     {
-        "Name": "dsc-lib-jsonschema-macros",
-        "Kind": "Library",
-        "RelativePath": "lib/dsc-lib-jsonschema-macros",
-        "IsRust": true
+      "Name": "dsc-lib-jsonschema-macros",
+      "Kind": "Library",
+      "RelativePath": "lib/dsc-lib-jsonschema-macros",
+      "IsRust": true
     },
     {
       "Name": "dsc-lib-osinfo",
--- a/build.helpers.psm1
+++ b/build.helpers.psm1
@@ -181,7 +181,10 @@ function Import-DscBuildData {
         | ConvertFrom-Json -AsHashtable
 
         if ($RefreshProjects) {
-            $data.Projects = Get-DscProjectData
+            [DscProjectDefinition[]]$rootProject = $data.Projects.Where({
+                $_.Name -eq 'root'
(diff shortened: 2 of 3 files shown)


andyleejordan (approved):
Thanks Mikey! This is tested and working with the caveat that it requires `-Architecture` to specified in both the build and release steps. As in, if left out, the default `current` doesn't work, and that seems to be a different (if related) issue.

michaeltlombardi:
> Thanks Mikey! This is tested and working with the caveat that it requires `-Architecture` to specified in both the build and release steps. As in, if left out, the default `current` doesn't work, and that seems to be a different (if related) issue.

My understanding is that was by-design when I was migrating from the old build script. What I would really like to change is needing separate invocations to build and package, at least for local development - functionally, not defaulting `-SkipBuild` to `$false` when `PackageType` is defined. I left it alone for now, but that would be more convenient.

I can also look into editing the behavior for requiring the `-Architecture` parameter in this or a follow up PR, too.

PR #1324, opened 2025-12-16
Pass -UseCFS and -UseCFSAuth to Set-CargoEnvironment

andyleejordan:
So that I can actually add packages to the feed. What a headache.


michaeltlombardi (approved):

michaeltlombardi:
Oops, mea culpa!

PR #1327, opened 2025-12-16
Add sshdconfig match to export

tgauth:
# PR Summary
- fix `get` bug where the command was not ignoring any input filtering
- add support to `get` and `export` regarding  `Match` directives
- add warning to `get` and `export` regarding `Include` directives
- add resource-level tests for above `get` and `export` scenarios 
- update tree-sitter-ssh-server-config grammar/tests for more flexibility with comments and empty lines
- enable sshd_config tests on all platforms (when elevated and sshd is available)

## PR Context
- These tweaks to `get` will be required to enable the `set` & `clobber: false` scenario

### DIFF
--- a/grammars/tree-sitter-ssh-server-config/test/corpus/valid_expressions.txt
+++ b/grammars/tree-sitter-ssh-server-config/test/corpus/valid_expressions.txt
@@ -611,7 +611,7 @@ parse multiple comments and multiple lines
 ====
 # comment line 1
 # comment line 2
-# comment lin3 3
+# comment line 3
 
 
 passwordauthentication yes


grammars/tree-sitter-ssh-server-config/grammar.js
```diff
@@ -18,20 +18,23 @@ export default grammar({
 
     // check for an empty line that is just a /n character
     _empty_line: $ => '\n',
-    comment: $ => /#.*\n/,
+    comment: $ => seq(/#.*/, '\n'),
```
SteveL-MSFT: I'm a bit surprised you would include the newline here.  I would think a comment would be a token that starts with a `#` and all arbitrary characters after it, so it doesn't really matter if it's a `comment` or `_inline_comment`.

Look here as an example: https://github.com/airbus-cert/tree-sitter-powershell/blob/main/grammar.js#L15

Looks like you can just define what a `comment` is and put it under `extras` which gets ignored I believe

tgauth: Adding `comment` to `extras` works - thanks! I had gotten too focused on defining it similar to `keyword`

grammars/tree-sitter-ssh-server-config/grammar.js
```diff
@@ -11,27 +11,26 @@ const PREC = {
 export default grammar({
   name: 'ssh_server_config',
 
-  extras: $ => [' ', '\t', '\r'],
+  extras: $ => [' ', '\t', '\r', $.comment],
 
   rules: {
-    server_config: $ => seq(repeat(choice($._empty_line, $.comment, $.keyword)), repeat($.match)),
+    server_config: $ => seq(repeat(choice($._new_line, $.keyword)), repeat($.match)),
```
SteveL-MSFT, line 17: Can't `_new_line` just be in extras as well?

SteveL-MSFT (changes requested):

grammars/tree-sitter-ssh-server-config/grammar.js
tgauth: It's a required terminator for the `keyword` sequence on l. 27, so I don't think so. It's required in the sequence to differentiate between the end of a keyword's `arguments` and the start of the next keyword since they could both just be alphabetical characters

SteveL-MSFT: I see, I guess it's ok then

PR #1330, opened 2025-12-18
gRPC server for Bicep

andyleejordan:
This adds a new binary `dscbicep` to the DSC build which is a small, self-contained Rust executable hosting a gRPC server. The methods are defined in `bicep.proto` from [Bicep's extensibility APIs](https://github.com/Azure/bicep/blob/main/src/Bicep.Local.Rpc/extension.proto) (hence the addition of `protoc` as a build dependency). It's actually fairly similar to the MCP server, in that it's simply invoking find/get/set/delete resource functions via `dsc-lib`.

Using [`bicep local-deploy`](https://github.com/Azure/bicep/blob/main/docs/experimental/local-deploy.md) with this is actually done through a "Bicep extension" as defined in the project [bicep-types-dsc](https://github.com/microsoft/bicep-types-dsc). In essence, it's an OCI artifact with two things: the type information in a big JSON file and this `dscbicep` binary. When all put together and hosted on an OCI registry, Bicep will auto-restore the extension and just deploy the DSC resources defined in any arbitrary set of Bicep files, handling all Bicep functionality and orchestration itself.


.github/workflows/rust.yml
```diff
@@ -26,6 +26,10 @@ jobs:
     runs-on: ${{matrix.platform}}
     steps:
       - uses: actions/checkout@v5
+      - name: Add WinGet links to PATH
+        if: matrix.platform == 'windows-latest'
+        run: |
+          "$env:LOCALAPPDATA\\Microsoft\\WinGet\\Links" | Out-File -Append -Encoding utf8 -FilePath $env:GITHUB_PATH
```
andyleejordan, line 32: This was necessary for Protobuf because `protoc.exe` is made available by WinGet in this local app data path, which on GitHub images is not already in the PATH, and that is unlike the rest of our dependencies (Node.js is already installed system-wide and so in the PATH, tree-sitter isn't installed by installs via Cargo, this was our first installation on GitHub Actions using WinGet that doesn't run a system installer).

.pipelines/DSC-Official.yml
```diff
@@ -321,7 +321,7 @@ extends:
             ob_restore_phase: true
         - pwsh: |
             apt update
-            apt -y install musl-tools rpm dpkg build-essential
+            apt -y install musl-tools rpm dpkg build-essential protobuf-compiler
```
andyleejordan, line 324: I only have a loose understanding of the state of the OneBranch pipeline for this project. The build module will run the same apt install (and that is the correct package name) but I'm inferring this is the "restore" phase where it needs to be run? Or is this just for bootstrapping Rust? This package _does_ eventually have to be installed.

dsc-bicep-ext/proto/bicep.proto
```diff
@@ -0,0 +1,66 @@
+syntax = "proto3";
```
andyleejordan, line 5: I'd put a header and a comment in this file except that it's verbatim copied from Bicep.

dscbicep/src/main.rs
```diff
@@ -0,0 +1,425 @@
...
+        let mut dsc = DscManager::new();
+        let Some(resource) = dsc.find_resource(&DiscoveryFilter::new(&resource_type, version.as_deref(), None)).unwrap_or(None) else {
+            return Err(Status::not_found("Resource not found"));
+        };
+
+        let SetResult::Resource(result) = resource
+            .set(&properties, false, &ExecutionKind::Actual)
+            .map_err(|e| Status::aborted(e.to_string()))?
+        else {
+            return Err(Status::unimplemented("Group resources not supported"));
```
andyleejordan: This won't in general be hit due to the extensibility API calling this function on a per resource basis, but I have to unwrap the `SetResult` return struct so that's the error to satisfy the typing of the function.

dsc-bicep-ext/src/main.rs
```diff
@@ -0,0 +1,425 @@
...
+
+        Ok(Response::new(LocalExtensibilityOperationResponse {
+            resource: Some(proto::Resource {
+                r#type: resource_type,
+                api_version: version,
+                identifiers: properties,
+                properties: result.after_state.to_string(),
+                status: None,
+            }),
+            error_data: None,
```
andyleejordan, line 88: Once in use I suspect we may want to transform the error data a little better, though returning `Err` above works just fine.

dsc-bicep-ext/src/main.rs
```diff
@@ -0,0 +1,425 @@
...
+
+        tracing::debug!("Preview called for {resource_type}@{version:?}: {properties}");
+
+        let mut dsc = DscManager::new();
+        let Some(resource) = dsc.find_resource(&DiscoveryFilter::new(&resource_type, version.as_deref(), None)).unwrap_or(None) else {
+            return Err(Status::not_found("Resource not found"));
+        };
+
+        let SetResult::Resource(result) = resource
+            .set(&properties, false, &ExecutionKind::WhatIf)
```
andyleejordan, line 127: `WhatIf` is the only difference between this and the above, as that's the difference between set and preview. A reasonable ask for Bicep is to finish wiring up "what-if" in `local-deploy` though I doubt it's a high priority for any of us (@anthony-c-martin).

dsc-bicep-ext/src/main.rs
```diff
@@ -0,0 +1,425 @@
...
+        else {
+            return Err(Status::unimplemented("Group resources not supported"));
+        };
+
+        Ok(Response::new(LocalExtensibilityOperationResponse {
+            resource: Some(proto::Resource {
+                r#type: resource_type,
+                api_version: version,
+                identifiers: properties,
+                properties: result.after_state.to_string(),
```
andyleejordan, line 140: I lied, this is the other difference, as the result of a `WhatIf` has before/after properties, we're mapping back the `after_state` bag.

dsc-bicep-ext/src/main.rs
```diff
@@ -0,0 +1,423 @@
...
+        let Some(resource) = dsc.find_resource(&DiscoveryFilter::new(&resource_type, version.as_deref(), None)).unwrap_or(None) else {
+            return Err(Status::not_found("Resource not found"));
+        };
+
+        let GetResult::Resource(result) = resource
+            .get(&identifiers)
+            .map_err(|e| Status::aborted(e.to_string()))?
+        else {
+            return Err(Status::unimplemented("Group resources not supported"));
+        };
```
andyleejordan, line 188: This is where things really started to differ from DSC-native config management. This function is called for any resource declaration in Bicep that uses the `existing` keyword...which can be mixed in. Fundamentally, as shown in my demo, you can do something like "get the registry resource that matches these properties (and then the key path and value name) and pass the output to this next resource" meaning Bicep will infer the resource dependency, execute the get, retrieve the output, and then pass it to the resource in which its used.

E.g. https://github.com/microsoft/bicep-types-dsc/blob/bd8ea0e1e376ca751a9964402f31427d1ec52eec/windows.bicep#L28-L47

```bicep
resource currentRegistryKey 'Microsoft.Windows/Registry@1.0.0' existing = {
  keyPath: keys[^1].path
  valueName: keys[^1].name
}

var current = max(currentRegistryKey.valueData.?DWord ?? 0, 0)

var valueData = darkMode == 'toggle'
  ? (current + 1) % 2 // flip the bit
  : darkMode == 'on' ? 0 : 1

resource registryThemeKeys 'Microsoft.Windows/Registry@1.0.0' = [
  for key in keys: {
    keyPath: key.path
    valueName: key.name
    valueData: {
      DWord: valueData
    }
  }
]
```

dsc-bicep-ext/src/main.rs
```diff
@@ -0,0 +1,423 @@
...
+        request: Request<ResourceSpecification>,
+    ) -> Result<Response<LocalExtensibilityOperationResponse>, Status> {
+        let spec = request.into_inner();
+        let resource_type = spec.r#type;
+        let version = spec.api_version;
+        let properties = spec.properties;
+
+        tracing::debug!("CreateOrUpdate called for {resource_type}@{version:?}: {properties}");
+
+        let mut dsc = DscManager::new();
```
andyleejordan, line 57: Question for @michaeltlombardi should this perhaps scope to the process? That is, move it from each function call into main and make it a singleton. I'm not sure what all this constructer entails.

dsc-bicep-ext/src/main.rs
```diff
@@ -0,0 +1,423 @@
...
+        else {
+            return Err(Status::unimplemented("Group resources not supported"));
+        };
+
+        Ok(Response::new(LocalExtensibilityOperationResponse {
+            resource: Some(proto::Resource {
+                r#type: resource_type,
+                api_version: version,
+                identifiers: identifiers,
+                properties: result.actual_state.to_string(),
```
andyleejordan, line 195: Again unwrapping the correct properties bag that reflects the state of the resource we're getting, as DSC's result struct is actually several property bags.

dsc-bicep-ext/src/main.rs
```diff
@@ -0,0 +1,423 @@
...
+        );
+
+        let mut dsc = DscManager::new();
+        let Some(resource) = dsc.find_resource(&DiscoveryFilter::new(&resource_type, version.as_deref(), None)).unwrap_or(None) else {
+            return Err(Status::not_found("Resource not found"));
+        };
+
+        resource
+            .delete(&identifiers)
+            .map_err(|e| Status::aborted(e.to_string()))?;
```
andyleejordan, line 238: My understanding is that DSC's "deletions" are often done via `set` with `_ensure: false`; however, resources _can_ be deleted, and this is wired up. In Bicep this is doing via [deployment stacks](https://learn.microsoft.com/en-us/azure/azure-resource-manager/bicep/deployment-stacks?tabs=azure-powershell#control-detachment-and-deletion). Open design questions with @SydneyhSmith how that will look.

dscbicep/src/main.rs
```diff
@@ -0,0 +1,423 @@
...
+
+    async fn get_type_files(
+        &self,
+        _request: Request<Empty>,
+    ) -> Result<Response<TypeFilesResponse>, Status> {
+        tracing::debug!("GetTypeFiles called");
+
+        // TODO: Return actual Bicep type definitions...yet the extension already has these?
+        // Perhaps this is where we can dynamically get them from the current system.
+        Err(Status::unimplemented("GetTypeFiles not yet implemented"))
```
andyleejordan: This TODO is exciting as this is the dynamic resource discovery @SteveL-MSFT has been asking for, though since DSC is written in Rust I won't have a Bicep types package to depend on (they're available in C# and TypeScript), but this will be great to do.

dsc-bicep-ext/src/main.rs
```diff
@@ -0,0 +1,423 @@
...
+    #[arg(long)]
+    socket: Option<String>,
+
+    /// The named pipe to connect on (Windows)
+    #[arg(long)]
+    pipe: Option<String>,
+
+    /// The HTTP address to listen on (e.g., 127.0.0.1:50051)
+    #[arg(long)]
+    http: Option<String>,
```
andyleejordan, line 281: Not how Bicep connects, but how this can be executed against without Bicep at all. Useful for debugging, and is why the `tonic-reflection` service is added below. You can bring up a gRPC web UI to execute resource requests live.

dsc-bicep-ext/src/main.rs
```diff
@@ -0,0 +1,423 @@
...
+
+    /// Wait for debugger to attach before starting
+    #[arg(long)]
+    wait_for_debugger: bool,
```
andyleejordan, line 285: There's a TODO for Bicep to implement this flag. It's in the spec but not yet wired up. For attaching the debugger to the process, I check a custom environment variable instead for now.

dsc-bicep-ext/src/main.rs
```diff
@@ -0,0 +1,423 @@
...
+        let uds = UnixListener::bind(&socket_path)?;
+        let uds_stream = UnixListenerStream::new(uds);
+
+        Server::builder()
+            .add_service(BicepExtensionServer::new(service))
+            .serve_with_incoming(uds_stream)
+            .await?;
+
+        return Ok(());
+    }
```
andyleejordan, line 322: Standard and easy!

dsc-bicep-ext/src/main.rs
```diff
@@ -0,0 +1,423 @@
...
+            }
+        };
+
+        Server::builder()
+            .add_service(BicepExtensionServer::new(service))
+            .serve_with_incoming(incoming)
+            .await?;
+
+        return Ok(());
+    }
```
andyleejordan, line 434: See the first comment. This works, it's just a lot more verbose than I would've liked as I did indeed have to relay on Copilot to generate the boilerplate necessary to map one interface to another. Since the Rust community has been arguing about Unix socket support on Windows for half a decade, it's not coming, so we have to use named pipes, which is more Windows-native, but not nearly as well supported in Rust/tonic/tokio. If this can be simplified, please let me know. I was just happy to get it working.

dsc-bicep-ext/src/main.rs
```diff
@@ -0,0 +1,423 @@
...
+    let reflection_service = tonic_reflection::server::Builder::configure()
+        .register_encoded_file_descriptor_set(proto::FILE_DESCRIPTOR_SET)
+        .build_v1()
+        .unwrap();
+
+    Server::builder()
+        .add_service(reflection_service)
+        .add_service(BicepExtensionServer::new(service))
+        .serve(addr)
+        .await?;
```
andyleejordan, line 456: The aforementioned "debug" HTTP server where requests can be sent to DSC over gRPC without Bicep at all (which is really neat).

dsc-bicep-ext/build.rs
```diff
@@ -0,0 +1,14 @@
...
+// Licensed under the MIT License.
+
+use std::{env, path::PathBuf};
+
+fn main() -> Result<(), Box<dyn std::error::Error>> {
+    let descriptor_path = PathBuf::from(env::var("OUT_DIR").unwrap()).join("bicep.bin");
+
+    tonic_prost_build::configure()
+        .build_client(false)
+        .file_descriptor_set_path(&descriptor_path)
```
andyleejordan, line 11: This is used for the reflection in the HTTP server.

dsc-bicep-ext/Cargo.toml
```diff
@@ -0,0 +1,25 @@
+[package]
+name = "dscbicep"
+version = "0.1.0"
+edition = "2021"
+
+[[bin]]
+name = "dscbicep"
+path = "src/main.rs"
```
andyleejordan, line 8: I have no strong opinions on the name, version, or file structure other than suggesting that changes be made after this PR, as the extension has to incorporate the binary.

lib/dsc-lib/src/dscresources/command_resource.rs
```diff
@@ -765,24 +765,39 @@ pub fn invoke_command(executable: &str, args: Option<Vec<String>>, input: Option
...
+        },
+        // Otherwise create a new runtime
+        Err(_) => {
+            tokio::runtime::Builder::new_multi_thread()
+                .enable_all()
+                .build()
+                .unwrap()
+                .block_on(run_async)
+        }
+    }
```
andyleejordan, line 800: This is the only real change inside of DSC. Since under gRPC there's _already_ a runtime, we need to re-use it (and tell it that we're spawning a blocking task so that it manages the threads correctly and doesn't deadlock). Whereas when the DSC CLI invokes these resources, it spins up a runtime just for the sake of threading the processes (yes I know that sounds funny).

.github/workflows/rust.yml
SteveL-MSFT: I think you should have this environment change in the build.helpers.psm1 for protobuf tool install instead of a one-off here

.pipelines/DSC-Official.yml
SteveL-MSFT: This installs all the tools not already available in the OneBranch container so doing it here seems appropriate

dsc-bicep-ext/proto/bicep.proto
SteveL-MSFT: Should probably have a comment indicating where this was copied from

dsc-bicep-ext/src/main.rs
SteveL-MSFT: I think it would be better to create it as a singleton

dscbicep/src/main.rs
SteveL-MSFT: The only group resource we know people are using is the Assertion resource

dscbicep/src/main.rs
```diff
@@ -0,0 +1,423 @@
...
+    async fn create_or_update(
+        &self,
+        request: Request<ResourceSpecification>,
+    ) -> Result<Response<LocalExtensibilityOperationResponse>, Status> {
+        let spec = request.into_inner();
+        let resource_type = spec.r#type;
+        let version = spec.api_version;
+        let properties = spec.properties;
+
+        tracing::debug!("CreateOrUpdate called for {resource_type}@{version:?}: {properties}");
```
SteveL-MSFT: All the text for traces need to be localizable, see the i18n crate used in dsc_lib.  Maybe Copilot can do this for you?

dsc-bicep-ext/src/main.rs
SteveL-MSFT: This might be where the design is inverted between bicep and DSC.  In DSC, `delete` is called if `set` doesn't support `_exist`, but in this case, `delete` might have to call `set` with `_exist=false` to get the same operation.  Although maybe this is something the DSC engine could perform.

dscbicep/src/main.rs
SteveL-MSFT: Since the types are only useful for Bicep, perhaps write a small CLI in C# (AOT) to generate the type information and this exe can call out to that?  This can be done later

dsc-bicep-ext/src/main.rs
```diff
@@ -0,0 +1,423 @@
...
+
+        Server::builder()
+            .add_service(BicepExtensionServer::new(service))
+            .serve_with_incoming(incoming)
+            .await?;
+
+        return Ok(());
+    }
+
+    // Default to HTTP server on [::1]:50051 if no transport specified
```
SteveL-MSFT, line 436: Where did 50051 come from?  Is there a risk of conflict?

.github/workflows/rust.yml
andyleejordan: This is specific to GitHub Actions. When you install Protobuf via WinGet on Windows normally, `protoc` is found in the `PATH`. It appears that GitHub Actions Windows images don't correctly include the WinGet path.

dsc-bicep-ext/src/main.rs
andyleejordan: It would make the most sense to me that dsc-lib's implementation of `set` simply handles a set with `_exist=false` (does that mean it delegates to `delete` internally?), and `delete` handles deleting if available or delegates to `set(_exist=false)`. At the library layer both functions should "just work" from a DSC user's perspective.

dscbicep/src/main.rs
andyleejordan: Yeah, definitely an option to explore!

dsc-bicep-ext/src/main.rs
andyleejordan: That's the default gRPC port number (and this is only used in debug/test scenarios, not when connecting via Bicep over a socket or named pipe).

dsc-bicep-ext/src/main.rs
andyleejordan: I briefly just attempted that, problem is sharing it among the async functions in a multi-threaded process we'd need to lock the singleton on use. Looking at what `DscManager` does I think it makes sense that it exists per invocation so parallel execution is handled without issue. The good news is that IIRC it caches the resources it's found, no? So fairly idemopotent within one process instance?

dscbicep/src/main.rs
andyleejordan: Done.

dsc-bicep-ext/proto/bicep.proto
andyleejordan: Done.

andyleejordan:
@SteveL-MSFT changes implemented except `DscManager` is still per-function since it would otherwise need to be locked and therefore seems to me like it would constrain parallel execution, please let me know if I'm misunderstanding that though. I tried sharing an instance in the `BicepExtensionService` struct but using it meant putting it into a mutex.

.github/workflows/rust.yml
michaeltlombardi: I think we could check for the `GITHUB_ACTIONS` environment variable if we want to move this into the install function - that variable is always defined as `true` when running in GHA and unlikely to be accidentally set for local development.

Cargo.toml
```diff
@@ -4,6 +4,7 @@ resolver = "2"
 # the path to a crate.
 members = [
     "dsc",
+    "dscbicep",
```
michaeltlombardi: Nit: I know we already have some binaries that follow this convention of using `dsc` as a prefix without a separator, but it _looks weird_ and doesn't follow the general convention for related binaries in rust, which typically follows `<prefix>-<suffix>` for binaries and libraries both, though this is _just_ a convention.

I know I'd rather write `dsc-bicep` in the docs than `dscbicep` and if we're going to name it differently, better to do it now than later.

What I would _really_ prefer is for it to be named `dsc-bicep-ext` to clarify that it's not remotely meant as a standalone application outside of either Bicep or DSC

.github/workflows/rust.yml
andyleejordan: Or it could just be in the GitHub Actions script 😅

dsc-bicep-ext/Cargo.toml
michaeltlombardi: I know we already have some binaries that follow this convention of using `dsc` as a prefix without a separator, but it _looks weird_ and doesn't follow the general convention for related binaries in rust, which typically follows `<prefix>-<suffix>` for binaries and libraries both, though this is _just_ a convention.

I know I'd rather write `dsc-bicep` in the docs than `dscbicep` and if we're going to name it differently, better to do it before stable release than after- but definitely agree to defer to follow up PR.

What I would _really_ prefer is for it to be named `dsc-bicep-ext` to clarify that it's not remotely meant as a standalone application outside of either Bicep or DSC

dsc-bicep-ext/proto/bicep.proto
```diff
@@ -0,0 +1,70 @@
...
+  rpc Preview (ResourceSpecification) returns (LocalExtensibilityOperationResponse);
+  rpc Get (ResourceReference) returns (LocalExtensibilityOperationResponse);
+  rpc Delete (ResourceReference) returns (LocalExtensibilityOperationResponse);
+  rpc GetTypeFiles(Empty) returns (TypeFilesResponse);
+  rpc Ping(Empty) returns (Empty);
+}
+
+message Empty {}
+
+message ResourceSpecification {
```
michaeltlombardi, line 22: Nit: Would really like for these definitions to be documented so that they're correctly documented both here for future maintainers and in the library when working in VS Code.

Based on a quick look at the [docs for `prost_build`][01], it looks like added line comments above the type and field definitions will get pulled over as triple-slash docs comments in the generated rust.

For example, I can't currently remember what the `config` field is for - I know the others, but I'll have to dig through the code to see what `config` means in this context.

I'm _okay_ to defer the docs for this PR but I think sooner is better.

[01]: https://docs.rs/prost-build/latest/prost_build/

dsc-bicep-ext/src/main.rs
```diff
@@ -0,0 +1,423 @@
...
+            .map_err(|e| Status::aborted(e.to_string()))?
+        else {
+            return Err(Status::unimplemented("Group resources not supported"));
+        };
+
+        Ok(Response::new(LocalExtensibilityOperationResponse {
+            resource: Some(proto::Resource {
+                r#type: resource_type,
+                api_version: version,
+                identifiers: properties,
```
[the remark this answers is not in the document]
michaeltlombardi: We can't actually resolve this meaningfully until the DSC resource schema supports an extended vocabulary that recognizes a keyword to identify key properties, which we can then use to uniquely identify each instance of a given type.

dsc-bicep-ext/src/main.rs
michaeltlombardi: I think we can defer to future, but it would be nice to only conditionally include the HTTP stuff (and maybe the debugger? not sure whether that's normal for released implementations of extensions) behind a feature flag and default to it for debug but not release builds.

dsc-bicep-ext/src/main.rs
```diff
@@ -0,0 +1,508 @@
...
+            "{}",
+            t!(
+                "dscbicep.starting",
+                transport = "socket",
+                address = socket_path
+            )
+        );
+
+        // Remove the socket file if it exists
+        let _ = std::fs::remove_file(&socket_path);
```
michaeltlombardi, line 311: Can we just invoke this function without capturing the output into an unused variable?

```suggestion
        std::fs::remove_file(&socket_path);
```

dsc-bicep-ext/src/main.rs
```diff
@@ -0,0 +1,508 @@
...
+        // does work, but most of the problem lies in minimal Windows support
+        // inside the Tokio library (and no support for UDS).
+        use std::pin::Pin;
+        use std::task::{Context, Poll};
+        use tokio::io::{AsyncRead, AsyncWrite};
+        use tokio::net::windows::named_pipe::ServerOptions;
+        use tonic::transport::server::Connected;
+
+        // Wrapper to implement Connected trait for NamedPipeServer
+        struct NamedPipeConnection(tokio::net::windows::named_pipe::NamedPipeServer);
```
michaeltlombardi, line 336: Nit: It feels a little strange to define this newtype in an `if let` statement instead of a separate code block, but for this PR, I think it's okay - we should _probably_ decompose the monolithic `main.rs` to move definitions into library code for the crate. We could define something like

```rust
#[cfg(windows]
mod named_pipe_connection;
#[cfg(windows)]
pub use named_pipe_connection::NamedPipeConnection;
```

Where `named_pipe_connection.rs` then contains the definition and `impl` blocks, which would make this `if let` block functionally start with line 380 after the import statements.

I think there's a fair amount of decomposition that could be helpful but by no means required for this PR, just don't want to forget to mention it while reviewing.

dsc-bicep-ext/src/main.rs
michaeltlombardi: Assuming we don't need mutex, can we use [`tokio::sync::RwLock`](https://docs.rs/tokio/0.2.10/tokio/sync/struct.RwLock.html)? My (shaky) understanding is that as long as we're just _reading_ data from the instance we can read in parallel executions, and any execution that _needs_ to update the discovery cache will temporarily block other executions until it releases the write lock.

michaeltlombardi (commented):
Nothing blocking, except maybe that it would be good to have some minimal integration tests and documentation, but open to deferring both.

dsc-bicep-ext/src/main.rs
michaeltlombardi: I think we resolved the bidirectional invocation for delete/set depending on resource definition in #1317.

The code flows should be:

<details open><summary>Set operation</summary>

1. User invokes `set` for resource with `_exist` as `false`.
1. DSC checks whether the resource has the `setHandlesExist` capability:

   - If it has the capability, DSC invokes the `set` operation.
   - If it doesn't have the capability, DSC checks for whether the resource implements the `delete` operation.

     - If the resource implements `delete`, DSC invokes the `delete` operation and then synthesizes the `set` result.
     - If the resource doesn't implement `delete`, DSC raises an error for an unsupported operation.
1. DSC reports the result for the `set` operation to the user (or an error).

</details>

<details open><summary>Delete operation</summary>

1. User invokes `delete` for resource instance.
1. DSC checks whether the resource defines the `delete` operation:

   - If it does, DSC invokes the `delete` operation.
   - If it doesn't, DSC checks for whether the resource has the `setHandlesExist` capability:

     - If it has the capability, DSC invokes the `set` operation.
     - If it doesn't have the capability, DSC raises an error for an unsupported operation.
1. DSC reports the result for the `delete` operation to the user (or an error).

</details>

dsc-bicep-ext/proto/bicep.proto
andyleejordan: That would be documenting them upstream in Bicep, I think that's totally reasonable to do myself or ask @anthony-c-martin since some of these were certainly "debug and look at it" to really figure it out.

dsc-bicep-ext/src/main.rs
andyleejordan: I can give it a shot. Would it be: acquire write lock, execute `dsc.find_resource()`, release write lock, and then...well do we even need a read lock after that once we have the returned resource?

If `find_resource()` is the only thing we need to lock, like I mean if I can do that in a new scope and everything I need to do on the returned _found_ resource no longer requires the `dsc` instance, then we're good. But it's the last bit I don't know; what side effects exist inside `DscManager`?

dsc-bicep-ext/src/main.rs
andyleejordan: Okay great, that sounds to me like it is handled as expected at the `dsc-lib` layer. Curious, this implies to me that `delete()` came along in the design of DSCv3 after `set with _ensure=false` was implemented?

dsc-bicep-ext/src/main.rs
andyleejordan: Could certainly do that, but yes I'd do it in a short PR after this.

dsc-bicep-ext/src/main.rs
andyleejordan: Uhh yes you're right. Copilot helped with this 🫣

dsc-bicep-ext/Cargo.toml
andyleejordan: I can do that, it's easier to do now before setting up the release pipeline than later. `dsc-bicep-ext` works for me, I didn't like just `dsc-bicep` because to _me_ that implies there's a CLI of `dsc bicep <args>` which there is not. It's just a binary hidden in an OCI artifact used by Bicep.

dsc-bicep-ext/src/main.rs
andyleejordan: Heh or not, Clippy complains without it:

warning: unused `Result` that must be used

dsc-bicep-ext/src/main.rs
andyleejordan: Well, this is proving difficult as there are side effects of `find_resource` on the `resources` of the singleton `DscManager`. Once you've found one resource, that's it! Under the covers there's a `RESOURCES` global that has _all_ discovered resources cached, but that's not what's checked at the upper level.

andyleejordan: I think asking to use `DscManager` as a singleton is the opposite of the pattern its programmed around. An instance of it is really just an instance of `Discovery` (literally that's all that's in it) and if I understand the design of it (I ought to after debugging it for the last few hours) _that instance_ is meant to represent one discovered set of resources from the filter. Literally it takes `RESOURCES` from underneath which has everything, filters it, then saves it to `resources`. It's not meant to have `find_resources` called on it with different arguments. What's the opposite of idempotent?

I'll confirm that without using this as a singleton, the `RESOURCES` cache is correctly reused. I would hope so, but I guess experimenting is the only answer.

andyleejordan: Ok: I verified via debugging that the part which matters, `locked_extend!(RESOURCES, resources);` is called once and only once in the existing code, which means command discovery is correctly happening once per spawn of the process, not per invocation. This should _not_ be a singleton, it's not meant to be per my above explanation.

It's a shame because the `Arc::RwLock` was pretty neat (and took a bit to figure out). Live and learn: https://github.com/andyleejordan/DSC/commit/decc6bd975cbfd93d1d9c72dc897528dc447bc93

dsc-bicep-ext/proto/bicep.proto
anthony-c-martin: I've opened https://github.com/Azure/bicep/pull/18863 to pull in the comments from the source spec

dsc-bicep-ext/src/main.rs
andyleejordan: FWIW I tried to move this and the end result was a little messy. Doesn't seem worth doing, as it is a useful scenario, and is only engaged manually (starting up the `dsc-bicep-ext` binary manually). No harm I can think of compiling the code for release too, and in fact would be useful to users for debugging or simply exploring DSC.

dsc-bicep-ext/Cargo.toml
andyleejordan: It has been renamed and moved 😄

andyleejordan: (Caught myself lying: I hadn't pushed those two commits yet! Now it's done.)

PR #1332, opened 2025-12-20
Add `dataUri` and `dataUriToString` functions

Gijsreyn:
# PR Summary

This pull request adds the `dataUri` and `dataUriToString` functions, including tests and documentation.

## PR Context

Partially addresses #57.

### DIFF
--- a/lib/dsc-lib/src/functions/data_uri_to_string.rs
+++ b/lib/dsc-lib/src/functions/data_uri_to_string.rs
@@ -70,21 +70,19 @@ impl Function for DataUriToString {
 
         // Validate remaining parts: must be either "base64" or "charset=value"
         let mut has_base64 = false;
-        let mut charset_count = 0;
-        let mut charset_value = None;
+        let mut charset = None;
 
         for part in &parts[1..] {
             if *part == "base64" {
                 has_base64 = true;
             } else if part.starts_with("charset=") {
-                charset_count += 1;
-                if charset_count > 1 {
+                if charset.is_some() {
                     return Err(DscError::FunctionArg(
                         "dataUriToString".to_string(),
                         t!("functions.dataUriToString.invalidDataUri").to_string(),
                     ));
                 }
-                charset_value = Some(&part[8..]);
+                charset = Some(&part[8..]);
             } else {
                 return Err(DscError::FunctionArg(
                     "dataUriToString".to_string(),
@@ -99,7 +97,7 @@ impl Function for DataUriToString {
                 t!("functions.dataUriToString.notBase64").to_string(),
             ));
         }
-        if let Some(charset) = charset_value {
+        if let Some(charset) = charset {
             let charset_lower = charset.to_lowercase();
             if charset_lower != "utf-8" && charset_lower != "utf8" {
                 return Err(DscError::FunctionArg(


lib/dsc-lib/src/functions/data_uri.rs
```diff
@@ -0,0 +1,77 @@
...
+            accepted_arg_ordered_types: vec![vec![FunctionArgKind::String]],
+            remaining_arg_accepted_types: None,
+            return_types: vec![FunctionArgKind::String],
+        }
+    }
+
+    fn invoke(&self, args: &[Value], _context: &Context) -> Result<Value, DscError> {
+        let string_to_convert = args[0].as_str().unwrap_or_default();
+        let base64_encoded = general_purpose::STANDARD.encode(string_to_convert);
+        let result = format!("data:text/plain;charset=utf8;base64,{base64_encoded}");
```
SteveL-MSFT: Looking at the ARM function implementation, it will use `application/json` if the string is JSON.

lib/dsc-lib/src/functions/data_uri_to_string.rs
```diff
@@ -0,0 +1,161 @@
...
+        if metadata.contains(";base64") {
+            // Decode base64
+            let decoded_bytes = general_purpose::STANDARD.decode(encoded_data).map_err(|_| {
+                DscError::FunctionArg(
+                    "dataUriToString".to_string(),
+                    t!("functions.dataUriToString.invalidBase64").to_string(),
+                )
+            })?;
+
+            let result = String::from_utf8(decoded_bytes).map_err(|_| {
```
SteveL-MSFT: I don't think we can assume utf8 since the charset is part of the URI

lib/dsc-lib/src/functions/data_uri_to_string.rs
```diff
@@ -0,0 +1,161 @@
...
+                DscError::FunctionArg(
+                    "dataUriToString".to_string(),
+                    t!("functions.dataUriToString.invalidUtf8").to_string(),
+                )
+            })?;
+
+            Ok(Value::String(result))
+        } else {
+            // TODO: Not sure if this is also done in ARM
```
SteveL-MSFT: ARM implementation just fails if `base64` is not declared

lib/dsc-lib/src/functions/data_uri_to_string.rs
Gijsreyn: What should the function check? If a charset is specified in the URI and error if it's not UTF-8? Do we need to support other charsets?

SteveL-MSFT: I think if not specified, we can assume utf-8 since it's prevalent, but if the value is something else (like utf-16), we can just error saying it's not supported for now.  We can add more charset support in the future based on user feedback.

dsc/tests/dsc_functions.tests.ps1
```diff
@@ -1714,4 +1714,68 @@ Describe 'tests for function expressions' {
...
+"@
+    $out = $config_yaml | dsc config get -f - | ConvertFrom-Json
+    $out.results[0].result.actualState.output | Should -Be $expected
+  }
+
+  It 'dataUriToString function error handling: <expression>' -TestCases @(
+    @{ expression = "[dataUriToString('not a data uri')]" ; expectedError = 'Invalid data URI format' }
+    @{ expression = "[dataUriToString('data:text/plain;base64')]" ; expectedError = 'Invalid data URI format' }
+    @{ expression = "[dataUriToString('data:;base64,invalid!@#')]" ; expectedError = 'Invalid base64 encoding' }
+    @{ expression = "[dataUriToString('data:text/plain;charset=utf-16;base64,SGVsbG8=')]" ; expectedError = 'Unsupported charset' }
```
SteveL-MSFT, line 1764: Can you add a few more tests:

double-colon: data::text/plain;base64,
double-semi-colon: data:text/plain;;base64,
double-char-set: data:text/plain;charset=utf-8;charset=utf-8,base64,
double-comma: data:text/plain;base64,,
invalid-property: data:text/plain;foo=bar;base64,

SteveL-MSFT (changes requested):

dsc/tests/dsc_functions.tests.ps1
```diff
@@ -1714,4 +1714,90 @@ Describe 'tests for function expressions' {
...
+    $LASTEXITCODE | Should -Not -Be 0
+    $errorContent = Get-Content $TestDrive/error.log -Raw
+    $errorContent | Should -Match $expectedError
+  }
+
+  It 'dataUriToString function handles edge cases: <expression>' -TestCases @(
+    @{ expression = "[dataUriToString('data:text/plain;;base64,SGVsbG8=')]" ; expected = 'Hello' }
+    @{ expression = "[dataUriToString('data::text/plain;base64,SGVsbG8=')]" ; expected = 'Hello' }
+    @{ expression = "[dataUriToString('data:text/plain;charset=utf-8;charset=utf-8;base64,SGVsbG8=')]" ; expected = 'Hello' }
+    @{ expression = "[dataUriToString('data:text/plain;foo=bar;base64,SGVsbG8=')]" ; expected = 'Hello' }
```
SteveL-MSFT: The ARM implementation seems more strict in the format validation and I believe these would all be errors

SteveL-MSFT (changes requested):

docs/reference/schemas/config/functions/dataUri.md
```diff
@@ -0,0 +1,191 @@
...
+`data:text/plain;charset=utf8;base64,<encoded-content>` where `<encoded-content>` is the base64
+representation of the **stringToConvert** value.
+
+```yaml
+Type: string
+```
+
+<!-- Link reference definitions -->
+[01]: https://en.wikipedia.org/wiki/Data_URI_scheme
+[02]: base64.md
```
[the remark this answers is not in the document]
Gijsreyn: Not referencing anything here

lib/dsc-lib/src/functions/data_uri_to_string.rs
```diff
@@ -52,23 +52,59 @@ impl Function for DataUriToString {
...
+        // Validate remaining parts: must be either "base64" or "charset=value"
+        let mut has_base64 = false;
+        let mut charset_count = 0;
+        let mut charset_value = None;
+
+        for part in &parts[1..] {
+            if *part == "base64" {
+                has_base64 = true;
+            } else if part.starts_with("charset=") {
+                charset_count += 1;
```
SteveL-MSFT: Instead of having `charset_count`, you should just check if `charset.is_some()` (after renaming `charset_value` to just `charset` since it'd be the only variable)

SteveL-MSFT (changes requested):

PR #1333, opened 2025-12-21
Add `--what-if` option to resource set command for previewing changes

### LINKED ISSUE #566: Support `--what-if` mode for single-resource invocation

> As an infrastructure engineer, I want to be able to invoke a resource's set operation with the `--what-if` flag to see how the resource would change the system, so that I can use one-off resource invocations more reliably for interactive and exploratory operations.

> As a resource developer, I want to test the behavior of my resource in what-if mode for DSC with minimal overhead and unrelated operations, so that I can progressively enhance my test suite quickly and cleanly.

Right now, invoking resources in what-if mode requires creating a configuration document. While having what-if mode is _more_ valuable for configuration documents, it's also useful for singular resource invocations like:

```powershell
$properties | dsc resource set -r myOrg/myResource --what-if
```

This would be useful both for interactive/script users and for resource authors, who can then test the what-if messaging and behavior for their resources with less overhead and a simpler output data model.

I propose that we add the `w`/`--what-if` flag to the `dsc resource set` command, reusing (as much as possible) the logic and code for invoking resources in what-if mode used by the `dsc config set` command.

Gijsreyn:
# PR Summary

This PR adds support for `--what-if` mode when invoking a resource's `set` operation directly via the CLI.

## PR Context

Fix #566.

### DIFF
--- a/dsc/src/resource_command.rs
+++ b/dsc/src/resource_command.rs
@@ -182,15 +182,6 @@ pub fn set(dsc: &mut DscManager, resource_type: &str, version: Option<&str>, inp
                 error!("{err}");
                 exit(EXIT_DSC_ERROR);
             }
-
-            match resource.get(input) {
-                Ok(GetResult::Resource(response)) => response.actual_state,
-                Ok(_) => unreachable!(),
-                Err(err) => {
-                    error!("{err}");
-                    exit(EXIT_DSC_ERROR);
-                }
-            }
         };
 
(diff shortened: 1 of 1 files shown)


tgauth:
@Gijsreyn - if you feel like resolving the merge conflicts on this, I will review! 
I'm planning to do the same changes for the resource delete command soon

Gijsreyn:
@tgauth - hey Tess, should be good to go with the merge conflicts (there are none) :)

PR #1348, opened 2026-01-12
Add sshdconfig _purge=false support for regular keywords

tgauth:
# PR Summary
- renames `clobber` to `purge` in various locations
- adds sshdconfig `set` support with `_purge: false` for regular keywords
- refactors some of the existing code in `set_sshd_config()` by creating `write_config_map_to_text()`, a helper function in util.rs
- refactors writing the config file into 2 distinct parts: global config then match block(s), in order to simplify appending any added keyword(s) to the global config without needing to worry about insertion/order preservation
- adds Pester tests for set scenarios, including if sshd_config file does not exist - the file creation will be implicit for `_purge: true` but will result in an error for `_purge: false` since there would be no existing settings to preserve

### DIFF
--- a/resources/sshdconfig/src/set.rs
+++ b/resources/sshdconfig/src/set.rs
@@ -133,6 +133,8 @@ fn set_sshd_config(cmd_info: &mut CommandInfo) -> Result<(), SshdConfigError> {
         for (key, value) in &cmd_info.input {
             let key_contains = key.as_str();
 
+            // TODO: remove when design for handling repeatable and multi-arg keywords is finalized
+            // and consider using SshdConfigValue instead of any remaining contains() checks
             if REPEATABLE_KEYWORDS.contains(&key_contains)
                 || MULTI_ARG_KEYWORDS_COMMA_SEP.contains(&key_contains)
                 || MULTI_ARG_KEYWORDS_SPACE_SEP.contains(&key_contains) {


resources/sshdconfig/src/set.rs
```diff
@@ -106,37 +107,49 @@ fn remove_registry(name: &str) -> Result<(), SshdConfigError> {
...
     // i.e. if the key can be repeated or have multiple values, etc.
     // or if the value is something besides a string (like an object to convert back into a comma-separated list)
     debug!("{}", t!("set.writingTempConfig"));
     let mut config_text = SSHD_CONFIG_HEADER.to_string() + "\n" + SSHD_CONFIG_HEADER_VERSION + "\n" + SSHD_CONFIG_HEADER_WARNING + "\n";
-    if cmd_info.clobber {
+    if cmd_info.purge {
+        let match_map = cmd_info.input.remove("match");
+        config_text.push_str(&write_config_map_to_text(&cmd_info.input, match_map)?);
+    } else {
+        let mut get_cmd_info = cmd_info.clone();
```
SteveL-MSFT, line 120: cmd_info doesn't appear to be used after this if..else section, so why clone it?  It's already mut

resources/sshdconfig/src/set.rs
```diff
@@ -106,37 +107,49 @@ fn remove_registry(name: &str) -> Result<(), SshdConfigError> {
...
+            Ok(config) => config,
+            Err(SshdConfigError::FileNotFound(_)) => {
+                return Err(SshdConfigError::InvalidInput(
+                    t!("set.purgeFalseRequiresExistingFile").to_string()
+                ));
+            }
+            Err(e) => return Err(e),
+        };
         for (key, value) in &cmd_info.input {
             let key_lower = key.to_lowercase();
```
SteveL-MSFT: I know this wasn't changed in this PR, but it would be better to do any `to_lowercase()` during `CommandInfo::new()` so that subsequent code that relies on it can just expect it to be lowercase rather than missing a code path that didn't convert it.

resources/sshdconfig/src/set.rs
```diff
@@ -106,37 +107,49 @@ fn remove_registry(name: &str) -> Result<(), SshdConfigError> {
...
-fn set_sshd_config(cmd_info: &CommandInfo) -> Result<(), SshdConfigError> {
+fn set_sshd_config(cmd_info: &mut CommandInfo) -> Result<(), SshdConfigError> {
     // this should be its own helper function that checks that the value makes sense for the key type
     // i.e. if the key can be repeated or have multiple values, etc.
     // or if the value is something besides a string (like an object to convert back into a comma-separated list)
     debug!("{}", t!("set.writingTempConfig"));
     let mut config_text = SSHD_CONFIG_HEADER.to_string() + "\n" + SSHD_CONFIG_HEADER_VERSION + "\n" + SSHD_CONFIG_HEADER_WARNING + "\n";
-    if cmd_info.clobber {
+    if cmd_info.purge {
+        let match_map = cmd_info.input.remove("match");
```
SteveL-MSFT: Not clear to me why here and line 151 removes the `match` field instead of just doing a get.  It doesn't appear to impact use of cmd_info later?

resources/sshdconfig/src/util.rs
```diff
@@ -114,38 +114,33 @@ pub fn enable_tracing(trace_level: Option<&TraceLevel>, trace_format: &TraceForm
...
-
+    let result = format_value_as_string(value, MULTI_ARG_KEYWORDS_COMMA_SEP.contains(&key_lower.as_str()));
     if result.is_empty() {
         return Err(SshdConfigError::ParserError(t!("util.invalidValue", key = key).to_string()))
     }
     Ok(result)
 }
 
-fn format_value_as_string(value: &Value, is_comma_separated: bool) -> Result<String, SshdConfigError> {
+fn format_value_as_string(value: &Value, is_comma_separated: bool) -> String {
```
SteveL-MSFT: It would be better to have a custom struct that encapsulates `Value` and also whether it's comma separated or potentially other properties

resources/sshdconfig/src/util.rs
```diff
@@ -379,3 +376,48 @@ fn format_match_block(match_obj: &Value) -> Result<String, SshdConfigError> {
...
+    let mut config_text = String::new();
+
+    for (key, value) in global_map {
+        let key_lower = key.to_lowercase();
+
+        // Handle repeatable keywords - write multiple lines
+        if REPEATABLE_KEYWORDS.contains(&key_lower.as_str()) {
+            if let Value::Array(arr) = value {
+                for item in arr {
+                    let formatted = format_sshd_value(key, item)?;
```
SteveL-MSFT: If you had a struct for the keywords instead of using Value, you could just implement `Display` and not have to explicitly call the helper to format

SteveL-MSFT (changes requested):

resources/sshdconfig/src/set.rs
tgauth: cmd_info is used again on line 183

resources/sshdconfig/src/set.rs
tgauth: looks like this was redundant because any input keys were already being lowercased in `build_command_info()` but regardless, agree that should move into `CommandInfo::new()`

resources/sshdconfig/src/set.rs
tgauth: Good point. My goal here was to ensure that any match blocks are written to the config file after all the global values, but probably makes more sense to handle inside of `write_config_map_to_text()` directly

resources/sshdconfig/src/inputs.rs
```diff
@@ -7,28 +7,39 @@ use std::path::PathBuf;
...
+        include_defaults: bool,
+        input: Map<String, Value>,
+        metadata: Metadata,
+        purge: bool,
+        sshd_args: Option<SshdCommandArgs>
+    ) -> Self {
+        // Lowercase keys for case-insensitive comparison
+        let input = input.into_iter()
+            .map(|(k, v)| (k.to_lowercase(), v))
+            .collect();
```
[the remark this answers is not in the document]
tgauth: This is valid but I think we can defer preserving original key casing for a future PR if/when that feature is explicitly requested

resources/sshdconfig/src/set.rs
```diff
@@ -106,37 +108,46 @@ fn remove_registry(name: &str) -> Result<(), SshdConfigError> {
...
-                        writeln!(&mut config_text, "{key} {formatted}")?;
-                    }
-                } else {
-                    // Single value for repeatable keyword, write as-is
-                    let formatted = format_sshd_value(key, value)?;
-                    writeln!(&mut config_text, "{key} {formatted}")?;
-                }
+            let key_contains = key.as_str();
+
+            if REPEATABLE_KEYWORDS.contains(&key_contains)
```
SteveL-MSFT, line 138: Is there a way to use SshdConfigValue here to check those properties instead of the `contains()` checks?

tgauth: Probably, but I'm not sure it's necessary since this code is more of a placeholder until the design for repeatable/multi-arg keywords when `_purge: false` is decided

tgauth: For now, I'll add a comment about that

Issue #1221, opened 2025-10-29
Replace `deny_unknown_fields` with warn and ignore logic


andyleejordan:
Just a heads up that as noted in the PR, Bicep has and will likely continue to add new fields (especially when using experimental features) and since we're setting `#[serde(deny_unknown_fields)]` we're allowing that to **break** us rather than warn us. I highly recommend we switch from that to an ignore and warn approach instead. /cc @michaeltlombardi

_Originally posted by @andyleejordan in https://github.com/PowerShell/DSC/issues/1210#issuecomment-3459127353_

michaeltlombardi:
We can remove the `additionalProperties` and/or `unevaluatedProperties` keywords fairly easily. It's possibly worth considering an _ARM_ form for the configuration document, where we remove those keywords.

My main concern is that if we relax the schema validation, it becomes harder to detect whether you have mistyped a property name or are using an invalid construct. We should definitely loosen the underlying requirements on the `struct` definition but I think we should probably keep the keyword for validating general configuration documents.

andyleejordan:
Resolved by eliminating requirement to parse Bicep-generated ARM JSON files.

Issue #1223, opened 2025-10-29
Fix `dependsOn` for symbolic names


andyleejordan:
Looking at the ARM docs https://learn.microsoft.com/en-us/azure/azure-resource-manager/templates/resource-declaration#use-symbolic-name, I don't think you should just clobber the name with the symbolic name.  Instead, looking at https://learn.microsoft.com/en-us/azure/azure-resource-manager/templates/resource-dependency, it appears that symbolic names can be used in `dependsOn` in addition to `resourceId()` in the classical sense.  So I think what we may need to eventually (this can be a follow-up PR) do is maintain a list of symbolic names and for resolving dependencies allow for either symbolic name or resourceid().

For this PR, I would just change the `TODO` comment to reflect what I said above.

_Originally posted by @SteveL-MSFT in https://github.com/PowerShell/DSC/pull/1210#discussion_r2466875002_

SteveL-MSFT:
Yes, my understanding with 2.0 is that `dependsOn()` can take a symbolic name or `resourceId()`.

andyleejordan:
Resolved by eliminating requirement to parse Bicep-generated ARM JSON files.

PR #1353, opened 2026-01-14
Revert changes required for ARMv2 experiment

andyleejordan:
# PR Summary

Since we no longer expect DSC to ingest ARMv2 as generated by Bicep (in favor of Bicep invoking DSC resources via the gRPC extension), we can remove this unfinished work that simply got our demos running again.

## PR Context

This deserialization approach was a cheap workaround, we had not yet revisited the design to properly support symbolic names (such as fixing resource IDs, dependency resolution, etc.).

We also don't need to add the Bicep emitted fields that we were just skipping.


andyleejordan:
This is a revert of #1210, but with merge conflicts resolved, and still-relevant tests kept. We can also close #1221 and #1223. I would strongly encourage we remove the `copy()` (loop) implementation and rely on higher-order tooling for complex resource orchestration.

Gijsreyn:
@andyleejordan - isn't the `copy()` functionality a v1 feature?

andyleejordan:
> @andyleejordan - isn't the `copy()` functionality a v1 feature?

🤷 as far as I could tell it was a lot of hassle for little if any gain as it was really to support `for` loops that came from Bicep. We don't necessarily need to do that in DSC-native configs, but it's just a suggestion from me and not relevant to this PR (except that I removed the part of the `copy()` test dealing with symbolic names).

Gijsreyn:
> > @andyleejordan - isn't the `copy()` functionality a v1 feature?
> 
> 🤷 as far as I could tell it was a lot of hassle for little if any gain as it was really to support `for` loops that came from Bicep. We don't necessarily need to do that in DSC-native configs, but it's just a suggestion from me and not relevant to this PR (except that I removed the part of the `copy()` test dealing with symbolic names).

Thanks for the explanation, Andy, and I get that part indeed. Symbolic names are more towards v2 (or even are). There might be other ways to loop for sure, but I can't think of any atm.

michaeltlombardi (approved):
Code changes look good to me - I think there's still some open questions on impact to other functionality (copy loops), but I prefer that functionality be pushed to higher order tools (bicep, a DSC config builder in PowerShell, whatever) instead of being inherent to the engine / configuration document.

andyleejordan:
@SteveL-MSFT this is ready to merge!

SteveL-MSFT (approved):

PR #1360, opened 2026-01-21
Remove unused async-stream package

andyleejordan:
It was briefly used when attempting to wire up the named pipes on Windows, but ended up not being used and I forgot to delete it.


andyleejordan:
>  let incoming = async_stream::stream! {

Nevermind, it _is_ in use, just only on Windows so it was hard to spot.

PR #1363, opened 2026-01-23
Remove bicep extension for DSC

SteveL-MSFT:
# PR Summary

Since we formally can no longer completely support JSON generated from Bicep, we need to remove the Bicep extension for DSC


kilasuit:
@SteveL-MSFT / @tgauth  how come this has happened?

Issue #1170, opened 2025-10-09
Have metadata in config to have dsc engine ignore missing resources


SteveL-MSFT:
### Summary of the new feature / enhancement

Some resources may be installed at runtime via a package manager resource.  Currently, DSC engine before spawning resources will attempt to discover all used resources.  If any one is missing, the config operation fails before any resources are called.

### Proposed technical implementation details (optional)

Have new metadata to declare to DSC engine to ignore missing resources:

```yaml
metadata:
  Microsoft.DSC:
    missingResources: ignore
resources:
- name: Install Foo
  type: Microsoft.WinGet/Package
  properties:
    id: foo
- name: Use Foo
  type: Foo/bar
  properties:
    hello: world
```

`missingResources` would accept `ignore`, `warn`, `error` where if not specified defaults to `error`.  `warn` emits a warning message, but proceeds.

denelon:
This is a great solution for WinGet as a "high-order" tool to still allow a configuration to be executed even if WinGet and DSC.exe can't validate the resources that will end up getting installed as some subset of the configuration document.

SteveL-MSFT:
Thinking about this a bit more, there's actually a bit more complication to this.

What happens now is that when given a config, the DSC engine will first validate it is valid in terms of resources.  So it does a complete discovery of all resources and determines if the ones used are in that list.  If any are missing, an error is returned and deployment stops.

In this scenario, a resource may install other resources used in the config.  So `ignore` won't work because discovery only happens at the start.  So instead, I think what we need is:

```yaml
metadata:
  Microsoft.DSC:
    resourceDiscovery: [ preDeployment | duringDeployment ]
```

`preDeployment` is current behavior and the default if not specified.  It attempts during validation stage to find all resources and reconcile them with the ones specified in the config.  Any missing results in no deployment attempt.

`duringDeployment` will defer discovery until the first resource is used and re-discovery has to happen for each new resource used.

This metadata is specified at the config doc level only.

SteveL-MSFT:
Thinking about this further, I think I might prefer having resources like `WinGet/Package` declare in their manifest a capability like resource installation.  Then, any resources that `dependsOn` that resource won't be flagged as missing and after execution of `WinGet/Package`, it would force rediscovery of resources.  However, this does require the config author to denote potentially not installed resources as potentially being installed at deployment time.  This makes it more specific than the global metadata above which would skip over resources that aren't there and won't be installed.

ThomasNieto:
[...]

PR #1366, opened 2026-01-27
Enable re-discovery of resources during deployment

SteveL-MSFT:
# PR Summary

To support the scenario where the config installs resources that will get used later, this change introduces new metadata under the `Microsoft.DSC` namespace to enable re-discovery of resources during deployment:

```yaml
$schema: https://aka.ms/dsc/schemas/v3/bundled/config/manifest.json
  metadata:
    Microsoft.DSC:
      resourceDiscovery: duringDeployment
      resources:
        ...
```

The new `resourceDiscovery` property accepts `preDeployment` (the default) or `duringDeployment`.  `preDeployment` means that all resources (except within a group) are discovered and validated before deployment.  So if any resource is missing, no resources are executed (this PR fixes that regression).

For testing, this PR creates a new test resource where you can specify a manifest file to copy and rename the type name to simulate the installation of a new resource (the new resource still needs to adhere to current discovery rules).

One side effect here is that to support dynamically installed resources, before EVERY resource invocation, a discovery is performed.  We can optimize this in the future later based on other ideas in the issue below.

This change required storing the mode in Discovery and also CommandDiscovery to force rediscovery since both previously optimized where if it had results cached, it skipped discovery every time.

Also fixed an issue in the `Install-Protobuf` helper function as `winget` returns a non-zero exit code if the package is already installed.  Fix is to use `--force` to force an install.

## PR Context

Fix https://github.com/PowerShell/DSC/issues/1170

### DIFF
--- a/build.helpers.psm1
+++ b/build.helpers.psm1
@@ -652,6 +652,13 @@ function Install-Protobuf {
             if (Test-CommandAvailable -Name 'winget') {
                 Write-Verbose -Verbose "Using winget to install Protobuf"
                 winget install Google.Protobuf --accept-source-agreements --accept-package-agreements --source winget --force
+                # need to add to PATH
+                $protocFolder = "$env:USERPROFILE\AppData\Local\Microsoft\WinGet\Packages\Google.Protobuf_Microsoft.Winget.Source_8wekyb3d8bbwe\bin"
+                if (Test-Path $protocFolder) {
+                    $env:PATH += ";$protocFolder"
+                } else {
+                    throw "protoc folder not found after installation: $protocFolder"
+                }
             } else {
                 Write-Warning "winget not found, please install Protobuf manually"
             }


lib/dsc-lib/src/discovery/mod.rs
```diff
@@ -133,12 +136,15 @@ impl Discovery {
...
             // If resources are already discovered, no need to re-discover.
             return Ok(());
         }
 
-        let command_discovery = CommandDiscovery::new(progress_format);
+        let mut command_discovery = CommandDiscovery::new(progress_format);
+        if self.refresh_cache {
+            command_discovery.set_discovery_mode(&ResourceDiscoveryMode::DuringDeployment);
+        }
         let discovery_types: Vec<Box<dyn ResourceDiscovery>> = vec![
```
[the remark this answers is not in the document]
SteveL-MSFT: Will fix

dsc/tests/dsc_discovery.tests.ps1
```diff
@@ -296,4 +296,47 @@ Describe 'tests for resource discovery' {
...
+            $traceLog | Should -Match "ERROR.*?Resource not found: Test/$guid"
+            $traceLog | Should -Not -Match "Invoking get for 'Test/CopyResource'"
+        } else {
+            $LASTEXITCODE | Should -Be 0 -Because (Get-Content -Raw -Path "$testdrive/tracing.txt")
+            $output = $out | ConvertFrom-Json
+            $output.results[0].result.actualState.typeName | Should -BeExactly "Test/$guid" -Because $out
+            $output.results[1].result.actualState.output | Should -BeExactly 'Hello World' -Because $out
+            $traceLog | Should -Match "Invoking get for 'Test/$guid'"
+            $traceLog | Should -Match "Skipping resource discovery due to 'resourceDiscovery' mode set to 'DuringDeployment'"
+        }
```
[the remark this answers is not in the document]
SteveL-MSFT: Will fix

lib/dsc-lib/src/discovery/mod.rs
```diff
@@ -133,12 +136,18 @@ impl Discovery {
...
+        } else if self.refresh_cache {
+            self.resources.clear();
+            self.extensions.clear();
         }
 
-        let command_discovery = CommandDiscovery::new(progress_format);
+        let mut command_discovery = CommandDiscovery::new(progress_format);
+        if self.refresh_cache {
+            command_discovery.set_discovery_mode(&ResourceDiscoveryMode::DuringDeployment);
+        }
```
tgauth:
```suggestion
        let mut command_discovery = CommandDiscovery::new(progress_format);
        if self.refresh_cache {
            self.resources.clear();
            self.extensions.clear();
            command_discovery.set_discovery_mode(&ResourceDiscoveryMode::DuringDeployment);
        }
```

dsc/tests/dsc_discovery.tests.ps1
```diff
@@ -69,9 +79,18 @@ Describe 'tests for resource discovery' {
...
+            $env:DSC_RESOURCE_PATH = $testdrive
+
+            Set-Content -Path "$testdrive/test.dsc.resource.txt" -Value $resourceInput
+            $resources = dsc resource list | ConvertFrom-Json
+            $resources.Count | Should -Be 0
+        }
+        finally {
+            $env:PATH = $oldPath
+            $env:DSC_RESOURCE_PATH = $null
+        }
```
tgauth: What's the difference between the try/finally block in each of these 3 tests vs. adding a `Context` block with one `AfterEach`?

SteveL-MSFT: Actually having a Context for these 3 would be better, I'll make that change

Issue #142, opened 2023-08-04
DSC Resource manifest schema should allow for setting properties as "keys"


JohnMcPMS:
### Summary of the new feature / enhancement

While we can use the schema to define the `properties` and their individual schemas, there is no way to specify that a particular property is a "key".  This is distinct from required/optional.  The set of all "key" property values given to a resource defines the identity of the data under management.  Thus, passing the same set of key values, while varying other values, will result in a potential overwrite of the managed data.

The goal is to be able take two sets of properties that are intended for the same resource and determine if they are in conflict.  As an example:
```YAML
<config1 contains>
- name: A
  type: my/packages
  properties:
    id: foo
    version: [1.0-2.0)

<config2 contains>
- name: B
  type: my/packages
  properties:
    id: foo
    version: [3.0-]
```

With the resource `my/packages` having a property key set containing `id`, we can determine that config1:A and config2:B are potentially in conflict (they have matching key values and differing non-key values).  This can be reported to the user so that they may determine how to proceed.

michaeltlombardi:
This should be part of the JSON Schema that describes a resources' instances. We could define it as part of a JSON Schema dialect as an annotation keyword, like `keyProperty`:

```yaml
type: my/package
schema:
  embedded:
    type: object
    properties:
      id:
        title:       Package ID
        description: The unique ID for the package.
        type:        string
        keyProperty: true
      version:
        title:       Package Version
        description: The exact version or version range for the package.
        type:        string
```

Annotation keywords are (relatively) easy to add to a JSON Schema dialect, and can be safely ignored by tools that don't _need_ to care about them (that is, they don't change the processing of the instance itself, like the `properties` applicator keyword, they provide information _about_ the instance).

I've been thinking about the dialect problem for a while and I think it more-likely-than-not that we'll eventually need a dialect for DSC Resource instance schemas - you need dialects for authoring meta schemas if you want to support authoring/processing for those schemas beyond the built-in keywords.

<details>
<summary>Aside on the value of a schema dialect for instance schemas</summary>

A schema dialect can also really improve the authoring experience by providing shorthand for otherwise complex requirements.

For example, instead of defining an ensurable resource that uses the well-known `_ensure` property like this:

```yaml
schema:
  embedded:
    type: object
    properties:
      _ensure:
        $ref: https://raw.githubusercontent.com/PowerShell/DSC/main/schemas/2023/08/resource/properties/ensure.json
```

You could do this:

```yaml
schema:
  embedded:
[...]

PR #1367, opened 2026-01-27
Add Subsystem and SubsystemList Resources

### LINKED ISSUE #1340: Proposal: SSHDConfig Set Support for Repeatable Keywords

The `Microsoft.OpenSSH/SSHDConfig` resource currently supports setting sshd_config keywords only when `_purge: true`, which replaces the entire configuration file. When `_purge: false`, there is ambiguity in handling keywords that can have multiple arguments. Keywords that can have multiple arguments are further divided into three categories: repeatable, comma-separated, and space-separated. Repeatable keywords (like `Subsystem` and `Port`) are spread over multiple lines. Comma-separated (like `Ciphers`) and space-separated (like `AuthorizedKeysFile`) keywords can have multiple arguments on a single line. Note, the keywords that are both space-separated and repeatable (i.e. `AllowUsers`) are classified under "repeatable" for the scope of this discussion.

This proposal builds on the ideas previously discussed in #790 and #1070 to suggest handling these keywords through the following:
1. For all keywords: the `Microsoft.OpenSSH/SSHDConfig` resource simply passes the declared value through directly (overwriting any existing value(s) that may exist for that keyword)
2. For repeatable keywords: the `Microsoft.OpenSSH/SSHDConfig.Repeatable` resource with `_exist` facilitates addition/removal of specific entries 

Examples:
##### Set - behaves as overwrite for all keyword types
```yaml
- type: Microsoft.OpenSSH/SSHDConfig
  properties:
    PasswordAuthentication: false
    Ciphers:
     - +aes256-gcm@openssh.com
    Port:
     - 2222
```
##### Set - behaves as "add" for repeatable keywords
```yaml
- type: Microsoft.OpenSSH/SSHDConfig.Repeatable
  properties:
    _exist: true
    subsystem:
      - "powershell /usr/bin/pwsh -sshs -NoLogo"
    port:
      - 3333
```
##### Set - behaves as "remove" for repeatable keywords
```yaml
- type: Microsoft.OpenSSH/SSHDConfig.Repeatable
  properties:
    _exist: false
    subsystem:
      - "powershell /usr/bin/pwsh -sshs -NoLogo"
    port:
      - 3333
```

tgauth:
# PR Summary
- modify `sshdconfig` set behavior to overwrite for all keywords, instead of erroring for repeatable/multi-arg keywords
- add resources for `subsystem` and `subsystemList`, handled internally by sshdconfig via `repeat_keyword.rs` module
- add `canonical_properties.rs` module to centralize handling of non-ssh keywords (`_metadata`, `_exist`, etc.) 

## PR Context
- fix #1340 [partially]


ThomasNieto:
With two resources that manage the same underlying configuration. When key properties are introduced how would that work? I added a comment on this issue: https://github.com/PowerShell/DSC/issues/142#issuecomment-3787881058

We might want to discuss in the WG group what the plan is for that.

SteveL-MSFT:
> With two resources that manage the same underlying configuration. When key properties are introduced how would that work? I added a comment on this issue: [#142 (comment)](https://github.com/PowerShell/DSC/issues/142#issuecomment-3787881058)
> 
> We might want to discuss in the WG group what the plan is for that.

Tagged for WG discussion.  I don't think we want to make the two mutually exclusive in that you can only use one or the other.  However, currently this would naturally be resolved with last-one-wins.  Since the resource is instantiated separately for each type, it doesn't know that it was already used before.  Is documentation sufficient here?

theJasonHelmick:
The DSC WG discussed and we think this behavior should be documented. We don't believe we want to enforce or detect conflict across resources.

dsc/tests/dsc_sshdconfig.tests.ps1
```diff
@@ -177,4 +177,197 @@ resources:
...
+    Context 'Subsystem and SubsystemList Tests' {
+        BeforeAll {
+            # Create a temporary test directory for sshd_config files
+            $TestDir = Join-Path $TestDrive "sshd_test"
+            New-Item -Path $TestDir -ItemType Directory -Force | Out-Null
+            $script:TestConfigPath = Join-Path $TestDir "sshd_config"
+
+            # Define OS-specific paths with spaces
+            if ($IsWindows) {
+                $script:PathWithSpaces = "C:\Program Files\OpenSSH\sftp-server.exe"
```
SteveL-MSFT: Don't assume C: drive, use `$env:SystemDrive` or `$env:ProgramFiles` instead

dsc/tests/dsc_sshdconfig.tests.ps1
```diff
@@ -177,4 +177,197 @@ resources:
...
+Subsystem test2 /path/to/test2
+PasswordAuthentication yes
+"@
+            Set-Content -Path $script:TestConfigPath -Value $initialContent
+        }
+
+        AfterEach {
+            # Clean up test config file after each test
+            if (Test-Path $script:TestConfigPath) {
+                Remove-Item -Path $script:TestConfigPath -Force -ErrorAction SilentlyContinue
```
SteveL-MSFT: In general, use `-ErrorAction Ignore` instead of `-ErrorAction SilentlyContinue`, the difference is that `SilentlyContinue` still populates the ErrorRecord in `$Error` so if something expects that to be empty, it won't be

dsc/tests/dsc_sshdconfig.tests.ps1
```diff
@@ -177,4 +177,197 @@ resources:
...
+            $config_yaml = @"
+`$schema: https://aka.ms/dsc/schemas/v3/bundled/config/document.json
+metadata:
+  Microsoft.DSC:
+    securityContext: elevated
+resources:
+- name: newsub
+  type: Microsoft.OpenSSH.SSHD/Subsystem
+  metadata:
+    filepath: $($script:TestConfigPath -replace '\\', '/')
```
SteveL-MSFT: `TestConfigPath` should have the appropriate slashes for the OS.  Does the resource always expect forward slash?  Seems wrong for Windows.  I think the resource should expect the correct slashes for the OS.

dsc/tests/dsc_sshdconfig.tests.ps1
```diff
@@ -177,4 +177,197 @@ resources:
...
+            $out.hadErrors | Should -BeFalse
+            $out.results.Count | Should -Be 1
+            $out.results[0].type | Should -BeExactly 'Microsoft.OpenSSH.SSHD/Subsystem'
+            $out.results[0].result.afterState._exist | Should -Be $true
+            $out.results[0].result.afterState.subsystem.name | Should -Be 'newsubsystem'
+
+            # Verify file contains the new subsystem
+            $subsystems = Get-Content $script:TestConfigPath | Where-Object { $_ -match '^\s*subsystem\s+' }
+            $subsystems.Count | Should -Be 3
+            $subsystems | Should -Contain "subsystem newsubsystem /path/to/newsubsystem"
```
SteveL-MSFT: Can you use the resource itself via a `get` to validate this?

dsc/tests/dsc_sshdconfig.tests.ps1
```diff
@@ -177,4 +177,197 @@ resources:
...
+            $LASTEXITCODE | Should -Be 0
+            $out.hadErrors | Should -BeFalse
+            $out.results[0].type | Should -BeExactly 'Microsoft.OpenSSH.SSHD/SubsystemList'
+            $out.results[0].result.afterState.subsystem.Count | Should -Be 4
+
+            # Verify all subsystems are present (old + new)
+            $subsystems = Get-Content $script:TestConfigPath | Where-Object { $_ -match '^\s*subsystem\s+' }
+            $subsystems.Count | Should -Be 4
+            $subsystems | Should -Contain "subsystem newsub1 /path/to/newsub1"
+            $subsystems | Should -Contain "subsystem newsub2 /path/to/newsub2"
```
SteveL-MSFT, line 309: Although the order isn't important here, should it verify the new ones are appended?

resources/sshdconfig/src/canonical_properties.rs
```diff
@@ -0,0 +1,112 @@
...
+    /// Metadata object containing filepath and other configuration info
+    Metadata,
+    /// Boolean flag indicating if an entry should exist or be removed
+    Exist,
+    /// Boolean flag indicating if non-specified entries should be removed
+    Purge,
+    /// Contains SSH default values inherited from system
+    InheritedDefaults,
+    /// Boolean flag to include default values in output
+    IncludeDefaults,
```
SteveL-MSFT: Alpha order?

dsc/tests/dsc_sshdconfig.tests.ps1
SteveL-MSFT: I still see this hardcoded to `c:` drive?

resources/sshdconfig/src/parser.rs
```diff
@@ -8,7 +8,14 @@ use tracing::debug;
 use tree_sitter::Parser;
 
 use crate::error::SshdConfigError;
-use crate::metadata::{MULTI_ARG_KEYWORDS_COMMA_SEP, MULTI_ARG_KEYWORDS_SPACE_SEP, REPEATABLE_KEYWORDS};
+use crate::repeat_keyword::{KeywordInfo, ValueSeparator};
+
+/// Unescape backslashes in strings. sshd -T outputs paths with escaped backslashes
+/// (e.g., "c:\\openssh\\bin\\sftp.exe") but we want to normalize them to the original
+/// form (e.g., "c:\openssh\bin\sftp.exe") for comparison and storage.
+fn unescape_backslashes(s: &str) -> String {
```
SteveL-MSFT, line 17: Does this helper only need to do this on Windows?

resources/sshdconfig/tests/sshdconfigRepeat.tests.ps1
```diff
@@ -0,0 +1,216 @@
...
+Describe 'sshd-config-repeat Set Tests' -Skip:($skipTest) {
+    BeforeAll {
+        # Create a temporary test directory for sshd_config files
+        $TestDir = Join-Path $TestDrive "sshd_test"
+        New-Item -Path $TestDir -ItemType Directory -Force | Out-Null
+        $TestConfigPath = Join-Path $TestDir "sshd_config"
+
+        # Define OS-specific paths with spaces
+        if ($IsWindows) {
+            $script:PathWithSpaces = "C:\Program Files\OpenSSH\sftp-server.exe"
```
SteveL-MSFT: Same hardcoded `C:` driver here

resources/sshdconfig/tests/sshdconfigRepeatList.tests.ps1
```diff
@@ -0,0 +1,324 @@
...
+Describe 'sshd-config-repeat-list Set Tests' -Skip:($skipTest) {
+    BeforeAll {
+        # Create a temporary test directory for sshd_config files
+        $TestDir = Join-Path $TestDrive "sshd_test"
+        New-Item -Path $TestDir -ItemType Directory -Force | Out-Null
+        $TestConfigPath = Join-Path $TestDir "sshd_config"
+
+        # Define OS-specific paths with spaces
+        if ($IsWindows) {
+            $script:PathWithSpaces = "C:\Program Files\OpenSSH\sftp-server.exe"
```
SteveL-MSFT: Hardcoded `C:` drive

dsc/tests/dsc_sshdconfig.tests.ps1
tgauth: I must have forgot to commit but it's just to make the value realistic for Windows - does not impact the test outcome

resources/sshdconfig/src/parser.rs
tgauth: yes, will fix

PR #1374, opened 2026-02-02
Update set what-if handling

### LINKED ISSUE #1361: Add new arg type for `set` and `delete` to support what-if

Need a way to specify to delete to run in `what-if` mode.  Currently we have a dedicated operation in the resource manifest for `whatif`, but it only applies to `set`.  We should deprecate that operation and instead have a new argKind `whatIfArg` that gets passed if `set` or `delete` is being run in what-if mode and the engine can see if this arg is in the arg list to determine if it falls back to synthetic what-if

_No response_

tgauth:
# PR Summary
- adds `whatIfArg` within set block in manifest
- adds deprecation warning when what-if block in manifest is used and removes what-if from capabilities for dsc list
- splits `process_args()` into `process_get_args()` and `process_set_delete_args()`

## PR Context
- fix #1361 (delete PR to follow)

### DIFF
--- a/dsc/tests/dsc_whatif.tests.ps1
+++ b/dsc/tests/dsc_whatif.tests.ps1
@@ -15,14 +15,15 @@ Describe 'whatif tests' {
                 output: hello
 "@
         $what_if_result = $config_yaml | dsc config set -w -f - | ConvertFrom-Json
+        $LASTEXITCODE | Should -Be 0
         $set_result = $config_yaml | dsc config set -f - | ConvertFrom-Json
+        $LASTEXITCODE | Should -Be 0
         $what_if_result.metadata.'Microsoft.DSC'.executionType | Should -BeExactly 'whatIf'
         $what_if_result.results.result.beforeState.output | Should -Be $set_result.results.result.beforeState.output
         $what_if_result.results.result.afterState.output | Should -Be $set_result.results.result.afterState.output
         $what_if_result.results.result.changedProperties | Should -Be $set_result.results.result.changedProperties
         $what_if_result.hadErrors | Should -BeFalse
         $what_if_result.results.Count | Should -Be 1
-        $LASTEXITCODE | Should -Be 0
     }
 
     It 'config set whatif when actual state does not match desired state' -Skip:(!$IsWindows) {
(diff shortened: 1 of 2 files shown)


dsc/tests/dsc_whatif.tests.ps1
```diff
@@ -97,4 +97,41 @@ Describe 'whatif tests' {
...
+        $config_yaml = @"
+        `$schema: https://aka.ms/dsc/schemas/v3/bundled/config/document.json
+        resources:
+        - name: WhatIfArgKind
+          type: Test/WhatIfArgKind
+          properties:
+            executionType: Actual
+"@
+        $what_if_result = $config_yaml | dsc config set -w -f - | ConvertFrom-Json
+        $set_result = $config_yaml | dsc config set -f - | ConvertFrom-Json
```
SteveL-MSFT: In general, anytime running `dsc`, should check `$LASTEXITCODE`, so there would be two checks here

dsc/tests/dsc_whatif.tests.ps1
```diff
@@ -97,4 +97,41 @@ Describe 'whatif tests' {
...
+          properties:
+            output: test
+"@
+        $what_if_result = $config_yaml | dsc config set -w -f - | ConvertFrom-Json
+        $set_result = $config_yaml | dsc config set -f - | ConvertFrom-Json
+        $what_if_result.metadata.'Microsoft.DSC'.executionType | Should -BeExactly 'whatIf'
+        $set_result.metadata.'Microsoft.DSC'.executionType | Should -BeExactly 'actual'
+        $what_if_result.hadErrors | Should -BeFalse
+        $set_result.hadErrors | Should -BeFalse
+        $LASTEXITCODE | Should -Be 0
```
SteveL-MSFT, line 137: Check exit code immediately after running `dsc`

lib/dsc-lib/src/dscresources/command_resource.rs
SteveL-MSFT: Should update the description of this function since it's now specialized to `get`

SteveL-MSFT (approved):

PR #1375, opened 2026-02-02
Enable support for adapted resource manifests

SteveL-MSFT:
# PR Summary

Enable support for adapted resource manifests.  This allows adapted resources to pre-define a simpler resource manifest pointing to an adapter without requiring the adapter to perform discovery.  The schema is what adapters are expected to return for resources via the `list` operation.  The `path` field is required and must point to an existing file.  The adapted manifest also supports an optional `condition` field that ignores the resource if it evaluates to `false`.

For manifest lists, there is a new `adaptedResources` array property.

Much of the changes is that instead of passing the resource type name across the internal APIs, it passes a `DscResource` instance instead.  Because the `path` and `directory` fields of `DscResource` are private, some code was changed to use `::new()` to construct and populate the public fields while `set_path()` and `set_directory()` are needed to set them.  `get_path()` and `get_directory()` will return an error if either are not populated (this shouldn't happen unless there's a bug in the engine).  Because the `DscResource` struct is being reused as the adapted resource manifest schema, several fields needed to be changed to `Option<>`.

To enable future additions, a new `CommandResourceInfo` struct is used to pass the resource type name and optionally the path.  The path is needed with a new arg kind `resourcePathArg` which will pass the fully resolved path to the adapter so it doesn't need to perform discovery (update to PSAdapter will come separately).

The `Test/Adapter` resource was enhanced to cover the new test cases.

It appears that resource output validation against its schema wasn't enforced and this PR fixes that so some resources/tests needed to be udpated.

## PR Context

Thought I had opened an issue for this, but can't find it

### DIFF
--- a/tools/dsctest/adaptedTest.dsc.adaptedResource.json
+++ b/tools/dsctest/adaptedTest.dsc.adaptedResource.json
@@ -16,9 +16,9 @@
     "schema": {
         "embedded": {
             "$schema": "http://json-schema.org/draft-07/schema#",
-            "$id": "https://raw.githubusercontent.com/PowerShell/DSC/main/schemas/resources/Microsoft/OSInfo/v0.1.0/schema.json",
-            "title": "OsInfo",
-            "description": "Returns information about the operating system.\n\nhttps://learn.microsoft.com/powershell/dsc/reference/microsoft/osinfo/resource\n",
+            "$id": "https://raw.githubusercontent.com/PowerShell/DSC/main/schemas/resources/Adapted/Three/v0.1.0/schema.json",
+            "title": "Adapted/Three",
+            "description": "An adapted resource for testing.",
             "type": "object",
             "required": [],
             "additionalProperties": false,


lib/dsc-lib/src/discovery/command_discovery.rs
```diff
@@ -720,10 +757,28 @@ pub fn load_manifest(path: &Path) -> Result<Vec<ImportedManifest>, DscError> {
...
+        if let Some(adapted_resources) = &manifest_list.adapted_resources {
+            for resource in adapted_resources {
+                let directory = path.parent().unwrap();
+                let resource_path = directory.join(resource.get_path()?);
+                if !resource_path.exists() {
+                    warn!("{}", t!("discovery.commandDiscovery.adaptedResourcePathNotFound", path = resource_path.to_string_lossy(), resource = resource.type_name).to_string());
+                    continue;
+                }
+                if !evaluate_condition(resource.condition.as_deref())? {
+                    debug!("{}", t!("discovery.commandDiscovery.conditionNotMet", path = path.to_string_lossy(), condition = resource.condition.as_ref() : {:?}, resource = resource.type_name));
```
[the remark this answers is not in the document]
SteveL-MSFT: Incorrect for the macro

ThomasNieto:
The `requireAdapter` in the adapted resource manifest has the issue where if an org wants to fork or add different capabilities and public that adapter publicly.

lib/dsc-lib/src/discovery/command_discovery.rs
```diff
@@ -655,8 +661,42 @@ fn evaluate_condition(condition: Option<&str>) -> Result<bool, DscError> {
...
 pub fn load_manifest(path: &Path) -> Result<Vec<ImportedManifest>, DscError> {
     let contents = read_to_string(path)?;
-    let file_name_lowercase = path.file_name().and_then(OsStr::to_str).unwrap_or("").to_lowercase();
+    let Some(file_name_lowercase) = path.file_name().and_then(OsStr::to_str).map(|s| s.to_lowercase()) else {
+        return Err(DscError::InvalidManifest(t!("discovery.commandDiscovery.invalidManifestFileName", path = path.to_string_lossy()).to_string()));
+    };
     let extension_is_json = path.extension().is_some_and(|ext| ext.eq_ignore_ascii_case("json"));
+    if DSC_ADAPTED_RESOURCE_EXTENSIONS.iter().any(|ext| file_name_lowercase.ends_with(ext)) {
+        let mut resource = if extension_is_json {
+            match serde_json::from_str::<DscResource>(&contents) {
```
michaeltlombardi: Pinning the comment to this line for visibility.

Based on discussion during verbal review in a meeting yesterday, we're reusing the `DscResource` struct for the adapted resource manifest.

If at all possible, would prefer to define a new struct for this purpose due to the way that schema generation works.

Currently, we have the `DscResource` struct's JSON Schema mapped to the output schema for `dsc resource list` at `schemas/<version>/outputs/resource/list.json`:

https://github.com/PowerShell/DSC/blob/3780743aa774e7c41582c9db3b78b8b6ce0c202d/lib/dsc-lib/src/dscresources/dscresource.rs#L33

Where the generated schema and documentation is directed at what those values mean in the context of return data for users and integrating developers.

While it is _technically possible_ to write a transform to generate a second JSON Schema for the adapted resource manifest, it's clunky and a one-off practice in the repository.

I think it would be better to keep the two structs separate and implement the `From` trait for `DscResource` for both the typical resource manifest struct and the adapted resource manifest structs, which would make the code (and tests) simpler and more robust for converting between manifests and the generic representation of a DSC Resource.

SteveL-MSFT: I'll split it into two

SteveL-MSFT:
> The `requireAdapter` in the adapted resource manifest has the issue where if an org wants to fork or add different capabilities and public that adapter publicly.

Created https://github.com/PowerShell/DSC/issues/1378 to capture this

lib/dsc-lib/src/dscresources/command_resource.rs
```diff
@@ -931,14 +1041,16 @@ fn get_command_input(input_kind: Option<&InputKind>, input: &str) -> Result<Comm
...
-    debug!("{}", t!("dscresources.commandResource.verifyJson", resource = resource.resource_type));
+fn verify_json_from_manifest(resource: &DscResource, json: &str) -> Result<(), DscError> {
+    debug!("{}", t!("dscresources.commandResource.verifyJson", resource = resource.type_name));
+    let Some(manifest) = &resource.manifest else {
+        return Err(DscError::MissingManifest(resource.type_name.to_string()));
+    };
 
     // see if resource implements validate
-    if resource.validate.is_some() {
+    if  manifest.validate.is_some() {
```
tgauth: spacing nit
```suggestion
    if manifest.validate.is_some() {
```

tools/dsctest/adaptedTest.dsc.adaptedResource.json
```diff
@@ -0,0 +1,39 @@
...
+    "description": "An adapted resource for testing.",
+    "author": "DSC Team",
+    "requireAdapter": "Test/Adapter",
+    "path": "adaptedTest.dsc.adaptedResource.json",
+    "schema": {
+        "embedded": {
+            "$schema": "http://json-schema.org/draft-07/schema#",
+            "$id": "https://raw.githubusercontent.com/PowerShell/DSC/main/schemas/resources/Microsoft/OSInfo/v0.1.0/schema.json",
+            "title": "OsInfo",
+            "description": "Returns information about the operating system.\n\nhttps://learn.microsoft.com/powershell/dsc/reference/microsoft/osinfo/resource\n",
```
tgauth: may want to update this title/description for clarity, as well as the ones added to tools/dsctest/dsctest.dsc.manifests.json

SteveL-MSFT: Yeah, will fix :)

Issue #1098, opened 2025-09-10
ERROR Error: Operation: Failed to parse JSON from 'get': executable = 'powershell' stdout = 'WARNING: LCM state is changed by non-DSC operations. If you wish to change the state of LCM, please use Remove-DscConfigurationDocument cmdlet


bi-inf-user:
### Prerequisites

- [x] Write a descriptive title.
- [x] Make sure you are able to repro it on the latest version
- [x] Search the existing issues.

### Summary

When I set a configuration using the windows powershell adapter, the output cannot be parsed due to a warning written to stdout.
Seems to be related to #833 and the linked issues there.

For this specific Warning, i'd like to ask: Is there a way to configure the LCM to allow state changed from outside?

### Steps to reproduce

1. Install the xComputerManagement Module
2. Create the following Document
```{
  "$schema": "https://raw.githubusercontent.com/PowerShell/DSC/main/schemas/2023/10/config/document.json",
    "resources": [
    {
      "name": "ComputerName",
      "type": "xComputerManagement/xComputer",
      "properties": {
        "Name": "myservernamegoeshere"
      }
    }
  ]
}
```
3. run `dsc config get -f <pathToConfigFile>\001.computername.dsc.config.json` -> works
4. run `dsc config test -f <pathToConfigFile>\001.computername.dsc.config.json` -> works
5. run `dsc config set -f <pathToConfigFile>\001.computername.dsc.config.json` -> parsing output fails due to a warning in stdout

### Expected behavior

```console
`dsc config set -f <pathToConfigFile>\001.computername.dsc.config.json` shows parsed output
```

### Actual behavior

```console
C:\Users\Administrator> dsc config set -f C:\Users\Administrator\Desktop\DSC\DSC\UseCases\M123\test\001.computername.dsc.config.json
2025-09-10T05:37:17.346533Z ERROR Error: Operation: Failed to parse JSON from 'get': executable = 'powershell' stdout = 'WARNING: LCM state is changed by non-DSC operations. If you wish to change the state of LCM, please use
Remove-DscConfigurationDocument cmdlet.
[...]

Issue #423, opened 2024-04-25
Automatically handle messages from PSDSC resources


michaeltlombardi:
### Summary of the new feature / enhancement

> As a PSDSC resource author and DSCv3 user, I want the PowerShell adapter in DSCv3 to be able to automatically handle message stream objects from PSDSC resource invocations, so that I can see the messages in trace output.

Currently, any messages other than errors thrown by the DSC Resource get swallowed instead of bubbled up to the caller.

### Proposed technical implementation details (optional)

In the adapter, we're using the `Invoke()` method on the adapter module to invoke the appropriate DSC operation on a resource:

https://github.com/PowerShell/DSC/blob/9ee037ae9a4491d1c8a932a3224c3a281a313cc0/powershell-adapter/psDscAdapter/powershell.resource.ps1#L116

The [`Invoke-DscOperation`][01] function handles invoking the DSC Resource depending on its implementation details. It also handles writing JSON errors and exiting for DSC from the invocation, like when a user specifies a script-based PSDSC resource on a Linux machine:

https://github.com/PowerShell/DSC/blob/9ee037ae9a4491d1c8a932a3224c3a281a313cc0/powershell-adapter/psDscAdapter/psDscAdapter.psm1#L228-L232

I think we could write a small handler that uses stream redirection to capture all emitted stream messages, converts them to JSON[^1], and bubbles them up to DSC before continuing. Something like:

```powershell
filter Resolve-InvocationOutput {
    [cmdletbinding()]
    param(
        [Parameter(ValueFromPipeline = $true)]
        [psobject[]]$Output
    )

    foreach ($o in $Output) {
        switch ($o.GetType().FullName) {
            'System.Management.Automation.WarningRecord' {
                Write-JsonMessage -Level Warning -Message $o.Message
            }
            'System.Management.Automation.ErrorRecord' {
                Write-JsonMessage -Level Error -Message $o.Exception.Message
            }
            'System.Management.Automation.VerboseRecord' {
                Write-JsonMessage -Level Info -Message $o.Message
            }
            'System.Management.Automation.InformationRecord' {
                Write-JsonMessage -Level Info -Message $o.MessageData.ToString()
            }
            'System.Management.Automation.DebugRecord' {
                Write-JsonMessage -Level Trace -Message $o.Message
            }
            default {
                $o
            }
        }
    }
}
```
```powershell
$invokeParams = @{
  Method     = $Operation
  ModuleName = $cachedDscResourceInfo.ModuleName
[...]

PR #1379, opened 2026-02-05
Update PSAdapters to automatically convert streams to DSC traces

SteveL-MSFT:
# PR Summary

Automatically convert PS streams to DSC traces:

- `Write-Host` and `Write-Information` is INFO
- `Write-Warning` is WARN
- `Write-Error` is ERROR
- `Write-Debug` is TRACE (since we can't be certain raw information isn't written) and must be explicitly `Write-Debug -Debug`
- `Write-Verbose` is DEBUG (seemed too noisy to be INFO) and must be explicitly `Write-Verbose -Verbose`

The main changes:

- wrap the previous main code in `powershell.resource.ps1` in a scriptblock that gets executed in a runspace so streams can be captured
- set `$VerbosePreference` and `$DebugPreference` to `silentlycontinue` as it was generating too much noise in the traces, docs should state that either resources set those preferences variables to `continue` or explicitly use `Write-Verbose -Verbose` or `Write-Debug -Debug`
- since `HadErrors` is `true` if anything is written to stderr, it just gets traced but ignored.  Instead, a script is seen as failed if the `Error` stream had any records.  This means that resources need to use `-ErrorAction Ignore` instead of `-ErrorAction SilentlyContinue`, this change was made throughout the adapter
- removed use of `Write-DscTrace` in adapter and support modules have it write to the PS streams instead
- use `$env:DSC_TRACE_LEVEL` to determine which streams to actually listen to
- have the support modules emit objects and handle conversion to JSONLines in the resource script
- created new `StreamResource` class resource to validate conversion
- many tests were updated to include `-Because` with the tracelog to help identify why they broke
- updated the `powershell.dsc.yaml` example to use PS resources implicitly instead of explicitly

## PR Context

Fix https://github.com/PowerShell/DSC/issues/423
Fix https://github.com/PowerShell/DSC/issues/1098

### DIFF
--- a/adapters/powershell/psDscAdapter/powershell.resource.ps1
+++ b/adapters/powershell/psDscAdapter/powershell.resource.ps1
@@ -104,7 +104,7 @@ $ps = [PowerShell]::Create().AddScript({
 
     # Adding some debug info to STDERR
     Write-Debug -Debug ('PSVersion=' + $PSVersionTable.PSVersion.ToString())
-    Write-Debug -Debug ('PSPath=' + $PSHome)
+    Write-Debug -Debug ('PSHome=' + $PSHome)
     Write-Debug -Debug ('PSModulePath=' + $env:PSModulePath)
 
     if ($PSVersionTable.PSVersion.Major -le 5) {
--- a/adapters/powershell/psDscAdapter/psDscAdapter.psm1
+++ b/adapters/powershell/psDscAdapter/psDscAdapter.psm1
@@ -348,7 +348,7 @@ function Invoke-DscCacheRefresh {
 
             # fill in resource files (and their last-write-times) that will be used for up-do-date checks
             $lastWriteTimes = @{}
-            Get-ChildItem -Recurse -File -Path $dscResource.ParentPath -Include "*.ps1", "*.psd1", "*.psm1", "*.mof" -ErrorAction Ignore | ForEach-Object {
+            Get-ChildItem -Recurse -File -LiteralPath $dscResource.ParentPath -Include "*.ps1", "*.psd1", "*.psm1", "*.mof" -ErrorAction Ignore | ForEach-Object {
                 $lastWriteTimes.Add($_.FullName, $_.LastWriteTime)
             }
 
@@ -361,7 +361,7 @@ function Invoke-DscCacheRefresh {
 
         [dscResourceCache]$cache = [dscResourceCache]::new()
         $cache.ResourceCache = $dscResourceCacheEntries
-        $m = $env:PSModulePath -split [IO.Path]::PathSeparator | ForEach-Object { Get-ChildItem -Directory -Path $_ -Depth 1 -ErrorAction Ignore }
+        $m = $env:PSModulePath -split [IO.Path]::PathSeparator | ForEach-Object { Get-ChildItem -Directory -LiteralPath $_ -Depth 1 -ErrorAction Ignore }
         $cache.PSModulePaths = $m.FullName
         $cache.CacheSchemaVersion = $script:CurrentCacheSchemaVersion
 


adapters/powershell/psDscAdapter/powershell.resource.ps1
```diff
@@ -11,277 +11,418 @@ param(
...
-        Write-DscTrace -Operation Error -Message 'Unsupported operation. Please use one of the following: List, Get, Set, Test, Export, Validate'
+}
+$outputObjects = [System.Collections.Generic.List[Object]]::new()
+
+try {
+    $asyncResult = $ps.BeginInvoke()
+    while (-not $asyncResult.IsCompleted) {
+        Write-TraceQueue
+
+        Start-Sleep -Milliseconds 100
```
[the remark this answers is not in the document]
SteveL-MSFT: Leaving as 100

michaeltlombardi:
A couple of notes from a design/spec perspective:

- I would prefer that we map `Write-Information` to `INFO` rather than `DEBUG` given that the PowerShell documentation for `Write-Information` states:

  > Starting in Windows PowerShell 5.0, `Write-Host` is a wrapper for `Write-Information`. This allows you to use `Write-Host` to emit output to the information stream. This enables the capture or suppression of data written using `Write-Host` while preserving backwards compatibility.
  
  It's fairly common in community code to use `Write-Information` to provide non-critical information to a user that maps relatively well to the `INFO` stream. Few cmdlets emit to the information stream and it's generally less noisy than the verbose stream.

- I expect the change _requiring_ authors to use `Ignore` instead of `SilentlyContinue` to break some relatively large number of existing DSC Resources. We may need to only treat terminating errors as failures for PowerShell resources or make this an option somehow.

  For example, I did a quick search in the source for a few DSC Community PSDSC modules:
  
  - [`SqlServerDsc`](https://github.com/search?q=repo%3Adsccommunity%2FSqlServerDsc%20path%3A%2F%5Esource%5C%2F%2F%20SilentlyContinue&type=code)
  - [`ActiveDirectoryDsc`](https://github.com/search?q=repo%3Adsccommunity%2FActiveDirectoryDsc%20path%3A%2F%5Esource%5C%2F%2F%20SilentlyContinue&type=code)
  - [`ComputerManagementDsc`](https://github.com/search?q=repo%3Adsccommunity%2FComputerManagementDsc%20path%3A%2F%5Esource%5C%2F%2F%20SilentlyContinue&type=code)

SteveL-MSFT:
I’m not satisfied with the detection if the resource failed, so perhaps treat all records in error stream as non-terminating and only have unhandled exception as failure?

michaeltlombardi:
I think unhandled exceptions as failure makes sense. This would still bubble up the error messages from those commands in existing resources without causing the operation for a resource to fail (and thus halt a configuration operation).

Then resource authors can clean up their implementation to avoid surfacing erroneous error messages or we can define an option to _only_ surface unhandled exceptions or whatever makes sense.

SteveL-MSFT:
I'll make these changes:

- Information goes to `info`
- Errors in the error stream will be written as `warning` (this may still be a bit noisy due to people commonly using `SilentlyContinue`
- unhandled exceptions will be written as `Error` and result in a failure

michaeltlombardi:
> Errors in the error stream will be written as `warning` (this may still be a bit noisy due to people commonly using `SilentlyContinue`

Is it _possible_ to emit error messages without failing an operation? I thought that whether a non-adapted resource emitted error messages was separate from whether we considered an operation to have failed (e.g. emitting errors but exiting `0` still reports as a successful operation, it just _also_ had errors).

If you can emit error-level messages with exit code `0` by design then I would prefer that non-terminating errors still bubble up to DSC and higher order tools as `error` rather than `warning`. In the future, when we support options for adapters and extensions, I think it would make sense to let users decide whether to bubble up non-terminating errors as `error` messages or `warning` messages.

If the presence of error messages causes DSC to consider a resource operation as a failure even when the exit code is `0` then I think emitting non-terminating errors as warnings is coherent with the current design.

Otherwise the proposed changes all look good to me.

SteveL-MSFT:
@michaeltlombardi yes, error messages are not considered fatal, only the exit code.  I guess I was trying to cover for "badly written" resources too much that will spew errors implicitly.  However, it's probably better to trace every ErrorRecord as an error trace and leave it to resource authors to clean up their code.  Since both Error and Warning are, by default, displayed, it doesn't change much to shift non-terminating errors as warnings.

So new plan:

- just pass-through PS ErrorRecords as `error` traces and accept it'll be noisy at first (specifically the `SilentlyContinue` usage as I don't believe I can determine it was intended to be silent)
- `Write-Information` goes to `info` trace
- unhandled exception is the only way to indicate a fatal error which results in the PSAdapter exiting with non-zero exit code
- resource authors must still use `Write-Verbose -Verbose` and `Write-Debug -Debug`

adapters/powershell/psDscAdapter/powershell.resource.ps1
```diff
@@ -11,277 +11,436 @@ param(
...
+            } else {
+                Join-Path $env:HOME ".dsc" "PSAdapterCache.json"
+            }
+        }
 
-    # load private functions of psDscAdapter stub module
-    if ($PSVersionTable.PSVersion.Major -le 5) {
-        $psDscAdapter = Import-Module "$PSScriptRoot/win_psDscAdapter.psd1" -Force -PassThru
+        Remove-Item -Force -ErrorAction Ignore -Path $cacheFilePath
+        exit
```
SeeminglyScience:
```suggestion
        return
```

Shouldn't technically cause any issues currently but might if this runspace is ever passed a host. Also just may cause confusion from readers

adapters/powershell/psDscAdapter/powershell.resource.ps1
```diff
@@ -11,277 +11,436 @@ param(
...
-            else {
-                $description = ''
-            }
+                # this text comes directly from the resource manifest for v3 native resources
+                if ($DscResourceInfo.Description) {
+                    $description = $DscResourceInfo.Description
+                }
+                elseif ($module.Description) {
+                    # some modules have long multi-line descriptions. to avoid issue, use only the first line.
+                    $description = $module.Description.split("`r`n")[0]
```
SeeminglyScience:
```suggestion
                    $description = ($module.Description -split '\r?\n')[0]
```

Feel free to ignore, I know it's just white space changes. Also not necessary if you know with certainty that the source will always be CRLF.

adapters/powershell/psDscAdapter/powershell.resource.ps1
```diff
@@ -11,277 +11,436 @@ param(
...
-            if ($null -eq $dscResourceCache) {
-                Write-DscTrace -Operation Error -Message ("DSC resource '{0}' module not found." -f $ResourceType)
-                exit 1
+        { @('Get','Set','Test','Export') -contains $_ } {
+            if ($ResourceType) {
+                Write-Debug -Debug ("Using resource type override: $ResourceType")
+                $dscResourceCache = Invoke-DscCacheRefresh -Module $ResourceType.Split('/')[0]
+                if ($null -eq $dscResourceCache) {
+                    Write-Error ("DSC resource '{0}' module not found." -f $ResourceType)
+                    exit 1
```
SeeminglyScience: Hmm maybe I'm missing something, is the `1` capturable somewhere? Is this still in the same `[PowerShell]::Create().AddScript(` statement or did I miss where that ended?

adapters/powershell/psDscAdapter/powershell.resource.ps1
```diff
@@ -11,277 +11,431 @@ param(
...
-        if ($Operation -eq 'Test') {
-            $result = @{ result = $result; _inDesiredState = $inDesiredState } | ConvertTo-Json -Depth 10 -Compress
-        }
-        else {
-            $result = @{ result = $result } | ConvertTo-Json -Depth 10 -Compress
+if ($traceLevel -ge [DscTraceLevel]::Info) {
+    $null = Register-ObjectEvent -InputObject $ps.Streams.Information -EventName DataAdding -MessageData $traceQueue -Action {
+        $traceQueue = $Event.MessageData
+        if ($null -ne $EventArgs.ItemAdded.MessageData) {
+            $traceQueue.Enqueue(@{ info = $EventArgs.ItemAdded.MessageData.ToString() })
```
SeeminglyScience:
```suggestion
            $traceQueue.Enqueue(@{ info = [string]$EventArgs.ItemAdded.MessageData })
```

`MessageData` can be null. Should be exceedingly rare, but might as well cast instead.

adapters/powershell/psDscAdapter/powershell.resource.ps1
```diff
@@ -11,277 +11,431 @@ param(
...
+        $traceQueue.Enqueue(@{ trace = $EventArgs.ItemAdded.Message })
     }
-    Default {
-        Write-DscTrace -Operation Error -Message 'Unsupported operation. Please use one of the following: List, Get, Set, Test, Export, Validate'
+}
+$outputObjects = [System.Collections.Generic.List[Object]]::new()
+
+try {
+    $asyncResult = $ps.BeginInvoke()
+    while (-not $asyncResult.IsCompleted) {
```
SeeminglyScience, line 407:
```suggestion
    # This while loop with sleep must remain so that registered object events 
    # can take over the pipeline thread.
    while (-not $asyncResult.IsCompleted) {
```

Optional comment to avoid someone trying to swap to a blocking wait in the future

adapters/powershell/psDscAdapter/psDscAdapter.psm1
```diff
@@ -26,13 +13,12 @@ function Get-DSCResourceModules {
     $listPSModuleFolders = $env:PSModulePath.Split([IO.Path]::PathSeparator)
     $dscModulePsd1List = [System.Collections.Generic.HashSet[System.String]]::new()
     foreach ($folder in $listPSModuleFolders) {
-        if (!(Test-Path $folder)) {
+        if (!(Test-Path $folder -ErrorAction Ignore)) {
```
SeeminglyScience:
```suggestion
        if (!(Test-Path -LiteralPath $folder -ErrorAction Ignore)) {
```

(unless wildcard expansion is actually desired)

adapters/powershell/psDscAdapter/psDscAdapter.psm1
```diff
@@ -26,13 +13,12 @@ function Get-DSCResourceModules {
...
             continue
         }
 
-        foreach ($moduleFolder in Get-ChildItem $folder -Directory) {
-            $addModule = $false
-            foreach ($psd1 in Get-ChildItem -Recurse -Filter "$($moduleFolder.Name).psd1" -Path $moduleFolder.fullname -Depth 2) {
+        foreach ($moduleFolder in Get-ChildItem $folder -Directory -ErrorAction Ignore) {
```
SeeminglyScience:
```suggestion
        foreach ($moduleFolder in Get-ChildItem -LiteralPath $folder -Directory -ErrorAction Ignore) {
```

adapters/powershell/psDscAdapter/psDscAdapter.psm1
```diff
@@ -26,13 +13,12 @@ function Get-DSCResourceModules {
...
+            foreach ($psd1 in Get-ChildItem -Recurse -Filter "$($moduleFolder.Name).psd1" -Path $moduleFolder.fullname -Depth 2 -ErrorAction Ignore) {
```
SeeminglyScience:
```suggestion
            foreach ($psd1 in Get-ChildItem -Recurse -Filter "$($moduleFolder.Name).psd1" -LiteralPath $moduleFolder.fullname -Depth 2 -ErrorAction Ignore) {
```

`-Filter` doesn't have an equivalent but if the module name itself has a wildcard character then it'll probably break at a lot of places

adapters/powershell/psDscAdapter/psDscAdapter.psm1
SeeminglyScience, line 140:
```suggestion
    $methods = $ResourceType.GetMember(
        'Export',
        [System.Reflection.MemberTypes]::Method,
        [System.Reflection.BindingFlags]'Public, Instance')
```

very optional nit

adapters/powershell/psDscAdapter/psDscAdapter.psm1
```diff
@@ -193,12 +179,12 @@ function LoadPowerShellClassResourcesFromModule {
...
 
-    "Loading resources from module '$($moduleInfo.Path)'" | Write-DscTrace -Operation Trace
+    Write-Debug -Debug ("Loading resources from module '$($moduleInfo.Path)'")
 
     if ($moduleInfo.RootModule) {
         if (".psm1", ".ps1" -notcontains ([System.IO.Path]::GetExtension($moduleInfo.RootModule)) -and
             (-not $moduleInfo.NestedModules)) {
-            "RootModule is neither psm1 nor ps1 '$($moduleInfo.RootModule)'" | Write-DscTrace -Operation Trace
+            Write-Debug -Debug ("RootModule is neither psm1 nor ps1 '$($moduleInfo.RootModule)'")
             return [System.Collections.Generic.List[DscResourceInfo]]::new()
```
SeeminglyScience:
```suggestion
            return
```

Definitely optional nit. Since it's a function, returning an empty enumerable will be the same as not returning anything.

adapters/powershell/psDscAdapter/psDscAdapter.psm1
```diff
@@ -289,13 +275,13 @@ function Invoke-DscCacheRefresh {
                             $cache_LastWriteTime = $cache_LastWriteTime.AddTicks( - ($cache_LastWriteTime.Ticks % [TimeSpan]::TicksPerSecond));
 
                             if (-not ($file_LastWriteTime.Equals($cache_LastWriteTime))) {
-                                "Detected stale cache entry '$($_.Name)'" | Write-DscTrace
+                                Write-Debug -Debug ("Detected stale cache entry '$($_.Name)'")
```
SeeminglyScience:
```suggestion
                                Write-Debug -Debug "Detected stale cache entry '$($_.Name)'"
```

There's a few of these and I don't want to litter the PR with a bunch of suggestions all related to this. It's definitely not important, but the grouping expression can be omitted for an expandable string expression.

adapters/powershell/psDscAdapter/psDscAdapter.psm1
```diff
@@ -305,17 +291,16 @@ function Invoke-DscCacheRefresh {
                 }
 
                 if (-not $refreshCache) {
-                    "Checking cache for stale PSModulePath" | Write-DscTrace
+                    Write-Debug -Debug "Checking cache for stale PSModulePath"
 
-                    $m = $env:PSModulePath -split [IO.Path]::PathSeparator | % { Get-ChildItem -Directory -Path $_ -Depth 1 -ea SilentlyContinue }
+                    $m = $env:PSModulePath -split [IO.Path]::PathSeparator | % { Get-ChildItem -Directory -Path $_ -Depth 1 -ErrorAction Ignore }
```
SeeminglyScience:
```suggestion
                    $m = $env:PSModulePath -split [IO.Path]::PathSeparator | % { Get-ChildItem -Directory -LiteralPath $_ -Depth 1 -ErrorAction Ignore }
```

adapters/powershell/psDscAdapter/powershell.resource.ps1
SteveL-MSFT: This was a carry over from the old code before wrapping in a runspace.  Should probably just change this to a throw.

adapters/powershell/psDscAdapter/powershell.resource.ps1
```diff
@@ -11,277 +11,431 @@ param(
...
-            if ($null -eq $dscResourceCache) {
-                Write-DscTrace -Operation Error -Message ("DSC resource '{0}' module not found." -f $ResourceType)
-                exit 1
+        { @('Get','Set','Test','Export') -contains $_ } {
+            if ($ResourceType) {
+                Write-Debug -Debug ("Using resource type override: $ResourceType")
+                $dscResourceCache = Invoke-DscCacheRefresh -Module $ResourceType.Split('/')[0]
+                if ($null -eq $dscResourceCache) {
+                    Write-Error ("DSC resource '{0}' module not found." -f $ResourceType)
+                    exit 1
```
SteveL-MSFT:
```suggestion
                    throw ("DSC resource '{0}' module not found." -f $ResourceType)
```

adapters/powershell/psDscAdapter/powershell.resource.ps1
SteveL-MSFT: This was already fixed during test

adapters/powershell/psDscAdapter/psDscAdapter.psm1
SteveL-MSFT: Needed to revert this one since it throws if not found instead of returning an empty collection that later code expects

aursulis:
For what it is worth, I've tried the latest CI build of this PR to see if it fixes #833 for us at $company and it does! Looking forward to seeing it in a release :)

PR #1394, opened 2026-02-17
Update `runcommandonset` warning to specify the resource name

SteveL-MSFT:
# PR Summary

Currently when `runcommandonset` emits this warning, there's no way for the user to know it refers to `runcommandonset` since it's not mentioned.  Changed the message to state the resource.


tgauth:
Should the DSC engine prepend the resource name to traces?

tgauth (approved):

SteveL-MSFT:
> Should the DSC engine prepend the resource name to traces?

That's a good question.  Can you open a new issue on this?  The only complication I see is with adapted resources.  The engine won't know if the trace is from the adapter or the resource.

PR #1406, opened 2026-02-25
(GH-538) Initial work for exporting schemas from code

michaeltlombardi:
# PR Summary

The primary purpose of this change is to enable exporting JSON Schemas from Rust code. This change adopts the [`cargo xtask` pattern][01] to enable build tasks that require access to Rust code in our libraries.

As a maintainer or contributor, you can invoke `cargo xtask schema export` to export the next version of the DSC schemas to disk in the `schemas/vNext` folder.

This change required implementing several helpers and transformers for the generated schemas. Each subset of functionality is captured in a separate commit.

Note that this change doesn't modify the build, test, or release steps for any existing projects. This is an initial implementation that will require continuous improvement until we are able to reliably export canonical schemas for the project from source code.

## PR Context

As part of #538, we need to generate and export schemas from code to disk.

[01]: https://github.com/matklad/cargo-xtask

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -39,3 +39,7 @@ rustflags = [
 # The following is only needed for release builds
 [source.crates-io]
 replace-with = "POWERSHELL"
+
+# Enable running `cargo xtask <command>`
+[alias]
+xtask = "run --package xtask --"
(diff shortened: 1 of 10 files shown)


SteveL-MSFT:
@michaeltlombardi pull in main branch as I made a fix for CI https://github.com/PowerShell/DSC/pull/1411

PR #1409, opened 2026-02-27
Fix MSIX build logic and make the debug package be self-signed

andyleejordan:
The logic here was a _little_ wrong, and then in order to test, I had to add self-signed logic to the debug build (and left a note on how to install that for testing).

### DIFF
--- a/helpers.build.psm1
+++ b/helpers.build.psm1
@@ -1812,7 +1812,7 @@ function Build-DscMsixPackage {
     )
 
     begin {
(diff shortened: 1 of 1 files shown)


SteveL-MSFT:
@andyleejordan pick up latest from main to get the CI fix.  Also, update the release pipeline to use this instead of the legacy file to validate it works.  Don't delete the old packging.ps1 yet until we know it works completely.

PR #1410, opened 2026-02-27
Add `dsc-bicep-ext` to MSIX and expose on `PATH`

andyleejordan:
Also includes #1409.

With https://github.com/microsoft/bicep-types-dsc/commit/2429956be2336e31d73ea89c1e91d785266827ea the Bicep extension now has a facade to start `dsc-bicep-ext` and funnel requests between it and Bicep. This gets around all design assumptions in DSC that looks for resources relative to the binary executing the resources (e.g. `dsc-lib` isn't quite a library) and the sandboxing caused by bundling the resources into the MSIX package instead of a common and accessible known location.

Another win from unbundling is that the extension is serviced within DSC releases. Future work is to eliminate the facade in the Bicep extension by improving the Bicep Local Extension API to support the scenario of executing an arbitrary gRPC server.


andyleejordan:
If we build macOS and Linux packages too, I need to verify that dsc-bicep-ext is also shipped in those.

PR #1422, opened 2026-03-05
Update CFS and other OneBranch tooling

andyleejordan:
In the course of debugging why we couldn't so much as `cargo install tree-sitter-cli` I discovered a number of issues resolved in this PR.

For the `msrustup` toolchain, `ms-stable` stopped being supported in 2024, meaning under the covers we've just been using `ms-prod-1.88`. I've updated the OneBranch pipelines to specifically use `ms-prod-1.93`. The OneBranch documentation recommends a specific pin, and not e.g. `ms-prod`. They also recommend using a `rust-toolchain.toml` file, but that gets a bit in the way of our existing logic which detects and uses `msrustup` only when available.

I've switched us from the mscodehub CFS feed to a new dedicated `DSCCargoMirror` CFS feed in msazure. By not being cross-org, we don't have to deal with complicated authentication with a service connection, and can just rely on the `CargoAuthenticate` task. I know of no compelling reason to stick to the mscodehub feed when we can eliminate a lot of headache here.

If and when as a dev you cannot update the feed, there is a very hidden button titled "Allow externally-sourced versions" available only after going to the feed and searching the upstream sources for a package. I've found it sometimes just gets turned off, which then prevents us from adding packages to the feed, and this will appear as an authentication failure. Authenticating locally with either `artifacts-cargo-credprovider` or `msrustup` does indeed work, but that setting must be toggled on.

I removed `ob_restore_phase: true` as that was an old workaround for accessing network resources that were not allow-listed, and it messed up the timing of the tasks. Seems like it was put there because we were errantly adding a second `checkout: self` task (it's an additional since the OneBranch template already does that). With that gone, ordering is fine and signing and CodeQL is happy.

We also weren't passing through `-UseX64MakeAppx` when we migrated the MSIX build function, and that needs to happen for one of the OneBranch builds.

### DIFF
--- a/helpers.build.psm1
+++ b/helpers.build.psm1
@@ -1212,6 +1212,10 @@ function Update-PathEnvironment {
 function Find-MakeAppx {
     [CmdletBinding()]
     param(
+        # When packaging in OneBranch, the MSIX is created on an x64 image for
+        # the arm64 package, and our tooling expects to use the architecture
+        # passed, so we have to override it here. It may be possible to
+        # workaround this in another way, but deferring further investigation.
         [switch]$UseX64MakeAppx
     )
 
@@ -2166,14 +2170,13 @@ function Build-DscPackage {
             ArtifactDirectory = $artifactDirectory
             Architecture = $Architecture
             Release = $Release
-            UseX64MakeAppx = $UseX64MakeAppx
         }
     }
 
     process {
         Write-Verbose "Packaging DSC..."
         if ($packageType -match 'msix') {
-            Build-DscMsixPackage @buildParams -PackageType $packageType
+            Build-DscMsixPackage @buildParams -PackageType $packageType -UseX64MakeAppx:$UseX64MakeAppx
         } elseif ($packageType -eq 'tgz') {
             Build-DscTgzPackage @buildParams
         } elseif ($packageType -eq 'zip') {


andyleejordan:
Watching the OneBranch run, it failed on Windows only at the signing step due to the internal service timing out, so I think this is at least good to go for review.

Edit: found the reason for that, and for the previously added `ob_restore_phase` workaround. For some reason we had `checkout: self` on the Windows build, while the OneBranch template already does that. So we were checking out twice which trips up signing and CodeQL validation steps.

andyleejordan:
Last thing is symbols and uh, fixing the path highlighted that this wasn't working in the first place. Went back and looked at a "successful" build...

Linux
> Source indexing is not supported on this OS.
> Found 0 files.

macOS -- N/A

Windows
> Found 0 files.
> ##[warning]No files selected for indexing.

andyleejordan:
Assigning reviewers now as that's a green build. I still don't think the Windows symbols are source mapped but that's a different investigation.

andyleejordan:
By the way this is a follow-up to #1410 as I needed the OneBranch pipeline to work again so I could get the MSIX, RPM, and DEB packages and verify them.

andyleejordan:
Nope, one more thing to fix. MSIX bundle is getting created under `debug`.

andyleejordan:
Bingo, MSIX bundle now exists too.

helpers.build.psm1
```diff
@@ -2182,6 +2166,7 @@ function Build-DscPackage {
             ArtifactDirectory = $artifactDirectory
             Architecture = $Architecture
             Release = $Release
+            UseX64MakeAppx = $UseX64MakeAppx
```
SteveL-MSFT: I don't recall exactly, but I think this was here previously in packaging.ps1 to work around a problem in the past building the msix, but perhaps this isn't needed anymore since msix building happens separately on x64 and arm64 and the arm64 toolset works now?

helpers.build.psm1
```diff
@@ -370,7 +370,7 @@ function Get-RustUp {
         if ($null -ne (Get-Command msrustup -CommandType Application -ErrorAction Ignore)) {
             Write-Verbose -Verbose "Using msrustup"
             $rustup = 'msrustup'
-            $channel = 'ms-stable'
+            $channel = 'ms-prod-1.93'
```
SteveL-MSFT, line 373: Since this version is specified in multiple places, would it be possible to have something in root of repo like `msrustup-channel.json` that can be updated and used everywhere?

andyleejordan: In another PR we should setup a `rust-toolchain.toml` file per MSRustup's docs (or otherwise find a less disruptive way to specify this in one place).

helpers.build.psm1
andyleejordan: Adding for posterity, unfortunately the workaround is still necessary otherwise the "Create msix for aarch64-pc-windows-msvc" step on OneBranch fails with:

> C:\__w\1\s\build.ps1 : An error occurred: Build-DscMsixPackage : Program 'makepri.exe' failed to run: An error occurred trying to start process 'C:\Program Files (x86)\Windows Kits\10\bin\10.0.26100.0\arm64\makepri.exe' with working directory 'C:\__w\1\s'. The specified executable is not a valid application for this OS platform.At C:\__w\1\s\helpers.build.psm1:1916 char:9

andyleejordan: Yeah this step:

```- pwsh: |
    ./build.ps1 -PackageType msix -Architecture ${{ parameters.buildName }} -Release -UseX64MakeAppx
    Copy-Item ./bin/msix/*.msix "$(Build.ArtifactStagingDirectory)" -Verbose
  displayName: 'Create msix for ${{ parameters.buildName }}'
```

is executed in the default Windows image which is x64, but our build script internally will default to use the tool of the architecture specified.

helpers.build.psm1
SteveL-MSFT: We can defer this

helpers.build.psm1
```diff
@@ -2182,6 +2166,7 @@ function Build-DscPackage {
             ArtifactDirectory = $artifactDirectory
             Architecture = $Architecture
             Release = $Release
+            UseX64MakeAppx = $UseX64MakeAppx
         }
```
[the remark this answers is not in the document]
SteveL-MSFT: @andyleejordan since this param is only used for MSIX, yo ucould just move it to line 2180

helpers.build.psm1
andyleejordan: Filed https://github.com/PowerShell/DSC/issues/1425

andyleejordan:
@SteveL-MSFT that change is applied, it's low risk but I'll still kick off a pipeline to test as that's the only way to verify it. Also opened https://github.com/PowerShell/DSC/issues/1425

PR #1449, opened 2026-03-26
(GH-538) Plumb defined types for versions and names through CLI

michaeltlombardi:
# PR Summary

This set of changes:

- Defines the new `WildcardTypeName` struct and `TypeNameFilter` enum to enable strongly validated type name filters for use in discovery and in the CLI.
- Updates the resource and extension type structs to define their versions as `ResourceVersion` and `SemanticVersion` respectively.
- Updates the configuration structs to use `SemanticVersionReq` for the required DSC version and `ResourceVersionReq` for the required version of resource instances.
- Updates discovery to use `FullyQualifiedTypeName`, `ResourceVersion`, and `ResourceVersionReq`.
- Updates the CLI arguments to use `FullyQualifiedTypeName`, `TypeNameFilter`, and `ResourceVersionReq` types instead of strings, enabling fast-failing on invalid input.
- Updates the WMI and PowerShell adapters to ensure the adapted resources are returned with valid semantic versions.
- Updates the acceptance tests to match the canonicalized fields.

## PR Context

Prior to this change, the codebase in `dsc-lib` and `dsc` used strings to represent type names, filters, versions, and version requirements. This change updates the library and binary to use defined types for these values to improve safety, error messaging, and ensure that the types are canonical for the JSON Schemas.


lib/dsc-lib/src/discovery/discovery_trait.rs
```diff
@@ -13,35 +16,87 @@ pub enum DiscoveryKind {
...
     pub fn new(resource_type: &str, version: Option<&str>, adapter: Option<&str>) -> Self {
-        let version = version.map(|v| fix_semver(&v));
         Self {
-            require_adapter: adapter.map(|a| a.to_lowercase()),
-            r#type: resource_type.to_lowercase(),
-            version,
+            require_adapter: adapter.map(|a| a.parse().unwrap()),
+            r#type: resource_type.parse().unwrap(),
+            require_version: version.map(|v| v.parse().unwrap()),
+        }
```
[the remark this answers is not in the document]
michaeltlombardi: Leaving this to address in a follow-up PR targeting the bicep extension specifically, which is the last caller of this method.

lib/dsc-lib/tests/integration/types/fully_qualified_type_name.rs
```diff
@@ -1,113 +1,406 @@
...
+        )]
+        #[test_case("/Name" =>
+            FullyQualifiedTypeNameError::InvalidTypeName{
+                text: "/Name".to_string(),
+                errors: vec![
+                    FullyQualifiedTypeNameError::EmptyOwnerSegment,
+                ]
+            }; "empty owner segment before slash"
+        )]
+        #[test_case("Owner.Empty.Namespace..Segment/Name" =>
```
SteveL-MSFT, line 74: Would be good to add the variant `Owner.Trailing.Period./Name`

lib/dsc-lib/src/types/wildcard_type_name.rs
```diff
@@ -0,0 +1,716 @@
...
+                errors.push(WildcardTypeNameError::EmptyNamespaceSegment {
+                    // Insert the index as 1-based for more user-friendly error messages
+                    index: index + 1
+                });
+            } else if !validating_segment_regex.is_match(&namespace) {
+                errors.push(WildcardTypeNameError::InvalidNamespaceSegment {
+                    segment_text: namespace.clone(),
+                });
+            }
+        }
```
SteveL-MSFT: Isn't this the same as the FullyQualifiedTypeName implementation?  Can this be refactored to a helper both use?

lib/dsc-lib/src/configure/mod.rs
```diff
@@ -44,6 +44,58 @@ pub struct Configurator {
...
+///         adapter
+///     )
+/// )? else {
+///     return Err(DscError::ResourceNotFound(
+///         resource.resource_type.to_string(),
+///         resource.require_version.as_ref().map(|r| r.to_string()).unwrap_or("".to_string())
+///     ));
+/// };
+/// ```
+macro_rules! find_resource_or_error {
```
SteveL-MSFT, line 82: In this case would it be better to have a helper function instead of a macro?

lib/dsc-lib/src/configure/mod.rs
```diff
@@ -1210,8 +1257,11 @@ impl Configurator {
             // now check that each resource in the config was found
             for resource in config.resources.iter() {
                 let adapter = get_require_adapter_from_directive(&resource.directives);
-                let Some(_dsc_resource) = self.discovery.find_resource(&DiscoveryFilter::new(&resource.resource_type, resource.require_version.as_deref(), adapter.as_deref()))? else {
-                    return Err(DscError::ResourceNotFound(resource.resource_type.to_string(), resource.require_version.as_deref().unwrap_or("").to_string()));
+                let Some(_dsc_resource) = self.discovery.find_resource(&DiscoveryFilter::new_for_resource(&resource.resource_type, resource.require_version.clone(), adapter))? else {
```
SteveL-MSFT: Shouldn't this one also use the macro?

lib/dsc-lib/src/discovery/discovery_trait.rs
```diff
@@ -19,17 +18,69 @@ pub enum DiscoveryKind {
 pub struct DiscoveryFilter {
```
SteveL-MSFT, line 18: I wonder if it might be better to make the fields public and have a `Default` implementation so you can just create it specifying only the filters needed in-place instead of calling the constructor(s).

dsc/src/args.rs
```diff
@@ -178,7 +179,8 @@ pub enum ExtensionSubCommand {
     #[clap(name = "list", about = t!("args.listExtensionAbout").to_string())]
     List {
         /// Optional extension name to filter the list
-        extension_name: Option<String>,
+        #[clap(default_value_t)]
+        extension_name: TypeNameFilter,
```
SteveL-MSFT, line 183: This is a nice and elegant solution

SteveL-MSFT (changes requested):

lib/dsc-lib/src/types/wildcard_type_name.rs
michaeltlombardi: Yeah, will refactor this into a `parse_segments()` method set to public _only_ for the crate - no additional integration tests, since everything should surface through the `parse()` methods.

lib/dsc-lib/src/discovery/discovery_trait.rs
michaeltlombardi: I think that's okay, since the field values are all encapsulated types. Will make that update.

lib/dsc-lib/src/configure/mod.rs
michaeltlombardi: We could probably do a `must_find_resource()` method that bubbles up the appropriate error instead - I went with a macro in the interests of not poking around too much in the error handling.

I would eventually like to colocate the discovery errors with the discovery implementation and use the transparent passthru/wrapper model used in the type definition modules, which should make maintenance a bit easier.

lib/dsc-lib/src/discovery/discovery_trait.rs
michaeltlombardi: Actually, holding off on this for a future PR - we currently rely on the values being references (when retrieved from the type) and this will end up moving a lot more code around. We should circle back to the discovery implementation later to improve the ergonomics.

SteveL-MSFT (approved):

PR #1496, opened 2026-04-23
Windows - Secret vault DSC resources

mimachniak:
# PR Summary

Add as configuration Windows secret store as resource for better secuirty context


michaeltlombardi:
@mimachniak - was wondering why you chose to implement this resource in Rust, rather than PowerShell - it mostly seems to be arranging PowerShell invocations for the actual functionality.

I think it might be easier to maintain in PowerShell and we already have a few resources implemented as PowerShell scripts with colocated resource manifests.

mimachniak:
> @mimachniak - was wondering why you chose to implement this resource in Rust, rather than PowerShell - it mostly seems to be arranging PowerShell invocations for the actual functionality.
> 
> I think it might be easier to maintain in PowerShell and we already have a few resources implemented as PowerShell scripts with colocated resource manifests.

PowerShell was my first idead, I was trying to do in rust but I have few issue so I will go back to PowerShell I think

mimachniak:
I will close as I try another approach

PR #1507, opened 2026-05-02
feat: Add `--what-if` for `Microsoft.Windows/Service`

### LINKED ISSUE #1506: Support WhatIf for `Microsoft.Windows/Service`

As a user going through the Microsoft DSC 3.2 announcement, I can see the `Microsoft.Windows/Service` being used as WhatIf example:

Image

Unfortunately, this resource doesn't support it in this release.

_No response_

Gijsreyn:
# PR Summary

Adds what-if support for `Microsoft.Windows/Service` set operations so the resource can return projected state and `_metadata.whatIf` details without changing the service. 

## PR Context

Fixes #1506 and partially addresses #1529


Gijsreyn:
@SteveL-MSFT - I think something was wrong with one of the builds. Regardless, tests have succeeded for Windows.

resources/windows_service/locales/en-us.toml
```diff
@@ -1,7 +1,7 @@
 _version = 1
 
 [main]
-missingOperation = "Missing operation. Usage: windows_service get --input <json> | set --input <json> | export [--input <json>]"
+missingOperation = "Missing operation. Usage: windows_service get --input <json> | set [-w] --input <json> | export [--input <json>]"
```
SteveL-MSFT: Prefer to have the long version in the error message

```suggestion
missingOperation = "Missing operation. Usage: windows_service get --input <json> | set [--what-if] --input <json> | export [--input <json>]"
```

resources/windows_service/src/types.rs
```diff
@@ -87,6 +87,20 @@ pub struct WindowsService {
...
+    #[serde(rename = "_metadata", skip_serializing_if = "Option::is_none")]
+    pub metadata: Option<Metadata>,
+}
+
+/// Metadata returned by what-if operations.
+#[derive(Debug, Default, Serialize, Deserialize, Clone)]
+#[serde(rename_all = "camelCase")]
+pub struct Metadata {
+    /// Human-readable descriptions of the changes that would be applied.
+    #[serde(rename = "whatIf", skip_serializing_if = "Option::is_none")]
```
SteveL-MSFT: since you specify `camelCase` on line 99, you don't need the `rename` here

resources/windows_service/windows_service.dsc.resource.json
```diff
@@ -23,11 +23,15 @@
             {
                 "jsonInputArg": "--input",
                 "mandatory": true
+            },
+            {
+                "whatIfArg": "-w"
```
SteveL-MSFT: Better to use the long version here making it more obvious to someone reading this

```suggestion
                "whatIfArg": "--what-if"
```

resources/windows_service/src/service.rs
```diff
@@ -220,18 +221,6 @@ pub fn get_service(input: &WindowsService) -> Result<WindowsService, ServiceErro
 
     let svc = unsafe { read_service_state(service_handle.0, &service_key_name) }?;
 
-    // If both name and display_name were provided, verify they match
```
SteveL-MSFT, line 223: Thinking about this now, this change makes sense for `get` since you can specify other properties that simply get overwritten by the actual values.  Good catch.

SteveL-MSFT (changes requested):

resources/windows_service/locales/en-us.toml
Gijsreyn: Fixed!

SteveL-MSFT (approved):

PR #1514, opened 2026-05-06
Add `Microsoft.Windows/RegistryList` resource

SteveL-MSFT:
# PR Summary

Enhance the registry resource to also have a `Microsoft.Windows/RegistryList` type which contains an array of `Microsoft.Windows/Registry` entries.

- Bumped version to 1.1
- Remove tracing from registry lib since it doesn't get emitted anyways
- Simplify the code to always work off a list, so when used as `Microsoft.Windows/Registry`, the code just wraps it in a list handled in `import_input()` function
- Added new constructor in registry lib to take a `Registry` instance instead of JSON as that will get used iterating through the list
- Moved to a single registry.dsc.manifests.json file containing both resources
- `registry.exe` for `get` and `set` have a new `--list` switch indicating the new mode
- changed existing `debug` traces to `trace` since they emit the raw information which could be sensitive


lib/dsc-lib-registry/src/lib.rs
```diff
@@ -47,6 +47,28 @@ impl RegistryHelper {
         )
     }
 
+    /// Create a new `RegistryHelper` from registry configuration.
+    /// 
+    /// # Arguments
+    /// 
+    /// * `config` - The registry configuration struct.
```
tgauth:
```suggestion
    /// * `registry_entry` - The registry configuration struct.
```

lib/dsc-lib-registry/src/lib.rs
tgauth:
```suggestion
    /// * `registry_entry` - The string with registry configuration information.
```

tgauth (approved):

PR #1516, opened 2026-05-06
Small changes to GitHub templates

SteveL-MSFT:
# PR Summary

Changed based on copilot suggestions

### DIFF
--- a/.github/ISSUE_TEMPLATE/Bug_Report.yml
+++ b/.github/ISSUE_TEMPLATE/Bug_Report.yml
@@ -53,10 +53,10 @@ body:
 - type: textarea
   attributes:
     label: Expected behavior
-    render: console
+    render: shell
     placeholder: |
       dsc config get --path ./repro.dsc.config.yaml
-      
+
       results:
       - name: repro
         type: Microsoft.DSC.Debug/Echo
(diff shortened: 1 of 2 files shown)


.github/ISSUE_TEMPLATE/Feature_Request.yml
```diff
@@ -13,7 +13,7 @@ body:
       A clear and concise description of what the problem is that the
       new feature would solve. Try formulating it in user story style
       (if applicable).
-    placeholder: >-
+    placeholder: >
```
michaeltlombardi, line 16: This change removes the trimming of trailing spacing. It doesn't make a large difference here, but I'm not sure why copilot suggested it.

.github/ISSUE_TEMPLATE/Bug_Report.yml
```diff
@@ -53,10 +53,10 @@ body:
 - type: textarea
   attributes:
     label: Expected behavior
-    render: console
+    render: shell
```
michaeltlombardi, line 56: I'm not sure we actually want this change - console is accurate (we're asking users to submit the text from a terminal session, including both commands and output).

Shell highlighting will give POSIX shell syntax highlighting for the commands but may look strange for output and will incorrectly highlight when the user's shell was PowerShell.

michaeltlombardi (commented):
I'm surprised copilot recommended these changes. Except for the whitespace cleanup, the changes either don't affect the rendering (`>-` to `>`) or are semantically incorrect (`console` to `shell`).

SteveL-MSFT:
Let's just ignore these copilot suggestions

PR #1522, opened 2026-05-09
docs: Add reference documentation for `Microsoft.Windows/Service`

Gijsreyn:
# PR Summary

This pull request adds reference documentation for the built-in DSC resource `Microsoft.Windows/Service`. I've tried to do my best to follow proper syntax to easily get it merged on Microsoft Learn + redacted/omitted data to make it life version independently.


docs/reference/resources/Microsoft/Windows/Service/examples/get-service-status.md
```diff
@@ -0,0 +1,86 @@
...
+
+To retrieve the state of every service registered on the system, use the [dsc resource export][02]
+command without an input instance.
+
+```powershell
+dsc resource export --resource Microsoft.Windows/Service
+```
+
+DSC emits one JSON object per service to stdout. You can pipe the output to a file or process it
+further with other tools.
```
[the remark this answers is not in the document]
Gijsreyn: Fixed!

docs/reference/resources/Microsoft/Windows/Service/examples/configure-windows-service.md
```diff
@@ -0,0 +1,212 @@
...
+> [!IMPORTANT]
+> **Set** operations for this resource require an elevated (administrator) process context. Run
+> your terminal or PowerShell session as Administrator before using `dsc config set`.
+
+## Definition
+
+The configuration document for this example defines two instances of the `Service` resource.
+
+The first instance ensures that the Windows Update service (`wuauserv`) is stopped and configured
+for manual start. The second instance ensures that the Windows Time service (`w32tm`) is running
```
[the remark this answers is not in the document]
Gijsreyn: Fixed!

docs/reference/resources/Microsoft/Windows/Service/index.md
```diff
@@ -0,0 +1,424 @@
...
+  - [executablePath](#executablepath) - The fully qualified path to the service binary.
+  - [logonAccount](#logonaccount) - The account under which the service runs.
+  - [startType](#starttype) - The start type of the service.
+  - [status](#status) - The current or desired status of the service.
+
+- **Read-only properties:** <a id="read-only-properties"></a> The resource returns the following
+  properties, but they aren't configurable. For more information about read-only properties, see
+  the "Read-only resource properties" section in [DSC resource properties][06].
+
+  - [_exist](#_exist) - Indicates whether the service exists in the Service Control Manager.
```
michaeltlombardi, line 112: This is a departure from the canonical property contract, which states (emphasis mine):

> The `_exist` canonical property indicates that the resource **can enforce whether instances exist**, handling whether an instance should be added, updated, or removed during a set operation. This property provides shared semantics for DSC Resources and integrating tools. Resources that define this property indicate to DSC that they adhere to the contract for the canonical property.
>
> Resources should only define this property when their implementation adheres to the following behavior contract:
>
> 1. When the desired state for `_exist` is `true`, the resource expects the instance to exist. If it doesn't exist, the resource creates or adds the instance during the set operation.
> 1. When the desired state for `_exist` is `false`, the resource expects the instance to not exist. If it does exist, the resource deletes or removes the instance during the set operation.
> 1. When the get operation queries for an instance that doesn't exist, the returned JSON always defines the `_exist` property as `false`.

If we're not intending to support creation/deletion, we should rename this property to not include the `_` prefix for the canonical property (or to a different name entirely)

Otherwise we should update the resource to support creating and deleting services.

docs/reference/resources/Microsoft/Windows/Service/index.md
```diff
@@ -0,0 +1,424 @@
...
+```
+
+</details>
+
+The service key name as registered in the Service Control Manager. This is the short internal name
+used to identify the service. For example, `wuauserv` for Windows Update. This value is
+case-insensitive.
+
+When performing a **Get** operation you may supply either `name` or `displayName` (or both) to
+identify the service. For **Set** operations you must supply `name`.
```
michaeltlombardi, line 133: This indicates to me that we probably should just mark `name` as _required_ - when we implement separate schemas for export filters, `name` would be optional there.

The other way to represent this in the JSON schema would be with the following snippet:

```yaml
anyOf:
- required: [name]
- required: [displayName]
```

Which would fail validation for an instance when neither of those properties are defined.

Unfortunately, we don't have a way to make the schemas operation specific at this time.

I think we want to keep `export` as a special case though and require the same instance schema for all other operations. By that design, the `name` field should be required and `displayName` always optional.

michaeltlombardi (approved):
Completed editorial pass, ready for final technical review and merge.

docs/reference/resources/Microsoft/Windows/Service/examples/configure-windows-service.md
```diff
@@ -0,0 +1,197 @@
...
+Copy the configuration document and save it as `service.config.dsc.yaml`.
+
+## Setup
+
+The output in this example assumes that the system has the `wuauserv` service stopped with a manual
+startup and the `W32Time` service stopped with an automatic startup. You can set the system to
+this starting state with the following commands:
+
+```powershell
+Set-Service -Name wuauserv -StartupType Manual    -Status Stopped
```
SteveL-MSFT: Wouldn't it be safer to mess with the spooler service?  Or alternatively have examples to set WUA service in recommended state

docs/reference/resources/Microsoft/Windows/Service/index.md
```diff
@@ -0,0 +1,424 @@
...
+> This resource is installed with DSC itself on Windows systems.
+>
+> You can update this resource by updating DSC. When you update DSC, the updated version of this
+> resource is automatically available.
+
+## Requirements
+
+- The resource is only usable on a Windows system.
+- **Set** operations require an elevated (administrator) process context. Running `dsc` without
+  elevation when using the **Set** operation causes an access-denied error from the SCM.
```
SteveL-MSFT, line 64: When used unelevated, dsc will return an error before SCM is called

SteveL-MSFT (changes requested):

docs/reference/resources/Microsoft/Windows/Service/index.md
Gijsreyn: Fixed!

docs/reference/resources/Microsoft/Windows/Service/examples/configure-windows-service.md
Gijsreyn: Changed the examples.

Gijsreyn:
@SteveL-MSFT - changed the example to Spooler. Ready to get merged if it looks good.

docs/reference/resources/Microsoft/Windows/Service/index.md
SteveL-MSFT: I still see this wasn't changed?

docs/reference/resources/Microsoft/Windows/Service/examples/configure-windows-service.md
SteveL-MSFT: Did you push your changes?  Still see it modifying wua?

SteveL-MSFT (changes requested):

docs/reference/resources/Microsoft/Windows/Service/examples/configure-windows-service.md
Gijsreyn: Clearly, I didn't. My bad.

docs/reference/resources/Microsoft/Windows/Service/index.md
Gijsreyn: Clearly, I didn't. My bad.

Gijsreyn:
@SteveL-MSFT or @michaeltlombardi, I was looking back to the history and https://github.com/PowerShell/DSC/pull/1522/commits/dde8346d4347c50fa56b879d71807c301fa7074f resolved the remakrs of Steve, but then later they got overwritten by Mikey's changes. Which one do we prefer now? Spooler or Windows Update services?

SteveL-MSFT:
@Gijsreyn my general philosophy is that you can use WU, but only if the user runs the example and puts them in a good configuration.  Messing with Spooler is much more safe.

Gijsreyn:
> @Gijsreyn my general philosophy is that you can use WU, but only if the user runs the example and puts them in a good configuration. Messing with Spooler is much more safe.

Reverted back the earlier changes. Now Spooler is the example.

SteveL-MSFT (approved):

PR #1527, opened 2026-05-10
docs: Add `dism_dsc` reference documentation

Gijsreyn:
# PR Summary

This pull request adds reference documentation for the `dism_dsc` executable for Windows. Each resource contains 3 examples for users to get started easily.

### DIFF
--- a/docs/reference/resources/Microsoft/Windows/FeatureOnDemandList/examples/export-features-on-demand.md
+++ b/docs/reference/resources/Microsoft/Windows/FeatureOnDemandList/examples/export-features-on-demand.md
@@ -1,0 +1,151 @@
+---
+description: >
+  Examples showing how to export and filter Windows Features on Demand (capabilities) using the
+  Microsoft.Windows/FeatureOnDemandList resource.
+ms.date:     04/21/2026
+ms.topic:    reference
+title:       Export Features on Demand
+---
+
+# Export Features on Demand
+
+This example shows how you can use the `Microsoft.Windows/FeatureOnDemandList` resource to
+enumerate Windows Features on Demand (capabilities) on a system, optionally filtering the results
+by identity, state, display name, or description.
+
+> [!IMPORTANT]
+> All operations with `Microsoft.Windows/FeatureOnDemandList` require an elevated (administrator)
+> session. Run your terminal as administrator before executing these commands.
+
+## Export all capabilities
+
+To retrieve a complete list of all capabilities on the system, use the [dsc resource export][01]
+command without any input.
+
+```powershell
+dsc resource export --resource Microsoft.Windows/FeatureOnDemandList
+```
+
+DSC returns a configuration document that includes all capabilities and their current states:
+
+```yaml
+$schema: https://aka.ms/dsc/schemas/v3/bundled/config/document.json
+resources:
+- name: Microsoft.Windows/FeatureOnDemandList
+  type: Microsoft.Windows/FeatureOnDemandList
+  properties:
+    capabilities:
+    - identity: OpenSSH.Client~~~~0.0.1.0
+      state: Installed
+    - identity: OpenSSH.Server~~~~0.0.1.0
+      state: NotPresent
+    - identity: Language.Basic~~~en-US~0.0.1.0
+      state: Installed
+    # ... additional capabilities
+```
+
+> [!NOTE]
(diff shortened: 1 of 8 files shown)


michaeltlombardi:
Starting editorial pass, will be updating the commit directly and force pushing when completed. Please don't make any changes to this PR while I'm actively editing the documentation. Thank you!

docs/reference/resources/Microsoft/Windows/FeatureOnDemandList/examples/export-features-on-demand.md
```diff
@@ -0,0 +1,151 @@
...
+    - identity: Language.Basic~~~en-US~0.0.1.0
+      state: Installed
+    # ... additional capabilities
+```
+
+> [!NOTE]
+> When exporting without filters, the resource uses a fast enumeration path that returns only
+> `identity` and `state` for each capability. To retrieve additional properties such as
+> `displayName`, `description`, `downloadSize`, and `installSize`, use an export filter as shown
+> in the examples below.
```
michaeltlombardi, line 51: Not to resolve in this PR, but this is also behavior we should reconsider for the next release - the expected behavior for `get` and `export` is to return _all_ known data for an instance without requiring resource-specific steps.

If we wanted to support a limited data mode, we should define a write-only property that tells the resource to minimize the return data - this behavior should be opt-in, not opt-out.

Also, these fields (`displayName`, `description`, etc) should be marked as read-only in the JSON Schema since they aren't configurable.

docs/reference/resources/Microsoft/Windows/FeatureOnDemandList/index.md
```diff
@@ -0,0 +1,380 @@
...
+IsKey      : false
+IsReadOnly : true
+```
+
+</details>
+
+Indicates whether the capability exists on the system. The resource sets this property to `false`
+in the **Get** response when the requested `identity` is not recognized by DISM. When `_exist` is
+`false`, the `state`, `displayName`, `description`, `downloadSize`, and `installSize` properties
+are not returned.
```
michaeltlombardi, line 175: Design note: This implementation is non-canonical - the `_exist` canonical property should only be used to indicate/define whether a resource instance actually exists on the system.

We could pretty readily map this to install/uninstall for this resource and make `state` read-only.

Otherwise we should consider removing this property. If the desired capability doesn't exist we probably want to raise an error.

docs/reference/resources/Microsoft/Windows/FeatureOnDemandList/index.md
```diff
@@ -0,0 +1,380 @@
...
+```
+
+</details>
+
+The identity string that uniquely identifies the Windows capability. For **Get** and **Set**
+operations, this property is required for each entry. For **Export** operations, it's optional and
+supports wildcard (`*`) patterns for case-insensitive filtering.
+
+Capability identities typically follow the format `CapabilityName~~~~LanguageTag~Version`, for
+example `OpenSSH.Client~~~~0.0.1.0` or `Language.Basic~~~en-US~0.0.1.0`.
```
michaeltlombardi, line 154: Design note:

For the next release, we may want to consider updating the resource implementation to separate `identity` into `name`, `languageTag`, and `version` - which we concatenate for the user before handing off to the API.

As-is, this is very difficult to parse and reason about. It also requires the user to know the exact string, which may be different across operating system versions.

This resource serves a similar purpose to [PsdscResources\WindowsFeature](https://learn.microsoft.com/en-us/powershell/dsc/reference/resources/windows/windowsfeatureresource?view=dsc-1.1) but with a much less intuitive resource instance API.

Looking through this implementation I think we probably want to review the design for this resource all-up before our next release.

michaeltlombardi (approved):
Completed editorial review, ready for final technical review and merge.

Gijsreyn:
Thanks Mikey!

docs/reference/resources/Microsoft/Windows/FeatureOnDemandList/examples/export-features-on-demand.md
SteveL-MSFT: @michaeltlombardi can you open an issue to discuss what pattern we should define for resources?  In this specific case, every property is a remote call so it's SUPER slow.

SteveL-MSFT (approved):

Issue #1529, opened 2026-05-13
Support `--what-if` for built-in resources


Gijsreyn:
### Summary of the new feature / enhancement

> As a user,
> I want to be able to call `--what-if` on built-in resources,
> So I can preview the changes it is going to apply.

### Proposed technical implementation details (optional)

With DSC v3.2 shipped, a couple of new built-in resources are introduced with `get`, `set`, and `export`. But these built-in resources are perfect candidates to implement/showcase `--what-if`:

- [x] `Microsoft.Windows/FirewallRuleList`
- [x] `Microsoft.Windows/Service`
- [ ] `Microsoft.Windows/FeatureOnDemandList`
- [ ] `Microsoft.Windows/OptionalFeatureList`
- [x] `Microsoft.Windows/UpdateList`
- [x] `Microsoft.OpenSSH.SSHD/sshd_config`
- [x] `Microsoft.OpenSSH.SSHD/Subsystem`
- [x] `Microsoft.OpenSSH.SSHD/SubsystemList`
- [x] `Microsoft.OpenSSH.SSHD/Windows`

PR #1530, opened 2026-05-13
feat: Add `--what-if` for `Microsoft.Windows/FirewallRuleList`

Gijsreyn:
# PR Summary

This pull request adds `--what-if` support for `Microsoft.Windows/FirewallRuleList`. It also updates the `SKILL.md` file for re-usability across other resources created.

## PR Context

Partially addresses #1529

### DIFF
--- a/.github/skills/create-dsc-resource/SKILL.md
+++ b/.github/skills/create-dsc-resource/SKILL.md
@@ -298,7 +298,7 @@ In the resource's `*.dsc.resource.json`, add `whatIfArg` to the `set` (and `dele
 }
 ```
 
-- `whatIfArg` is the literal CLI flag DSC will append when the user runs `dsc config set --what-if`. Always use `"-w"` (short form) for consistency across resources.
+- `whatIfArg` is the literal CLI flag DSC will append when the user runs `dsc config set --what-if`. Always use `"--what-if"` (long form) for consistency across resources.
 - `whatIfReturns: "state"` tells DSC the executable prints the projected post-state JSON on stdout (same shape as `get`/`set` returns).
 - The `--list` (bulk) variant of a resource uses the same two manifest additions; do not invent new flag names.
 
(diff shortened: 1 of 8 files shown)


.github/skills/create-dsc-resource/SKILL.md
```diff
@@ -262,3 +262,302 @@ someError = "Failed to do something: %{error}"
...
+    "executable": "<resource_name>",
+    "args": [
+        "config", "delete",
+        { "jsonInputArg": "--input", "mandatory": true },
+        { "whatIfArg": "-w" }
+    ]
+}
+```
+
+- `whatIfArg` is the literal CLI flag DSC will append when the user runs `dsc config set --what-if`. Always use `"-w"` (short form) for consistency across resources.
```
SteveL-MSFT: Since most resources won't also be used as CLI tools (although technically they could), would prefer to advise agent to use long form in the manifest `--what-if` as it makes it more obvious to a human reading the manifest

SteveL-MSFT (changes requested):
Great job on adding the SKILL.md update!

.github/skills/create-dsc-resource/SKILL.md
Gijsreyn: Done!

.github/skills/create-dsc-resource/SKILL.md
```diff
@@ -262,3 +262,302 @@ someError = "Failed to do something: %{error}"
...
+
+In the resource's `*.dsc.resource.json`, add `whatIfArg` to the `set` (and `delete`, if it supports what-if) args array, and declare `whatIfReturns: "state"` on `set`:
+
+```json
+"set": {
+    "executable": "<resource_name>",
+    "args": [
+        "config", "set",
+        { "jsonInputArg": "--input", "mandatory": true },
+        { "whatIfArg": "-w" }
```
SteveL-MSFT: Should use long form here and all cases like line 296 below

SteveL-MSFT (changes requested):

.github/skills/create-dsc-resource/SKILL.md
Gijsreyn: Modified.

SteveL-MSFT (approved):

PR #1541, opened 2026-05-22
Add new `Microsoft.Windows/WindowsFeatureList` resource

### LINKED ISSUE #1540: New resource: `Microsoft.Windows/WindowsFeatureList` for Windows Server roles

DSC already ships two DSIM-backed resources, but neither can install Windows Server roles:

- `Microsoft.Windows/OptionalFeatureList` calls `DismEnableFeature` internally but does not expose the `EnableAll`, `SourcePaths`, or `LimitAccess` parameters. Without `EnableAll`, installing a role that has required sub-features (e.g. RDS-RD-Server) fails or produces an incomplete install.
- `Microsoft.Windows/FeatureOnDemandList` uses a completely different API, targeting mostly Windows 10/11 client "Feature on Demand."

> [!NOTE]
> This was a requested feature at the [OpenDSC](https://github.com/opendsc/opendsc/issues/36) side.

A new `Microsoft.Windows/WindowsFeatureList` resource backed by `DismEnableFeature` with full parameter support.

Gijsreyn:
# PR Summary

Add new `Microsoft.Windows/WindowsFeatureList` resource, including whatIf support.

> [!NOTE]
> In an earlier pull request, the rename was removed from the struct. To keep it consistent, I also removed it from the registry library.

## PR Context

Fixes #1540.


resources/windows_feature/src/dism.rs
```diff
@@ -0,0 +1,407 @@
...
+unsafe extern "system" {
+    fn LoadLibraryExW(
+        lpLibFileName: *const u16,
+        hFile: *mut c_void,
+        dwFlags: u32,
+    ) -> *mut c_void;
+}
+
+#[repr(C, packed(4))]
+struct DismFeature {
```
SteveL-MSFT: If this resource uses the DISM APIs, can it just be a resource under the common `dism` resource folder and reuse the Rust binding layer that's there already?

SteveL-MSFT (changes requested):

resources/windows_feature/src/dism.rs
Gijsreyn: Done!

SteveL-MSFT (approved):
LGTM

PR #1554, opened 2026-05-31
Change PSScript resource to not exit on non-terminating errors

### LINKED ISSUE #1547: [Windows]PowerShellScript resources hampered by use of HadErrors
### Prerequisites

### Summary

Due to longstanding issues with `HadErrors` (PowerShell/PowerShell#4613) and the [Windows]PowerShellScript resources use of `HadErrors`

https://github.com/PowerShell/DSC/blob/aa3e331d9245e606e607f7d56c206e79e53a1d82/resources/PSScript/psscript.ps1#L148-L152

after invoking any of the scripts, it is difficult to implement some patterns with these resources.  It is even more difficult to diagnose them, as you get back only a generic "it failed" message.  Manually running the script will not expose the issue obviously unless you already know about the problem (you must remove any `-ErrorAction` that is intended to make a call into a "try").

I certainly don't know enough about the PS model to strongly suggest how to do this, but the answer feels like implementing an externalized version of the suggestion to have a `HadTerminatingErrors` property.  Maybe just checking the state of `$Error` instead of `HadErrors` would be a better mechanism? If the user decides to `$Error.Clear()` in their script, it seems reasonable to treat that as success.  That of course also allows the use of `-ErrorAction` values that prevent the update of `$Error` in the first place.

### Steps to reproduce

```
# test.yml
$schema: https://aka.ms/dsc/schema/v3/bundled/config/document.json
resources:
  - name: TestHadErrors
    type: Microsoft.DSC.Transitional/PowerShellScript
    metadata:
      description: Install and default the stable Rust toolchain via rustup
    properties:
      getScript:
      testScript: |
        return [bool](Get-Command this-command-is-fake -ErrorAction SilentlyContinue)
      setScript:
```

`> dsc config test -i test.yml`

Replacing `this-command-is-fake` with an existing command produces a successful test (and in the desired state).

### Expected behavior

```console
The resource is able to execute the test without issue, returning a "not in desired state" result.
```

### Actual behavior

```console
> dsc config test -f test.yml
2026-05-28T17:54:23.691113Z ERROR PID 57844: Errors occurred during script execution.
2026-05-28T17:54:23.819646Z ERROR Failed to run process 'pwsh': Command: Resource 'pwsh' [exit code 1] manifest description: PowerShell script execution failed
2026-05-28T17:54:23.823655Z ERROR Command: Resource 'pwsh' [exit code 1] manifest description: PowerShell script execution failed
```

### Error details

```console

```

### Environment data

```PowerShell
> $PSVersionTable

PSVersion                      7.6.2
GitCommitId                    7.6.2
OS                             Microsoft Windows 10.0.28581
```

### Version

3.2.1

### Visuals

_No response_

SteveL-MSFT:
# PR Summary

The PSScript resource was too strict in that `HadErrors` simply means that `$Error` had errors in it including non-terminating (like SilentlyContinue) errors.  Change is to emit a warning message but continue.

## PR Context

Fix https://github.com/PowerShell/DSC/issues/1547

### DIFF
--- a/resources/PSScript/psscript.ps1
+++ b/resources/PSScript/psscript.ps1
@@ -159,7 +159,7 @@ try {
 
     if ($ps.HadErrors) {
         # Errors can be non-terminating, so we just write a warning and continue
-        Write-DscTrace -Now -Level Warn -Message 'Non-terminating errors occurred during script execution.'
+        Write-DscTrace -Now -Level Debug -Message 'Non-terminating errors occurred during script execution.'
     }
 
     foreach ($output in $outputCollection) {
--- a/resources/PSScript/psscript.ps1
+++ b/resources/PSScript/psscript.ps1
@@ -138,7 +138,7 @@ try {
     $asyncResult = $ps.BeginInvoke()
     while (-not $asyncResult.IsCompleted) {
         Write-TraceQueue
-    
+
         Start-Sleep -Milliseconds 100
     }
     $outputCollection = $ps.EndInvoke($asyncResult)
@@ -146,9 +146,8 @@ try {
 
 
     if ($ps.HadErrors) {
-        # If there are any errors, we will exit with an error code
-        Write-DscTrace -Now -Level Error -Message 'Errors occurred during script execution.'
-        exit 1
+        # Errors can be non-terminating, so we just write a warning and continue
+        Write-DscTrace -Now -Level Warn -Message 'Non-terminating errors occurred during script execution.'
     }
 
     foreach ($output in $outputCollection) {
--- a/resources/PSScript/psscript.ps1
+++ b/resources/PSScript/psscript.ps1
@@ -141,6 +141,18 @@ try {
 
         Start-Sleep -Milliseconds 100
     }
+
+    if ($ps.InvocationStateInfo.State -eq 'Failed') {
+        $record  = $ps.InvocationStateInfo.Reason.ErrorRecord
+        $message = "Script failed with terminating error at line {0} for statement ``{1}`` - {2}" -f @(
+            $record.InvocationInfo.ScriptLineNumber,
+            $record.InvocationInfo.Statement.Trim(),
+            $record.Exception.ToString()
+        )
+        Write-DscTrace -Now -Level Error -Message $message
+        exit 1
+    }
+
     $outputCollection = $ps.EndInvoke($asyncResult)
     Write-TraceQueue
 
--- a/resources/PSScript/psscript.ps1
+++ b/resources/PSScript/psscript.ps1
@@ -144,11 +144,15 @@ try {
 
     if ($ps.InvocationStateInfo.State -eq 'Failed') {
         $record  = $ps.InvocationStateInfo.Reason.ErrorRecord
-        $message = "Script failed with terminating error at line {0} for statement ``{1}`` - {2}" -f @(
-            $record.InvocationInfo.ScriptLineNumber,
-            $record.InvocationInfo.Statement.Trim(),
-            $record.Exception.ToString()
-        )
+        $message = if ($null -ne $record) {
+            "Script failed with terminating error at line {0} for statement ``{1}`` - {2}" -f @(
+                $record.InvocationInfo.ScriptLineNumber,
+                $record.InvocationInfo.Line,
+                $record.Exception.ToString()
+            )
+        } else {
+            "Script failed with terminating error: $($ps.InvocationStateInfo.Reason)"
+        }
         Write-DscTrace -Now -Level Error -Message $message
         exit 1
     }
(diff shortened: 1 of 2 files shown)


resources/PSScript/psscript.tests.ps1
```diff
@@ -324,4 +324,18 @@ Describe 'Tests for PSScript resource' {
...
     }
+
+    It 'Non-terminating errors do not cause script to fail' {
+        $yaml = @'
+        getScript: |
+          $ErrorActionPreference = 'Continue'
+          Write-Error "This is an error"
+          "This should still be output"
+'@
+        $result = dsc resource get -r 'Microsoft.DSC.Transitional/PowerShellScript' -i $yaml 2> $TestDrive/error.txt | ConvertFrom-Json
```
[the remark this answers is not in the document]
SteveL-MSFT: will fix

resources/PSScript/psscript.ps1
```diff
@@ -138,17 +138,16 @@ try {
...
     Write-TraceQueue
 
 
     if ($ps.HadErrors) {
-        # If there are any errors, we will exit with an error code
-        Write-DscTrace -Now -Level Error -Message 'Errors occurred during script execution.'
-        exit 1
+        # Errors can be non-terminating, so we just write a warning and continue
+        Write-DscTrace -Now -Level Warn -Message 'Non-terminating errors occurred during script execution.'
     }
```
michaeltlombardi, line 167: I think we _do_ still need a check on `$ps.InvocationStateInfo.State` to differentiate a script that _failed_ from a script that _succeeded with errors_.

Would be good to have a test show a script that fails with both terminating and non-terminating errors.

Separately (out of scope for this PR) I think we may want to address how we emit error messages to hide some of the plumbing. Our emitted errors currently include the top-level exception (like failure on `EndInvoke` or error action set to stop) which makes it more difficult to see the underlying error and provides no useful context to the user.

resources/PSScript/psscript.ps1
```diff
@@ -138,17 +138,16 @@ try {
     $asyncResult = $ps.BeginInvoke()
     while (-not $asyncResult.IsCompleted) {
         Write-TraceQueue
-    
+
         Start-Sleep -Milliseconds 100
     }
     $outputCollection = $ps.EndInvoke($asyncResult)
```
michaeltlombardi, line 160: When the script invokes this line for a script with terminating errors the user gets a relatively arcane error message referencing this statement.

Instead, we should do something like:


```suggestion

    if ($ps.InvocationStateInfo.State -eq 'Failed') {
        $record  = $ps.InvocationStateInfo.Reason.ErrorRecord
        $message = "Script failed with terminating error at line {0} for statement ``{1}`` - {2}" -f @(
            $record.InvocationInfo.ScriptLineNumber,
            $record.InvocationInfo.Statement.Trim(),
            $record.Exception.ToString()
        )
        Write-DscTrace -Now -Level Error -Message $message
        exit 1
    }

    $outputCollection = $ps.EndInvoke($asyncResult)
```

That construction gives a reasonably useful error message for the terminating statement, such as:

```console
Script failed with terminating error at line 8 for statement `Get-Command DoesNotExist` - System.Management.Automation.CommandNotFoundException: The term 'DoesNotExist' is not recognized as a name of a cmdlet, function, script file, or executable program.
Check the spelling of the name, or if a path was included, verify that the path is correct and try again.
```

resources/PSScript/psscript.ps1
michaeltlombardi, line 167: If we use the suggestion above for exiting with the terminating error surfaced to the user then I think we can skip this warning message.

That would be preferable because this will still get emited for user-defined scripts that handle an error in a `try/catch` block, which can be confusing.

Instead of inserting a warning whenever a script encounters an error of any kind, I think it's preferable to:

1. Emit error messages from the script as they occure.
1. If the script failed due to a terminating error message, emit that message with the appropriate context and consider the resource to have failed.
1. Continue processing as normal.

This leaves messaging in the control of the user defining the scripts for this resource - emitting only the errors, warnings, and so on that they _choose_ to emit.

The resource should only emit errors and warnings of its own for how the instance is _defined_ (for example, missing parameter definition or invalid syntax).

SteveL-MSFT: There is an existing test with a thrown exception, but I can add another test with `-ErrorAction Stop`, but don't think there's additional benefit for checking `InvocationStateInfo.State` since an exception gets thrown.

resources/PSScript/psscript.ps1
SteveL-MSFT: Getting the PS ErrorRecord does seem useful here

resources/PSScript/psscript.ps1
SteveL-MSFT: I'll change this to a debug message

michaeltlombardi (approved):

PR #1556, opened 2026-06-01
Use FlushWriter to ensure traces are flushed

### LINKED ISSUE #677: Tracing is not being flushed before `dsc` exits
### Prerequisites

### Summary

Particularly on macOS, some traces do not get emitted which the tracing test is expecting.  Tracing is async and it seems that there is a race condition where the `dsc` process exists before the buffered traces get written out.

### Steps to reproduce

During CI, typically it'll be macOS that fails due to a trace test can't find the trace, but likely succeeds on rerun

### Expected behavior

```console
all traces get written out
```

### Actual behavior

```console
some traces do not
```

### Error details

```console

```

### Environment data

```PowerShell
PSVersion                      7.6.0-preview.2
GitCommitId                    7.6.0-preview.2
OS                             Darwin 24.3.0 Darwin Kernel Version 24.3.0: Thu Jan  2 20:24:06 PST 2025; root:xnu-11215.81.4~3/RELEASE_ARM64_T8103
Platform                       Unix
```

### Version

dsc 3.1.0-preview.1

### Visuals

_No response_

SteveL-MSFT:
# PR Summary

Still seeing random failures particulary on macOS due to traces not being flushed before `dsc` exits.  Doing a bit more research, it seems that macOS itself caches stdio output and thus you need to explicitly flush.  Fix here is to use a custom `FlushWriter` to flushes on every write.

## PR Context

Fix https://github.com/PowerShell/DSC/issues/677

### DIFF
--- a/dsc/src/util.rs
+++ b/dsc/src/util.rs
@@ -325,6 +325,33 @@ pub fn write_object(json: &str, format: Option<&OutputFormat>, include_separator
     }
 }
 
+struct FlushWriter<W>(W);
+
+impl<W: Write> Write for FlushWriter<W> {
+    fn write(&mut self, buf: &[u8]) -> std::io::Result<usize> {
+        let n = self.0.write(buf)?;
+        let _ = self.0.flush();
+        Ok(n)
+    }
+
+    fn flush(&mut self) -> std::io::Result<()> {
+        self.0.flush()
+    }
+}
+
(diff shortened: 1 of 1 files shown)


SteveL-MSFT:
macOS test still failed, it appears that the flushing is only when using PS based classes, so making change in PS adapter

PR #1566, opened 2026-06-10
Enable `export` operation to have own schema

### LINKED ISSUE #1232: `export` should have its own schema

`export` takes an instance of the object schema, but the schema for export is used as a filter, so this means `export` can take wildcards, ranges, etc... that may not be appropriate for `get`, `set`, `test`.

This also becomes a problem when key properties are supported, but not required for `export`

In the resource manifest under the `export` section, have a new `schema` property.  If it doesn't exist the engine continues to use the top level schema.

SteveL-MSFT:
# PR Summary

Enables resources supporting `export` to optionally have separate schema from `get`, `set`, and `test`.

In the resource manifest under the `export` operation, there is now an optional `schema` property that can be:

`command` - same as normal schema which runs a command that returns JSON schema
`embedded` - same as normal schema which has embedded JSON schema

Alternatively, a resource manifest can have a `supportsFiltering` property that is a boolean.  `schema` and `supportsFiltering` are mutually exclusive.  If this is set to `true` and any input is provided by the user, then this returns an error without calling the resource.

New test resource added to validate these scenarios particularly where the `name` property is an enum, but the `export` allows `name` to be an arbitrary string to allow passing in wildcards.

## PR Context

Fix https://github.com/PowerShell/DSC/issues/1232

### DIFF
--- a/lib/dsc-lib/src/dscresources/command_resource.rs
+++ b/lib/dsc-lib/src/dscresources/command_resource.rs
@@ -7,8 +7,8 @@ use jsonschema::Validator;
 use rust_i18n::t;
 use serde::Deserialize;
 use serde_json::{Map, Value};
-use std::{collections::HashMap, env, path::Path, process::Stdio};
-use crate::{configure::{config_doc::{ExecutionKind, SecurityContextKind}, config_result::{ResourceGetResult, ResourceTestResult}}, dscresources::{resource_manifest::SchemaArgKind}, types::ExitCodesMap, util::canonicalize_which};
+use std::{collections::HashMap, env, path::{Path, PathBuf}, process::Stdio};
+use crate::{configure::{config_doc::{ExecutionKind, SecurityContextKind}, config_result::{ResourceGetResult, ResourceTestResult}}, dscresources::resource_manifest::{ExportSchemaKind, SchemaArgKind}, types::{ExitCodesMap, FullyQualifiedTypeName}, util::canonicalize_which};
 use crate::dscerror::DscError;
 use super::{
     dscresource::{get_diff, redact, DscResource},
@@ -586,6 +586,34 @@ pub fn get_schema(resource: &DscResource, target_resource: Option<&DscResource>)
     }
 }
 
+pub fn get_export_schema(resource: &DscResource, target_resource: Option<&DscResource>) -> Result<String, DscError> {
+    let Some(manifest) = &resource.manifest else {
+        return Err(DscError::MissingManifest(resource.type_name.to_string()));
+    };
+    let Some(export) = manifest.export.as_ref() else {
+        return Err(DscError::SchemaNotAvailable(resource.type_name.to_string()));
+    };
+
+    match export.schema {
+        Some(ExportSchemaKind::Command(ref command)) => {
+            let resource_type = match target_resource {
+                Some(r) => r.type_name.clone(),
+                None => resource.type_name.clone(),
+            };
+            let args = process_schema_args(command.args.as_ref(), &CommandResourceInfo { type_name: resource_type, path: None });
+            let (_exit_code, stdout, _stderr) = invoke_command(&command.executable, args, None, Some(&resource.directory), None, manifest.exit_codes.as_ref())?;
+            Ok(stdout)
+        },
+        Some(ExportSchemaKind::Embedded(ref schema)) => {
+            let json = serde_json::to_string(schema)?;
+            Ok(json)
+        },
+        _ => {
+            get_schema(resource, target_resource)
+        }
+    }
+}
+
 /// Invoke the export operation on a resource
 ///
 /// # Arguments
@@ -636,9 +664,26 @@ pub fn invoke_export(resource: &DscResource, input: Option<&str>, target_resourc
         None => resource,
     };
     validate_security_context(&export.require_security_context, &command_resource.type_name, "export")?;
+
     if let Some(input) = input {
+        let input = if export.schema == Some(ExportSchemaKind::NoFiltering) {
+            ""
+        } else {
+            input
+        };
         if !input.is_empty() {
-            verify_json_from_manifest(resource, input, target_resource)?;
+            let schema = serde_json::from_str(&get_export_schema(resource, target_resource)?)?;
+            let compiled_schema = match Validator::new(&schema) {
+                Ok(schema) => schema,
+                Err(e) => {
+                    return Err(DscError::Schema(e.to_string()));
+                },
+            };
+            let json: Value = serde_json::from_str(input)?;
+            if let Err(err) = compiled_schema.validate(&json) {
+                return Err(DscError::Schema(err.to_string()));
+            }
+
 
             command_input = get_command_input(export.input.as_ref(), input)?;
         }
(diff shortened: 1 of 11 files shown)


lib/dsc-lib/src/dscresources/command_resource.rs
```diff
@@ -616,6 +616,34 @@ pub fn get_schema(resource: &DscResource, target_resource: Option<&DscResource>)
...
+            Ok(stdout)
+        },
+        Some(ExportSchemaKind::Embedded(ref schema)) => {
+            let json = serde_json::to_string(schema)?;
+            Ok(json)
+        },
+        _ => {
+            get_schema(resource, target_resource)
+        }
+    }
```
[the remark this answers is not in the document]
SteveL-MSFT: this is incorrect as the code does compile

lib/dsc-lib/src/dscresources/command_resource.rs
```diff
@@ -671,9 +699,26 @@ pub fn invoke_export(resource: &DscResource, input: Option<&str>, target_resourc
         type_name: resource_type.clone(),
         path,
     };
+
     if let Some(input) = input {
+        let input = if export.schema == Some(ExportSchemaKind::NoFiltering) {
+            ""
+        } else {
+            input
+        };
```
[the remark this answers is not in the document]
SteveL-MSFT: incorrect as the code does compile

tools/dsctest/src/export_schema.rs
```diff
@@ -0,0 +1,86 @@
...
+    } else {
+        ExportSchema {
+            name: "*".to_string(),
+        }
+    };
+    let filtered_instances: Vec<Schema> = if filter.name.contains("*") {
+        // convert the wildcard to a regex
+        let regex = filter.name.replace("*", ".*");
+        let regex = regex::Regex::new(&regex).unwrap();
+        instances
```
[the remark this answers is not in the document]
SteveL-MSFT: this is a test resource

lib/dsc-lib/src/dscresources/command_resource.rs
```diff
@@ -616,6 +616,60 @@ pub fn get_schema(resource: &DscResource, target_resource: Option<&DscResource>)
...
+            .map(|s| s.trim().to_string())
+            .filter(|s| !s.is_empty())
+            .unwrap_or_else(|| t!("dscresources.commandResource.resourceInvalidJson").to_string());
+        return Err(DscError::Validation(reason));
+    }
+
+    let Some(export) = manifest.export.as_ref() else {
+        return Err(DscError::SchemaNotAvailable(resource.type_name.to_string()));
+    };
+
```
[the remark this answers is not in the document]
SteveL-MSFT: I'll make this change to fallback to validate

lib/dsc-lib/src/dscresources/command_resource.rs
```diff
@@ -616,6 +616,60 @@ pub fn get_schema(resource: &DscResource, target_resource: Option<&DscResource>)
     }
 }
 
+fn verify_with_export_schema(input: &str, resource: &DscResource, target_resource: Option<&DscResource>) -> Result<(), DscError> {
```
michaeltlombardi, line 589: Rather than pulling this every time we need to export, does it make more sense to add a field to `DscResource` like `export_schema: Option<Map<String, Value>>` where `Some(_)` is populated when `export.schema` is embedded/command and `None` when `export.schema` is `"noFiltering"`?

lib/dsc-lib/src/dscresources/command_resource.rs
```diff
@@ -616,6 +616,60 @@ pub fn get_schema(resource: &DscResource, target_resource: Option<&DscResource>)
...
+            let args = process_schema_args(command.args.as_ref(), &CommandResourceInfo { type_name: resource_type, path: None });
+            let (_exit_code, stdout, _stderr) = invoke_command(&command.executable, args, None, Some(&resource.directory), None, manifest.exit_codes.as_ref())?;
+            stdout
+        },
+        Some(ExportSchemaKind::Embedded(ref schema)) => {
+            serde_json::to_string(schema)?
+        },
+        _ => {
+            get_schema(resource, target_resource)?
+        }
```
michaeltlombardi: If I'm understanding correctly, this causes `export.schema: noFiltering` to validate the instance `properties` with the instance schema instead of forbidding  `properties` from being anything but an empty object?

I would expect that a manifest which explicitly forbids filtering to error when I pass a filtering property bag.

lib/dsc-lib/src/dscresources/command_resource.rs
```diff
@@ -616,6 +616,60 @@ pub fn get_schema(resource: &DscResource, target_resource: Option<&DscResource>)
...
+            return Ok(());
+        }
+
+        let reason = result
+            .reason
+            .map(|s| s.trim().to_string())
+            .filter(|s| !s.is_empty())
+            .unwrap_or_else(|| t!("dscresources.commandResource.resourceInvalidJson").to_string());
+        return Err(DscError::Validation(reason));
+    }
```
tgauth: can this be folded into the match block on line 642, since it's also going to check if `export.schema.is_none()`?

lib/dsc-lib/src/dscresources/command_resource.rs
SteveL-MSFT: No, if `noFiltering` is declared, the input becomes empty on new line 730 below and line 735 skils the schema validation.  I'm going to split out line 655 to address Tess' feedback anyways to make NoFiltering more clear should it ever hit this function.

lib/dsc-lib/src/dscresources/command_resource.rs
SteveL-MSFT: The export schema isn't used nearly as often as the non-export schema which is used for every operation.  I think we can handle this under the caching of schema work item generally.

lib/dsc-lib/src/dscresources/command_resource.rs
SteveL-MSFT: yes, I think that makes sense

lib/dsc-lib/src/dscresources/resource_manifest.rs
```diff
@@ -206,6 +206,18 @@ pub enum SchemaKind {
...
+#[derive(Debug, Clone, PartialEq, Deserialize, Serialize, JsonSchema, DscRepoSchema)]
+#[dsc_repo_schema(base_name = "manifest.exportSchema", folder_path = "definitions")]
+#[serde(rename_all = "camelCase")]
+pub enum ExportSchemaKind {
+    /// The export schema is returned by running a command.
+    Command(SchemaCommand),
+    /// The export schema is embedded in the manifest.
+    Embedded(Value),
+    /// The export operation does not support filtering.
+    NoFiltering,
```
tgauth: `NoFiltering` is a little confusing to me - would something like `IgnoreInput` or `Unfiltered` be more clear?

SteveL-MSFT: Let's discuss in WG today

SteveL-MSFT: Discussed in WG, change to:

- in manifest, either:
  - define `schema` property with `command` or `embedded` indicating separate export schema
  - or have `supportsFiltering: false` an no `schema` property and the engine will throw error if input is provided

SteveL-MSFT: fixed

tgauth (approved):

PR #1576, opened 2026-06-15
refactoring Jason/Mikey RFC

gaelcolas:
For @theJasonHelmick and @michaeltlombardi, I've tried to refactor the RFC into something lighter, easier to chew, by moving (maybe aggressively) all related explanations to `docs/` in order to keep the RFC concise on the proposal.

All parts link back to those docs.
Let me know how that look.

### DIFF
--- a/docs/concepts/resources/class-based-psdsc-v3.md
+++ b/docs/concepts/resources/class-based-psdsc-v3.md
@@ -1,0 +1,209 @@
+---
+description: >-
+  Explains how class-based PSDSC resources can participate in DSC v3 through the
+  PowerShell adapter.
+ms.date:     06/15/2026
+ms.topic:    conceptual
+title:       Author class-based PSDSC resources for DSC v3
+---
+
+# Author class-based PSDSC resources for DSC v3
+
+This article collects the background and examples for RFC 0001, which defines the contract for
+PowerShell **class-based** PSDSC resources that participate in DSC v3 through the PowerShell
+adapter.
+
+## Related DSC concepts
+
+This article assumes you're already familiar with the shared DSC concepts and data models described
+in the following docs:
+
+- [DSC Resources](./overview.md)
+- [DSC resource operations](./operations.md)
+- [DSC resource capabilities](./capabilities.md)
+- [DSC resource properties](./properties.md)
+- [DSC resource test operation stdout schema reference](../../reference/schemas/resource/stdout/test.md)
+- [DSC resource set operation stdout schema reference](../../reference/schemas/resource/stdout/set.md)
+- [DSC Resource _inDesiredState property schema](../../reference/schemas/resource/properties/inDesiredState.md)
+
+## Relationship between PSDSC and DSC operations
+
+Class-based PSDSC resources always expose the instance methods `Get()`, `Test()`, and `Set()`.
+Through the adapter, the same class can also expose optional static methods that line up with DSC v3
+operations.
+
+| Operation | PSDSC class contract | DSC v3 adapter contract |
+| --- | --- | --- |
+| Get | `[<ResourceClass>] Get()` | `static [<ResourceClass>] Get([<ResourceClass>]$instance)` |
+| Test | `[bool] Test()` | `static [System.Tuple[bool, <ResourceClass>]] Test(...)` or `static [System.Tuple[bool, <ResourceClass>, String[]]] Test(...)` |
+| Set | `[void] Set()` | `static [void] Set(...)`, `static [<ResourceClass>] Set(...)`, or `static [System.Tuple[<ResourceClass>, String[]]] Set(...)` |
+| Delete | not available | `static [void] Delete([<ResourceClass>]$instance)` |
+| Export | not available | `static [<ResourceClass>[]] Export()` and/or `static [<ResourceClass>[]] Export([<ResourceClass>]$filteringInstance)` |
+| Schema | MOF or class properties | `static [string] InstanceJsonSchema()` |
+
+The biggest semantic differences are:
+
+- DSC **Test** returns actual state and can also return differing properties.
+- DSC **Set** returns richer change information and may support explicit `whatIf`.
+- DSC adds **Delete** and **Export**, which have no PSDSC instance-method equivalent.
+
+## The PSDSC `Reasons` pattern
+
+Many PSDSC resources adopted a `Reasons` property to surface why an instance wasn't in the desired
+state.
+
+This pattern was especially important for Azure Machine Configuration and typically required:
+
+1. A resource-specific reason type with only `Code` and `Phrase` string properties.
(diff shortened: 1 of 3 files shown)


docs/concepts/resources/class-based-psdsc-v3.md
```diff
@@ -0,0 +1,209 @@
...
+operations.
+
+| Operation | PSDSC class contract | DSC v3 adapter contract |
+| --- | --- | --- |
+| Get | `[<ResourceClass>] Get()` | `static [<ResourceClass>] Get([<ResourceClass>]$instance)` |
+| Test | `[bool] Test()` | `static [System.Tuple[bool, <ResourceClass>]] Test(...)` or `static [System.Tuple[bool, <ResourceClass>, String[]]] Test(...)` |
+| Set | `[void] Set()` | `static [void] Set(...)`, `static [<ResourceClass>] Set(...)`, or `static [System.Tuple[<ResourceClass>, String[]]] Set(...)` |
+| Delete | not available | `static [void] Delete([<ResourceClass>]$instance)` |
+| Export | not available | `static [<ResourceClass>[]] Export()` and/or `static [<ResourceClass>[]] Export([<ResourceClass>]$filteringInstance)` |
+| Schema | MOF or class properties | `static [string] InstanceJsonSchema()` |
```
SteveL-MSFT, line 42: Wouldn't we just want the JSONSchema to be in the adapted manifest?

docs/concepts/resources/class-based-psdsc-v3.md
```diff
@@ -0,0 +1,209 @@
...
+
+Class-based PSDSC resources always expose the instance methods `Get()`, `Test()`, and `Set()`.
+Through the adapter, the same class can also expose optional static methods that line up with DSC v3
+operations.
+
+| Operation | PSDSC class contract | DSC v3 adapter contract |
+| --- | --- | --- |
+| Get | `[<ResourceClass>] Get()` | `static [<ResourceClass>] Get([<ResourceClass>]$instance)` |
+| Test | `[bool] Test()` | `static [System.Tuple[bool, <ResourceClass>]] Test(...)` or `static [System.Tuple[bool, <ResourceClass>, String[]]] Test(...)` |
+| Set | `[void] Set()` | `static [void] Set(...)`, `static [<ResourceClass>] Set(...)`, or `static [System.Tuple[<ResourceClass>, String[]]] Set(...)` |
```
SteveL-MSFT, line 39: How would `--what-if` work here for both `set` and `delete`?

docs/concepts/resources/class-based-psdsc-v3.md
```diff
@@ -0,0 +1,209 @@
...
+| Delete | not available | `static [void] Delete([<ResourceClass>]$instance)` |
+| Export | not available | `static [<ResourceClass>[]] Export()` and/or `static [<ResourceClass>[]] Export([<ResourceClass>]$filteringInstance)` |
```
SteveL-MSFT, line 41: There's a PR out to add support for separate schema for `export` filtering.  Reusing the same PS class for `export` would mean that `key` still needs to be populated even though it's not needed for `export`.  Alternatively, we could say that the code uses the same type, but any constraints is purely defined in the JSONSchema.

SteveL-MSFT (changes requested):

docs/concepts/resources/class-based-psdsc-v3.md
Gijsreyn: I would partially agree here because I think they can be _both_. I think we should keep the possibility of doing validation through the adapter also. And if this static method is found, it can be used. Makes it an optional escape hatch, but you can also declare constraints that attributes cannot express.

docs/concepts/resources/class-based-psdsc-v3.md
Gijsreyn: Just two dedicated methods?

```powershell
static [System.Tuple[SoftwarePackage, string[]]] WhatIf([SoftwarePackage] $instance)        # set preview
static [System.Tuple[SoftwarePackage, string[]]] DeleteWhatIf([SoftwarePackage] $instance)   # delete preview
```

I guess it would be addable in an adapted resource manifest, but regardless, the adapter would need a change.

Gijsreyn:
@gaelcolas - nice start! You've probably already given it a thought on a base class, but just throwing it in here:

```powershell
class DscResourceBase {
    # synthetic compare of current Get() state vs desired ($this).
    # Gives every derived resource a correct Test(), free what-if, and pretest question mark?
    [bool] Test() {
        $current = $this.Get()
        foreach ($name in $this.GetConfigurableProperties()) {
            $desired = $this.$name
            if ($null -eq $desired) { continue }   # unspecified property = not enforced
            if ($current.$name -ne $desired) { return $false }
        }
        return $true
    }

    # Authors MUST override these for a writable resource.
    [object] Get() { throw [System.NotImplementedException]::new('Get is not implemented.') }
    [void]   Set() { throw [System.NotImplementedException]::new('Set is not implemented.') }

    # capability ONLY when a derived class overrides the method (see below).
    static [object[]] Export()                { throw [System.NotImplementedException]::new('Export is not implemented.') }
    static [object[]] Export([object] $filter) { throw [System.NotImplementedException]::new('Filtered Export is not implemented.') }
    static [void]     Delete([object] $instance) { throw [System.NotImplementedException]::new('Delete is not implemented.') }

    # Helper available to derived classes: configurable (non-Key, non-NotConfigurable)
    # property names, discovered by reflection over [DscProperty()] attributes.
    hidden [string[]] GetConfigurableProperties() { <# reflection over $this.GetType() #> }
}
```

docs/concepts/resources/class-based-psdsc-v3.md
SteveL-MSFT: I'd like to move away from the adapter doing validation because it means you won't get good intellisense in the future

docs/concepts/resources/class-based-psdsc-v3.md
michaeltlombardi: Could also add a check for a boolean final parameter, like
```powershell
static [System.Tuple[SoftwarePackage, string[]]] Set(
    [SoftwarePackage] $instance,
    [bool ]$whatIf
) {
    # Implementation
}
static [System.Tuple[SoftwarePackage, string[]]] Set(
    [SoftwarePackage] $instance
) {
    [SoftwarePackage]::Set($instance, $false)
}
static [System.Tuple[SoftwarePackage, string[]]] Delete(
    [SoftwarePackage] $instance,
    [bool] $whatIf
) {
    # Implementation
}
static [System.Tuple[SoftwarePackage, string[]]] Delete(
    [SoftwarePackage] $instance
) {
    [SoftwarePackage]::Delete($instance, $false)
}
```

PR #1577, opened 2026-06-16
[SSHDConfig] Move filepath from metadata to properties

### LINKED ISSUE #1570: SSHDConfig Resource: filepath should be in properties instead of metadata
The sshdconfig resource modifies the default sshd_config file, unless a different file is provided under metadata, like:
```yaml
- name: sshdconfig
  type: Microsoft.OpenSSH.SSHD/sshd_config
  metadata:
    filepath: C:\path\to\non_default\sshd_config
```
Should the filepath be provided under `properties`, instead of `metadata` since, in a way, it directly impacts the resource behavior and is not just informative?

tgauth:
# PR Summary
- move filepath from metadata to properties for sshdconfig resources
- bump sshdconfig resource versions

## PR Context
- fix #1570


dsc/tests/dsc_sshdconfig.tests.ps1
```diff
@@ -27,9 +27,8 @@ metadata:
 resources:
 - name: sshdconfig
   type: Microsoft.OpenSSH.SSHD/sshd_config
-  metadata:
-    filepath: $filepath
   properties:
+    _filepath: $filepath
```
SteveL-MSFT: We want to only use leading underscore properties for canonical properties, so in this case, it should just be `filepath`.  I think with a concern about name collision with a future SSHD keyword, we should discuss in the WG whether we elevate to a canonical property `_filepath` particularly if such a scenario would affect many different resources which might be the case.

resources/sshdconfig/sshd-subsystem.dsc.resource.json
```diff
@@ -2,7 +2,7 @@
     "$schema": "https://aka.ms/dsc/schemas/v3/bundled/resource/manifest.json",
     "type": "Microsoft.OpenSSH.SSHD/Subsystem",
     "description": "Manage SSH Server Subsystem keyword (single entry)",
-    "version": "0.1.0",
+    "version": "1.1.0",
```
SteveL-MSFT: I think we should rev this to 0.2.0.  A 1.x release indicates it's design stable and It think it's too early to say that.

SteveL-MSFT (changes requested):

dsc/tests/dsc_sshdconfig.tests.ps1
SteveL-MSFT: Suggest: `sshd_config_filepath`

SteveL-MSFT (approved):

resources/sshdconfig/src/util.rs
```diff
@@ -262,14 +262,14 @@ pub fn build_command_info(input: Option<&String>, is_get: bool) -> Result<Comman
         sshd_config = serde_json::from_str(inputs.as_str())?;
         purge = CanonicalProperties::extract_bool(&mut sshd_config, CanonicalProperty::Purge, false)?;
         include_defaults = CanonicalProperties::extract_bool(&mut sshd_config, CanonicalProperty::IncludeDefaults, is_get)?;
-        metadata = if let Some(value) = sshd_config.remove(CanonicalProperty::Metadata.as_str()) {
+        filepath = if let Some(value) = sshd_config.remove(SSHD_CONFIG_FILEPATH) {
             serde_json::from_value(value)?
         } else {
-            Metadata::new()
+            None
         };
```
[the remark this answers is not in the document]
tgauth: will update

resources/sshdconfig/src/util.rs
```diff
@@ -246,14 +246,14 @@ pub fn extract_sshd_defaults() -> Result<Map<String, Value>, SshdConfigError> {
     Ok(sshd_config)
 }
 
-/// Extract _metadata field from the input string, if it can be parsed as JSON.
+/// Extract sshd_config_filepath field from the input string, if it can be parsed as JSON.
 ///
 /// # Errors
 ///
-/// This function will return an error if it fails to parse the input string and if the _metadata field exists, extract it.
+/// This function will return an error if it fails to parse the input string and if the sshd_config_filepath field exists, extract it.
```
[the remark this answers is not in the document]
tgauth: will update comment

PR #1595, opened 2026-06-23
(GH-1580) Fix table output for no-truncate option

### LINKED ISSUE #1580: The `table-no-truncate` output format emits YAML documents for `list` subcommands
### Prerequisites

### Summary

While documenting the new `table-no-truncate` value for the `--output-format` option of the `dsc extension list` and `dsc function list` subcommands I discovered that DSC doesn't emit a table at all - instead, it emits a series of YAML documents with the `---` separator between each item.

I suspect this is caused by the last `if` statement in the following lines:

https://github.com/PowerShell/DSC/blob/0b07d11f1419e9d63ea861419302ff99e1c5220f/dsc/src/subcommand.rs#L684-L697

https://github.com/PowerShell/DSC/blob/0b07d11f1419e9d63ea861419302ff99e1c5220f/dsc/src/subcommand.rs#L785-L797

For both `list_extensions()` and `list_functions()` the code _only_ sets `write_table` to `true` when the user doesn't specify `--output-format` _and_ stdout isn't being redirected or captured.

To fix this error, I think the code would need to be something like:

```rust
let write_table = if matches!(format, Some(ListOutputFormat::TableNoTruncate)) {
    // write as table if user specified the table format
    true
} else {
    // write as table if format is not specified and interactive
    format.is_none() && io::stdout().is_terminal()
};
let mut table = Table::new(&[
    t!("subcommand.tableHeader_type").to_string().as_ref(),
    t!("subcommand.tableHeader_version").to_string().as_ref(),
    t!("subcommand.tableHeader_capabilities").to_string().as_ref(),
    t!("subcommand.tableHeader_description").to_string().as_ref(),
]);
```

And we can drop the `if` statement from the current code.

### Steps to reproduce

1. Invoke the following commands:

   ```sh
   dsc extension list --output-format table-no-truncate
   dsc function list --output-format table-no-truncate
   ```

### Expected behavior

```console
dsc extension list --output-format table-no-truncate

Type                             Version  Capabilities  Description                                              
-----------------------------------------------------------------------------------------------------------------
Microsoft.PowerShell/Discover    0.1.0    d--           Discovers DSC resources packaged in PowerShell 7 modules.
Microsoft.Windows.Appx/Discover  0.1.0    d--           Discovers DSC resources packaged as Appx packages.
```

### Actual behavior

```console
dsc extension list --output-format table-no-truncate

type: Microsoft.PowerShell/Discover
version: 0.1.0
capabilities:
- discover
import: null
path: C:\code\dsc\dsc-pr-review\bin\debug\powershell.dsc.extension.json
deprecation_message: null
description: Discovers DSC resources packaged in PowerShell 7 modules.
directory: C:\code\dsc\dsc-pr-review\bin\debug
author: null
manifest:
  $schema: https://aka.ms/dsc/schemas/v3/bundled/resource/manifest.json
  type: Microsoft.PowerShell/Discover
  version: 0.1.0
  condition: '[not(equals(tryWhich(''pwsh''), null()))]'
  description: Discovers DSC resources packaged in PowerShell 7 modules.
  discover:
    executable: pwsh
    args:
    - -NoLogo
    - -NonInteractive
    - -ExecutionPolicy
    - Bypass
    - -NoProfile
    - -Command
    - ./powershell.discover.ps1
  import: null
  importParameters: null
  secret: null

---
type: Microsoft.Windows.Appx/Discover
version: 0.1.0
capabilities:
- discover
import: null
path: C:\code\dsc\dsc-pr-review\bin\debug\appx.dsc.extension.json
deprecation_message: null
description: Discovers DSC resources packaged as Appx packages.
directory: C:\code\dsc\dsc-pr-review\bin\debug
author: null
manifest:
  $schema: https://aka.ms/dsc/schemas/v3/bundled/resource/manifest.json
  type: Microsoft.Windows.Appx/Discover
  version: 0.1.0
  description: Discovers DSC resources packaged as Appx packages.
  discover:
    executable: powershell
    args:
    - -NoLogo
    - -NonInteractive
    - -ExecutionPolicy
    - Bypass
    - -NoProfile
    - -Command
    - ./appx-discover.ps1
  import: null
  importParameters: null
  secret: null
```

### Error details

```console

```

### Environment data

```PowerShell
PSVersion                      7.6.2
GitCommitId                    7.6.2
OS                             Microsoft Windows 10.0.26200
```

### Version

3.2.0

### Visuals

_No response_

michaeltlombardi:
# PR Summary

This change:

- Updates the logic for determining whether to write a table for both commands to correctly emit a table when the `table-no-truncate` output format is specified _or_ when no output format is specified and the output isn't being redirected, assigned, or piped to another command.
- Fixes #1580

## PR Context

Prior to this change, the implementation for the `dsc function list` and `dsc extension list` commands supported the `table-no-truncate` output format, but the output wasn't emitted correctly. Instead, the command emitted the output as a series of YAML documents with the `---` separator.

### DIFF
--- a/dsc/src/subcommand.rs
+++ b/dsc/src/subcommand.rs
@@ -615,18 +615,27 @@ pub fn resource(subcommand: &ResourceSubCommand, progress_format: ProgressFormat
     }
 }
 
+/// Indicates whether to emit a table based on the output format and whether stdout is a terminal.
+fn should_write_table(format: Option<&ListOutputFormat>) -> bool {
+    if matches!(format, Some(ListOutputFormat::TableNoTruncate)) {
+        // write as table if user specified the table format
+        true
+    } else {
+        // write as table if format is not specified and interactive
+        format.is_none() && io::stdout().is_terminal()
+    }
+}
+
 fn list_extensions(dsc: &mut DscManager, extension_name: &TypeNameFilter, format: Option<&ListOutputFormat>, progress_format: ProgressFormat) {
-    let mut write_table = false;
+    let write_table = should_write_table(format);
+
     let mut table = Table::new(&[
         t!("subcommand.tableHeader_type").to_string().as_ref(),
         t!("subcommand.tableHeader_version").to_string().as_ref(),
         t!("subcommand.tableHeader_capabilities").to_string().as_ref(),
         t!("subcommand.tableHeader_description").to_string().as_ref(),
     ]);
-    if format.is_none() && io::stdout().is_terminal() {
-        // write as table if format is not specified and interactive
-        write_table = true;
-    }
+
     let mut include_separator = false;
 
     for manifest_resource in dsc.list_available(&DiscoveryKind::Extension, extension_name, None, progress_format) {
@@ -682,7 +691,7 @@ fn list_extensions(dsc: &mut DscManager, extension_name: &TypeNameFilter, format
 }
 
 fn list_functions(functions: &FunctionDispatcher, function_name: Option<&String>, output_format: Option<&ListOutputFormat>) {
-    let mut write_table = false;
+    let write_table = should_write_table(output_format);
     let mut table = Table::new(&[
         t!("subcommand.tableHeader_functionCategory").to_string().as_ref(),
         t!("subcommand.tableHeader_functionName").to_string().as_ref(),
@@ -691,10 +700,7 @@ fn list_functions(functions: &FunctionDispatcher, function_name: Option<&String>
         t!("subcommand.tableHeader_argTypes").to_string().as_ref(),
         t!("subcommand.tableHeader_description").to_string().as_ref(),
     ]);
-    if output_format.is_none() && io::stdout().is_terminal() {
-        // write as table if format is not specified and interactive
-        write_table = true;
-    }
+
     let mut include_separator = false;
     let returned_types= [
         (FunctionArgKind::Array, "a"),
--- a/dsc/tests/dsc_function_list.tests.ps1
+++ b/dsc/tests/dsc_function_list.tests.ps1
@@ -1,6 +1,14 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT License.
 
+BeforeDiscovery {
+    try {
+        $windowWidth = [Console]::WindowWidth
+    } catch {
+        $consoleUnavailable = $true
+    }
+}
+
 Describe 'Tests for function list subcommand' {
     It 'Should list all available functions' {
         $out = dsc function list | ConvertFrom-Json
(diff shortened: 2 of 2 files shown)


dsc/tests/dsc_function_list.tests.ps1
```diff
@@ -22,4 +22,16 @@ Describe 'Tests for function list subcommand' {
         $out.returnTypes | Should -Be @('String')
         $out.description | Should -Not -BeNullOrEmpty
     }
+
+    It 'Table can be not truncated' -Skip:($consoleUnavailable) {
```
SteveL-MSFT, line 34: Where is `$consoleUnavailable` defined?

dsc/src/subcommand.rs
```diff
@@ -616,17 +616,21 @@ pub fn resource(subcommand: &ResourceSubCommand, progress_format: ProgressFormat
...
 
 fn list_extensions(dsc: &mut DscManager, extension_name: &TypeNameFilter, format: Option<&ListOutputFormat>, progress_format: ProgressFormat) {
-    let mut write_table = false;
+    let write_table = if matches!(format, Some(ListOutputFormat::TableNoTruncate)) {
+        // write as table if user specified the table format
+        true
+    } else {
+        // write as table if format is not specified and interactive
+        format.is_none() && io::stdout().is_terminal()
+    };
```
SteveL-MSFT: Since the code here and in `list_functions` is the same, can we turn this into a helper used by both?

SteveL-MSFT (dismissed):

dsc/tests/dsc_function_list.tests.ps1
michaeltlombardi: Missed when updating the tests, added in latest force push.

dsc/src/subcommand.rs
michaeltlombardi: I think we _probably_ could, but when I was looking through this code I was thinking it might be better to refactor things so that we have something like a `Listable` trait for collections to move the logic closer to the type definitions. Something like...

```rust
pub trait Listable {
    /// Defines the column headers for the table output.
    pub const TABLE_HEADERS: &'static [&'static str];
    /// Returns a vector of strings representing the row data for the item.
    pub fn to_row(&self) -> Vec<String>;
    /// Determines the canonical output format to emit the items in.
    pub fn canonicalize_format(format: Option<&ListOutputFormat>) -> ListOutputFormat {
        match format {
            Some(format) => format.clone(),
            None => {
                if io::stdout().is_terminal() {
                    ListOutputFormat::Table
                } else {
                    ListOutputFormat::Json
                }
            }
        }
    }
    /// Emit the items in the specified format.
    /// 
    /// Function expects the pre-filtered list of items to emit and the format
    /// to emit them in.
    pub fn emit(items: Vec<Self>, format: Option<&ListOutputFormat>) {
        // Uses the canonicalize_format to determine the actual format to emit,
        // then calls the appropriate emit function - defined in trait, not
        // per type.
    }
    /// Emit the items as a table, optionally truncating if the table is too
    /// wide for the console.
    pub fn emit_table(items: Vec<Self>, truncate: bool) {
        let mut table = Table::new(Self::HEADER);
        for item in items {
            table.add_row(item.to_row());
        }
        table.print(truncate);
    }
    /// Emit the items as JSON objects, optionally pretty-printing.
    pub fn emit_json(items: Vec<Self>, pretty: bool) {
        todo!()
    }
    /// Emit the items as YAML objects.
    pub fn emit_yaml(items: Vec<Self>) {
        todo!()
    }
}
```

Where the `Listable` trait handles emitting the objects. Could go further and reusably define filtering/retrieving but all of those considerations are bigger than this change (been thinking about long-term maintainability as we add more structs that we will eventually surface through a `list` subcommand).

I can extract this bit for the PR if needed though.

SteveL-MSFT: I think a trait is a good idea, but a bigger refactor that can be a separate PR.  For this one, just extract that bit.

SteveL-MSFT (approved):

PR #1612, opened 2026-07-08
Split PMC stages into separate pipeline with Azure release category

SteveL-MSFT:
## Motivation

The OneBranch `DSC-Official.yml` pipeline currently runs all stages -- including the two PMC (packages.microsoft.com) release stages -- under the `NonAzure` release category. The PMC stages require the `Azure` release category, and since the release category is set at the pipeline level in OneBranch, they need to live in a separate pipeline.

## Changes

Introduces a new `.pipelines/DSC-Official-PMC.yml` pipeline that:

- Uses `Azure` as the release category
- Always runs as an official build (hardcoded -- no `OfficialBuild` parameter exposed)
- Retains the `PublishToPMC` parameter to control whether packages are staged or published
- Includes a minimal `BuildAndSign` stage with only the jobs needed for PMC artifacts (`SetPackageVersion`, `BuildLinuxMusl`, `BuildLinuxArm64Musl`)
- Contains the `PrepPackagesForEv2PMC` and `Prod_PMCRelease` stages moved from the original pipeline

The original `DSC-Official.yml` is updated to:

- Remove the two PMC stages
- Remove the now-unnecessary `PublishToPMC` parameter and variable
- Continue using `NonAzure` release category for all remaining stages


anamnavi (approved):
LGTM

PR #1621, opened 2026-07-11
feat: implement export filter functionality for resource exports

### LINKED ISSUE #1486: Allow post-filtering for export

Currently, if a user wants to use wildcard or other filtering for `export`, the resource must implemented it directly.

Allow resources to declare in manifest that it supports filtering directly.  If it doesn't, DSC engine can retrieve full results and apply post-filtering.  Strings can accept `*` wildcard.  Arrays objects are logical OR while properties within a object are a logical AND.

This would provide consistency in user experience and resources can still implement specific filtering if they want.

Gijsreyn:
# PR Summary

Adds engine-side post-filtering for export via a new exportFilter resource directive.

## PR Context

Fixes #1486.


lib/dsc-lib/src/configure/mod.rs
```diff
@@ -33,6 +33,7 @@ pub mod config_doc;
 pub mod config_result;
 pub mod constraints;
 pub mod depends_on;
+pub mod export_filter;
```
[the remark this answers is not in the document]
Gijsreyn: Fixed.

lib/dsc-lib/src/configure/export_filter.rs
```diff
@@ -0,0 +1,206 @@
...
+    }
+
+    let original_count = instances.len();
+    instances.retain(|instance| instance_matches_filters(instance, filters));
+    debug!("{}", t!("configure.export_filter.filteredInstances", original = original_count, retained = instances.len()));
+}
+
+/// Check if an instance matches any of the filter objects (logical OR).
+#[must_use]
+pub fn instance_matches_filters(instance: &Value, filters: &[Map<String, Value>]) -> bool {
```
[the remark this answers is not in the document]
Gijsreyn: Fixed.

lib/dsc-lib/src/configure/export_filter.rs
```diff
@@ -0,0 +1,206 @@
...
+use serde_json::{Map, Value};
+use tracing::debug;
+
+/// Apply an export filter to a list of exported instances, retaining only matching instances.
+///
+/// # Arguments
+///
+/// * `instances` - The exported instances to filter.
+/// * `filters` - The filter objects from the `exportFilter` directive.
+pub fn apply_export_filter(instances: &mut Vec<Value>, filters: &[Map<String, Value>]) {
```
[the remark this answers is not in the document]
Gijsreyn: Fixed.

SteveL-MSFT (changes requested):
Changes look fine, but can you remove the wildcard filtering support in the `service`, `firewall`, and `dism` resources?  I would expect all the tests to still pass (assuming they don't call the resource exe directly, if so, they would need to be udpated to go through `dsc`).

Gijsreyn:
@SteveL-MSFT - good catch, but it had some caveats. I've replaced them with tests that append `*` to a known valid identifier. Exact-match and existing AND/OR filtering coverage remained in place.

resources/windows_service/src/service.rs
```diff
@@ -877,26 +828,26 @@ unsafe fn wait_for_status(
...
         }
     }
 
-    // display_name — wildcard match
-    if let Some(ref pattern) = filter.display_name {
+    // display_name — case-insensitive exact match
+    if let Some(ref expected) = filter.display_name {
         let dn = service.display_name.as_deref().unwrap_or("");
-        if !matches_wildcard(dn, pattern) {
+        if !dn.eq_ignore_ascii_case(expected) {
```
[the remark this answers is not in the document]
Gijsreyn: Fixed.

lib/dsc-lib/src/configure/mod.rs
```diff
@@ -111,12 +113,16 @@ macro_rules! find_resource_or_error {
 /// # Errors
 ///
 /// This function will return an error if the underlying resource fails.
-pub fn add_resource_export_results_to_configuration(resource: &DscResource, conf: &mut Configuration, input: &str) -> Result<ExportResult, DscError> {
+pub fn add_resource_export_results_to_configuration(resource: &DscResource, conf: &mut Configuration, input: &str, filter: Option<&[Map<String, Value>]>) -> Result<ExportResult, DscError> {
```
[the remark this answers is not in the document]
Gijsreyn: Fixed.

SteveL-MSFT:
@Gijsreyn you have a test failure and can you see if you can improve the code coverage?

Gijsreyn:
@SteveL-MSFT - I'm a bit surprised by the coverage report. The previous [run was 41%](https://github.com/PowerShell/DSC/actions/runs/29417708042/job/87364343988), and after adding more tests, it dropped to 21% in the latest one.

SteveL-MSFT:
@Gijsreyn let me look into the CC report, I think the problem is that deleted lines is being calculated wrong as not covered

SteveL-MSFT:
@Gijsreyn as part of this PR https://github.com/PowerShell/DSC/pull/1566/changes#diff-3c6f13d4710c208b61e3b3b90dd39808010bd0ef8788005fd7ee23812fcb4e9dR1504, there is a new property in the resource manifest indicating it doesn't support filtering so it all gets handled by the engine.  I would expect all the resources mentioned here would have that set and not do any comparisons within the resource and simply returns all instances.  Be sure to pick up the last commit that fixes the codecoverage.

Gijsreyn:
@SteveL-MSFT - should be done now. All the resources in this PR now set `supportsFiltering: false` in their manifest. Tests should aso be updated to match it. One note: the DSIM-based exports use the fast enumeration path, so exported instances only include `featureName` / `identity`. If you want `displayName` and `description` in export output, it needs a per-instance DISM info call. Let me know what you prefer.

SteveL-MSFT:
> @SteveL-MSFT - should be done now. All the resources in this PR now set `supportsFiltering: false` in their manifest. Tests should aso be updated to match it. One note: the DSIM-based exports use the fast enumeration path, so exported instances only include `featureName` / `identity`. If you want `displayName` and `description` in export output, it needs a per-instance DISM info call. Let me know what you prefer.

Let's keep the current behavior for now until we get more user feedback.  When I originally wrote this and retrieved all the properties, I thought the resource was stuck since it was taking so long.

SteveL-MSFT:
@Gijsreyn sorry to do this to you, but reviewing the changes, I think for FeatureOnDemand resource we should leave it to do the filtering specifically because it handles the fast path vs complete path.  So just revert changes specific to that resource.

Gijsreyn:
@SteveL-MSFT - done! Should be good to go.

lib/dsc-lib/src/configure/mod.rs
```diff
@@ -103,20 +104,48 @@ macro_rules! find_resource_or_error {
 /// * `resource` - The resource to export.
 /// * `conf` - The configuration to add the results to.
 /// * `input` - The input to the export operation.
-///
```
SteveL-MSFT, line 108: Why was this deleted?  Should have separation between sections

resources/dism_dsc/tests/optionalFeature_export.tests.ps1
```diff
@@ -39,98 +24,10 @@ Describe 'Microsoft.Windows/OptionalFeatureList - export operation' -Skip:(!$IsW
         )
     }
 
-    It 'exports features filtered by exact featureName' -Skip:(!$isElevated) {
```
SteveL-MSFT, line 42: I think we should re-discuss this in WG, if the directive is now required then that's a breaking change.  My expectation is that all these existing tests pass as-is, the only difference is that instead of the resource performing the filtering, the engine does it.

SteveL-MSFT (changes requested):

lib/dsc-lib/src/configure/mod.rs
```diff
@@ -103,20 +104,48 @@ macro_rules! find_resource_or_error {
...
+    filter: Option<&[Map<String, Value>]>,
+) -> Result<ExportResult, DscError> {
 
     let start_datetime = chrono::Local::now();
-    let export_result = resource.export(input)?;
+    let mut export_result = resource.export(input)?;
     let end_datetime = chrono::Local::now();
 
+    if let Some(filter) = filter {
+        export_filter::apply_export_filter(&mut export_result.actual_state, filter);
```
SteveL-MSFT: I seem to be missing something here, the logic should be if there is input:

1. if the resource supports filtering, let it handle it
2. if the resource does not support filtering, use the engine to perform filtering - should also emit an INFO message that it's using engine filtering

SteveL-MSFT (changes requested):

lib/dsc-lib/src/configure/mod.rs
Gijsreyn: Mistake.

Gijsreyn:
@SteveL-MSFT - I have totally misunderstood the original intent, so my bad on that part. I thought the `exportFilter` directive was meant to replace resource-side filtering, but re-reading your comments, it's clear the engine should only step in when a resource can't filter on its own. I've updated the PR to match your comments, meaning that I've reverted all the changes on the resources by keeping their native filtering and their existing tests to not make it a breaking change.

The engine now follows the logic: if a resource supports filtering, the input is passed through to it. If the manifest declares `supportsFiltering: false`, the engine exports everything, applies the input as a filter itself, and emits an INFO message that engine-side filtering is being used (should work for both `dsc resource` and `dsc config` command).

That said, now, the `exportFilter` directive stays as an optional engine post-filter.

resources/dism_dsc/tests/optionalFeature_export.tests.ps1
Gijsreyn: Reverted.

lib/dsc-lib/src/configure/mod.rs
Gijsreyn: Please see comment below.

SteveL-MSFT:
@Gijsreyn just to be clear, some of the resources simply implement regex filtering and those should have their code removed and the manfiest updated to indicate they don't support filtering so the engine takes care of it and the exact same tests should pass.  The Feature-on-Demand one handles the special perf case, so that manifest should indicate it DOES handle filtering.  Part of the goal here is to remove the redundant regex code inside the resources where there's no benefit for the resource to do it.  I believe the concern from the WG is difference in experience for filtering so I think it makes sense to have the optional `exportFilter` which overrides the resource filtering (if implemented) and also the INFO message so the user isn't surprised, but I'm still a bit concerned about the Feature-on-Demand case (which won't be unique, but rare), so want to clarify with the WG.

lib/dsc-lib/src/configure/export_filter.rs
```diff
@@ -0,0 +1,323 @@
...
+            if let Some(elements) = instance_obj.get_mut(name).and_then(Value::as_array_mut) {
+                elements.retain(|element| {
+                    element.as_object().is_some_and(|element_obj| {
+                        element_filters.iter().any(|element_filter| instance_matches_filter(element_obj, element_filter))
+                    })
+                });
+            }
+        }
+    }
+}
```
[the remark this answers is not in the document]
Gijsreyn: Fixed.

SteveL-MSFT:
WG discussed this:

- remove the exportFilter directive
- emit INFO message when engine filter is applied
- the INFO message should indicate this feature as Experimental

Gijsreyn:
@SteveL-MSFT - made the adjustments. Let me know if this is solid.

lib/dsc-lib/src/configure/export_filter.rs
```diff
@@ -0,0 +1,355 @@
...
+            .unwrap_or_default();
+        filter_array_elements(instance, &matching);
+        true
+    });
+    debug!("{}", t!("configure.export_filter.filteredInstances", original = original_count, retained = instances.len()));
+}
+
+/// Rewrite the instance's array properties targeted by element filters, retaining only the
+/// elements that match any of the element filter objects (logical OR).
+fn filter_array_elements(instance: &mut Value, filters: &[&Map<String, Value>]) {
```
SteveL-MSFT, line 42: Seems like we should have more tests for object array filtering.  Include the case where multiple OR filters can match the same instance, but the result shouldn't have duplicates.

lib/dsc-lib/src/configure/export_filter.rs
```diff
@@ -0,0 +1,355 @@
...
+
+    #[test]
+    fn wildcard_match_star() {
+        assert!(wildcard_match("*ssh*", "ssh"));
+        assert!(wildcard_match("*ssh*", "openssh-server"));
+        assert!(wildcard_match("ssh*", "sshd"));
+        assert!(wildcard_match("*shd", "sshd"));
+        assert!(wildcard_match("*", ""));
+        assert!(wildcard_match("*", "anything"));
+        assert!(wildcard_match("s*h*d", "sshd"));
```
SteveL-MSFT, line 182: It would be good to include a test case with two asterisks in a row.  Also another test case where the string to match includes an asterisk.

SteveL-MSFT (changes requested):

lib/dsc-lib/src/configure/export_filter.rs
Gijsreyn: Fixed.

lib/dsc-lib/src/configure/export_filter.rs
Gijsreyn: Fixed.

resources/dism_dsc/optionalfeature.dsc.resource.json
```diff
@@ -95,12 +95,12 @@
                             "displayName": {
                                 "type": "string",
                                 "title": "Display name",
-                                "description": "The display name of the optional feature. Returned by get and export operations. For export, wildcards (*) are supported for case-insensitive filtering."
+                                "description": "The display name of the optional feature. Returned by the get operation."
                             },
```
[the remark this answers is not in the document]
Gijsreyn: Done.

lib/dsc-lib/src/configure/mod.rs
```diff
@@ -98,25 +99,40 @@ macro_rules! find_resource_or_error {
 
 /// Add the results of an export operation to a configuration.
 ///
+/// If the resource supports export filtering natively, any input is passed through to the
+/// resource. If it does not, the input is interpreted as a filter and the engine performs
+/// the filtering on the exported instances instead.
+///
```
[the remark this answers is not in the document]
Gijsreyn: Done.

dsc/tests/dsc_export.tests.ps1
```diff
@@ -213,3 +213,53 @@ resources:
...
+  type: Test/ExportSchemaNoFiltering
+  properties:
+    name: 'Gijs'
+'@
+        $out = dsc --trace-level info config export -i $yaml 2>$TESTDRIVE/error.log | ConvertFrom-Json
+        $errorlog = Get-Content "$TESTDRIVE/error.log" -Raw
+        $LASTEXITCODE | Should -Be 0 -Because $errorlog
+        @($out.resources).Count | Should -Be 1
+        $out.resources[0].properties.name | Should -BeExactly 'Gijs'
+        $errorlog | Should -Match 'experimental'
```
SteveL-MSFT: Can you make this slightly more specific?  You can use `-BeLike` `*INFO*Resource 'Test/ExportSchemaNoFiltering' does not support export filtering, the engine will filter the exported instances (experimental feature)*`

- make sure it's an INFO message
- in case we have other experimental feature messages in the future we don't want a false positive

resources/dism_dsc/src/optional_feature/export.rs
```diff
@@ -5,90 +5,25 @@ use rust_i18n::t;
...
-                            should_get_full = true;
-                            break;
-                        }
-                }
-            }
-            if !should_get_full {
-                // This feature cannot satisfy any name-constrained filter, and there are
-                // no filters without a feature_name, so skip the expensive get_feature_info().
-                continue;
-            }
```
SteveL-MSFT, line 59: We should revert the filtering change in this resource since it's optimizing for perf here

resources/dism_dsc/src/windows_feature/export.rs
```diff
@@ -4,96 +4,24 @@
...
-            list.features
-        }
-    };
-
+pub fn handle_export(_input: &str) -> Result<String, String> {
     let session = DismSessionHandle::open()?;
     let all_basics = session.get_all_feature_basics()?;
 
-    let needs_full_info = filters
+    let features = all_basics
```
SteveL-MSFT: Same here.  I think in general the dism api is slow when retrieving all properties, so revert all the changes under dism_dsc

SteveL-MSFT (changes requested):

dsc/tests/dsc_export.tests.ps1
Gijsreyn: Thanks, applied the suggestion.

SteveL-MSFT (changes requested):
Sorry, just occurred to me looking over the latest change that the dism resource manifests need to declare they explicitly support filtering. Although for backwards compat it doesn't need to, we want those resource manifests to represent best practice.

Gijsreyn:
> Sorry, just occurred to me looking over the latest change that the dism resource manifests need to declare they explicitly support filtering. Although for backwards compat it doesn't need to, we want those resource manifests to represent best practice.

Done! And also bumped up the version number for good measure.

SteveL-MSFT (approved):

PR #1639, opened 2026-07-21
fix: add support for preserving quoted group names in sshd_config

### LINKED ISSUE #1508: In Microsoft.OpenSSH.SSHD/sshd_config allowgroups is not processed correctly
### Prerequisites

### Summary

When sshd_conf config on windows is:
`AllowGroups administrators "openssh users"`
export with command "dsc resource export -r Microsoft.OpenSSH.SSHD/sshd_config" is:
```
    allowgroups:
    - administrators
    - openssh
    - users
```

### Steps to reproduce

On windows enable sshd service and check AllowGroups parameter in sshd_conf, it should be:
`AllowGroups administrators "openssh users"`
Export sshd configuration with command:
`dsc resource export -r Microsoft.OpenSSH.SSHD/sshd_config`

### Expected behavior

```yaml
    allowgroups:
    - administrators
    - openssh users
```

### Actual behavior

```yaml
    allowgroups:
    - administrators
    - openssh
    - users
```

### Error details

```console

```

### Environment data

```PowerShell
PSVersion                      7.6.1
GitCommitId                    7.6.1
```

### Version

3.2.0

### Visuals

_No response_

Gijsreyn:
# PR Summary

Fixes a bug where `sshd_config` list keywords with quoted, space-containing values (e.g. the Windows group "openssh users") were split into separate entries on export and get.

## PR Context

Fixes #1508.

### DIFF
--- a/resources/sshdconfig/src/get.rs
+++ b/resources/sshdconfig/src/get.rs
@@ -17,6 +17,7 @@ use crate::canonical_properties::CanonicalProperty;
 use crate::error::SshdConfigError;
 use crate::inputs::{CommandInfo, SSHD_CONFIG_FILEPATH};
 use crate::parser::parse_text_to_map;
+use crate::repeat_keyword::MULTI_ARG_KEYWORDS_SPACE_SEP;
 use crate::util::{
     build_command_info,
     extract_sshd_defaults,
@@ -133,6 +134,12 @@ pub fn get_sshd_settings(cmd_info: &CommandInfo, is_get: bool) -> Result<Map<Str
         result.insert("match".to_string(), match_value.clone());
     }
 
+    // sshd -T normalizes space-separated list keywords by stripping the quotes that preserve
+    // values containing spaces (e.g. Windows group names like "openssh users"), splitting them
+    // into separate entries. Prefer the value parsed directly from the config file, which retains
+    // the quoting, for these keywords when they are explicitly set.
+    prefer_explicit_space_sep_lists(&mut result, &explicit_settings);
+
     if cmd_info.include_defaults {
         // get default from SSHD -T with empty config
         let mut defaults = extract_sshd_defaults()?;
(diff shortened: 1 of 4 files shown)


Gijsreyn:
@tgauth - is this something you can take a quick look at?

tgauth:
Thanks for opening this! My initial thoughts are that this issue is applicable to other arguments that can have spaces as well (i.e. regular arguments for paths) and I wonder if we can address those here as well. I'm going to think on it a bit and see if there's a good way to do that, and if not, we can defer it to a future PR.

resources/sshdconfig/src/get.rs
```diff
@@ -133,6 +134,12 @@ pub fn get_sshd_settings(cmd_info: &CommandInfo, is_get: bool) -> Result<Map<Str
         result.insert("match".to_string(), match_value.clone());
     }
 
+    // sshd -T normalizes space-separated list keywords by stripping the quotes that preserve
+    // values containing spaces (e.g. Windows group names like "openssh users"), splitting them
+    // into separate entries. Prefer the value parsed directly from the config file, which retains
+    // the quoting, for these keywords when they are explicitly set.
```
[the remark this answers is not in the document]
tgauth: the config file does have quoting so the comment is correct

resources/sshdconfig/src/get.rs
```diff
@@ -171,3 +178,65 @@ pub fn get_sshd_settings(cmd_info: &CommandInfo, is_get: bool) -> Result<Map<Str
     }
     Ok(result)
 }
+
+fn prefer_explicit_space_sep_lists(result: &mut Map<String, Value>, explicit_settings: &Map<String, Value>) {
+    for keyword in MULTI_ARG_KEYWORDS_SPACE_SEP {
```
tgauth: Ok to cover broader cases, I think there's two changes needed:
1. Expand this check to be _any_ keyword in `explicit_settings` that has spaces, rather than only checking for keywords in `MULTI_ARG_KEYWORDS_SPACE_SEP`
2. When the Rust parser is parsing keywords from `sshd -T`, it should not error on "regular, single-value" keywords with a spaced value (i.e. Banner "C:\Program Files\ssh\sample_banner.txt" is output as Banner c:\program files\ssh\sample_banner.txt which seems like Banner has two args "c:\program" and "file\ssh\sample_banner.txt" so the parser errors currently. We should have the parser combine them into a single value)

@Gijsreyn, if you want to leave the PR as-is and then I'll add these changes after, that's fine with me too!

Gijsreyn: @tgauth - thanks for the feedback! I tried to incorporate it already into this PR. Let me know what you think :)

tgauth (approved):
LGTM - thanks @Gijsreyn!

PR #1657, opened 2026-08-06
(GH-1563) Enhance the PSScript resources

### LINKED ISSUE #1563: `Microsoft.DSC.Transitional/*Script` design
As I was working on the documentation for `Microsoft.DSC.Transitional/PowerShellScript` and `WindowsPowerShellScript`, I noticed a few issues that I wanted to be sure to capture.

1. Enums should emit as strings, not integers - `ConvertTo-Json` should always specify the `-EnumsAsStrings` parameter. Currently, returning any enum value serializes it to its integer representation, which is unlikely to be what the author intended.

1. The `input` property erroneously forbids the input type from being a non-integer number. Defining the instance as:

   ```yaml
   input: 1.2
   getScript: |-
     param($inputData)

     "Input number is: $inputData"
   ```

   Fails the JSON Schema validation. However, the following definition is considered valid:

   ```yaml
   input:
     number: 1.2
   getScript: |-
     param($inputData)

     "Input number is: $($inputData.number)"
   ```

1. We do not have a good way to indicate changes from within the resource. When there are any changes, the `changedProperties` field in the result is always `["output"]`. To review the changes, users must carefully compare `beforeState.output` to `afterState.output`. The only way to clearly indicate changes is to use a `Write-*` cmdlet to emit a message to the user.

1. The `testScript` is only ever used for **Test** operations. For both `dsc resource set` and `dsc config set` commands the resource only invokes the `getScript` and `setScript`. There is no way to avoid actually invoking the `setScript` when the `testScript` would indicate that the resource is in the desired state.

   Although the resource manifest specifies `set.implementsPreTest` as `true`, the resource _doesn't_ actually do this - we would need the script to invoke the `testScript` if it's defined to determine whether to invoke `setScript`.

1. The output for the **Test** operation is unreadable because it automatically includes the full definition for `desiredState` and `actualState` is always `{"_inDesiredState":<true|false>}`. This tells the user almost nothing about the instance and just adds noise to the result, especially when the scripts are longer than a few lines.

   The `differingProperties` array then always shows `["input", "getScript", "setScript", "testScript"]` (assuming all properties are defined). This is technically true but useless noise for the user.

   We should consider a way to more usefully report the difference between desired state and actual state. It also makes me think about whether we should eventually introduce handling for read-only and write-only properties in the result objects, or perhaps a custom keyword like `x-dsc-hideFromDesiredState`.

1. Currently, `output` is always emitted as an array of JSON values. Instead, when `output` is a single item, we should hoist that value out of the array. Then the user could compare `output.foo` to `output.foo` instead of indexing into the array.

   I think we _do_ want to support emitting multiple objects but that we should encourage (and make readable) returning a single object to represent the state of the instance. I think that, most of the time, users are going to check a specific piece of data and report on it, which maps best to emitting a single object.

   Under this proposal, the behavior for the resource itself would be:

   1. If `$outputObjects` is _empty_, do not insert `output` into the return data for the operation.
   1. If `$outputObjects.Length` is `1`, emit that item as the value for `output`.
   1. If `$outputObjects.Length` is greater than `1`, emit the array as the value for `output`.

1. The JSON Schema for the resource needs some cleanup. In particular, the existing properties all need to be marked as `writeOnly` (`input`, `*Script`) and `readOnly` (`output` and `_inDesiredState`).

   The only property that can validly be defined as `null` is `output` (if we hoist the data for scripts that only emit a single item). If you define `input` as `null`, the resource fails.

   When we set `type` to include `null` we are indicating that the actual JSON value `null` is valid for that property. To represent whether the property is mandatory we need to use the `required` keyword. From a validation perspective, _none_ of the properties are _required_, so long as any one of the `*Script` properties is defined. We can represent that in the JSON Schema as (using YAML for readability):

   ```yaml
   type: object
   properties:
     input:           # elided for brevity
     getScript:       # elided for brevity
     testScript:      # elided for brevity
     setScript:       # elided for brevity
     _inDesiredState: # elided for brevity
   anyOf: # if the resource is defined without any `*Script` property it's invalid.
   - required: [getScript]
   - required: [setScript]
   - required: [testScript]
   ```

   We should also consider defining the `contentMediaType` keyword for the script properties as `text/vnd.microsoft.powershell.script` - those properties are expecting PowerShell scripts. While our validation engine can't operate on this keyword, it's still useful to indicate it as part of the schema.

   An updated recommendation for the schema (including previous suggestions) follows:

   ```yaml
   $schema: https://json-schema.org/draft/2020-12/schema
   type: object
   title: Transitional PowerShell script resource
   description: >-
     Defines an instance of the `Microsoft.DSC.Transitional/PowerShellScript`
     resource, which enables users to define inline scripts for the get, set, and
     test resource operations.
   anyOf:
     - required: [getScript]
     - required: [testScript]
     - required: [setScript]
   properties:
     getScript:
       title: Get operation script
       description: >-
         A PowerShell script that retrieves the current state of the instance.
       $ref: "#/$defs/powershellScript"
     setScript:
       title: Set operation script
       description: >-
         A PowerShell script that enforces the desired state for the instance.
       $ref: "#/$defs/powershellScript"
     testScript:
       title: Test operation script
       description: >-
         A PowerShell script that checks whether the instance is in the desired
         state.
       $ref: "#/$defs/powershellScript"
     input:
       title: Script input data
       description: >-
         Defines the data to pass to every script as a parameter. When defined,
         every script property must include a `param()` statement with a single
         parameter.
       writeOnly: true
       type: [string, boolean, integer, number, object, array]
     output:
       title: Script output data
       description: >-
         Defines output emitted to the success stream from the invoked script.
         The resource result for an operation only includes this property when
         the script emits one or more items to the success stream.
       readOnly: true
       type: [string, boolean, integer, number, object, array, 'null']
     _inDesiredState:
       $ref: "https://aka.ms/dsc/schemas/v3/resource/properties/inDesiredState.json"
   $defs:
     powershellScript:
       writeOnly: true
       type: string
       minLength: 1
       contentMediaType: text/vnd.microsoft.powershell
   
     "https://aka.ms/dsc/schemas/v3/resource/properties/inDesiredState.json":
       $schema: https://json-schema.org/draft/2020-12/schema
       $id: https://raw.githubusercontent.com/PowerShell/DSC/main/schemas/v3/resource/properties/inDesiredState.json
       title: Instance is in the Desired State
       description: >-
         Indicates whether the instance is in the desired state.
         This property is only returned by the `test` method.
       type: [boolean, 'null']
       readOnly: true
   ```

michaeltlombardi:
# PR Summary

This change:

1. Canonicalizes the JSON Schema for the transitional PowerShell script resources by

   - Adding the `title` and `description` keywords for every property.
   - Embedding the canonical property definition for `_inDesiredState` and referencing it instead of redefining it.
   - Using a shared subschema for the script properties that ensures they are strings with a minimum length of 1 and indicates that the expected content is PowerShell script code in with the `contentMediaType` keyword.
   - Adding constraints to verify the shape of an instance for input (requiring one or more script property) and output (requiring an empty object, an object with `output`, or an object with `_inDesiredState`).
   - Allowing users to pass non-integer numbers for `input` and forbidding passing `null`

1. Updates the emitted data for `output`. Prior to this change, the `output` property was always an array. This change checks whether the resource emitted any output and:

  - Emits an empty object when the script doesn't emit any data.
  - Emits an object with `output` defined as a scalar value when the script only emitted a single item.
  - Emits an object with `output` defined as an array when the script emitted two or more items.

1. Updates the version of the resource to reflect the changes in schema and behavior.

## PR Context

Addresses issues with the schema and implementation of the PSScript resources and fixes #1563.

### DIFF
--- a/resources/PSScript/psscript.ps1
+++ b/resources/PSScript/psscript.ps1
@@ -189,5 +189,22 @@ if ($Operation -eq 'Test') {
         exit 1
     }
 } else {
-    @{ output = $outputObjects } | ConvertTo-Json -Compress -Depth 10
+    $outputData = @{}
+    if ($outputObjects.Count -eq 1) {
+        $outputData.output = $outputObjects | Select-Object -First 1
+    } elseif ($outputObjects.Count -gt 1) {
+        $outputData.output = $outputObjects
+    }
+
+    $toJsonParams = @{
+        Compress = $true
+        Depth    = 10
+    }
+    # For PowerShell, use `-EnumsAsStrings` to ensure that enum values are serialized as strings instead of integers.
+    # No equivalent option for Windows PowerShell, so this is a limitation of the WindowsPowerShellScript resource.
+    if ($PSVersionTable.PSVersion -ge [Version]'6.0') {
+        $toJsonParams['EnumsAsStrings'] = $true
+    }
+
+    $outputData | ConvertTo-Json @toJsonParams
 }
--- a/resources/PSScript/psscript.ps1
+++ b/resources/PSScript/psscript.ps1
@@ -191,7 +191,7 @@ if ($Operation -eq 'Test') {
 } else {
     $outputData = @{}
     if ($outputObjects.Count -eq 1) {
-        $outputData.output = $outputObjects | Select-Object -First 1
+        $outputData.output = $outputObjects[0]
     } elseif ($outputObjects.Count -gt 1) {
         $outputData.output = $outputObjects
     }
--- a/dsc/tests/dsc_metadata.tests.ps1
+++ b/dsc/tests/dsc_metadata.tests.ps1
@@ -155,7 +155,7 @@ Describe 'metadata tests' {
           - name: return variable
             type: Microsoft.DSC.Transitional/PowerShellScript
             properties:
-              SetScript: |
+              setScript: |
                 if ($IsWindows) {
                   $env:myTestVariable
                 }
@@ -189,7 +189,7 @@ Describe 'metadata tests' {
           - name: return path
             type: Microsoft.DSC.Transitional/PowerShellScript
             properties:
-              SetScript: |
+              setScript: |
                 $env:PATH
 '@
         $oldUserPath = [System.Environment]::GetEnvironmentVariable('PATH', [System.EnvironmentVariableTarget]::User)
@@ -219,7 +219,7 @@ Describe 'metadata tests' {
           - name: return variable
             type: Microsoft.DSC.Transitional/PowerShellScript
             properties:
-              SetScript: |
+              setScript: |
                 if ($IsWindows) {
                   $env:myTestVariable
                 }
--- a/resources/PSScript/psscript.dsc.resource.json
+++ b/resources/PSScript/psscript.dsc.resource.json
@@ -58,6 +58,50 @@
     "schema": {
         "embedded": {
             "type": "object",
+            "oneOf": [
+                {
+                    "description": "Defines constraints for the return data of the resource operations.",
+                    "oneOf": [
+                        {
+                            "description": "Defines the valid return data for `get` and `set` operations with a defined script.",
+                            "required": [
+                                "output"
+                            ]
+                        },
+                        {
+                            "description": "Defines the return data as an empty object for operations with an undefined script.",
+                            "minProperties": 0,
+                            "maxProperties": 0
+                        },
+                        {
+                            "description": "Defines the valid return data for implemented `test` operations.",
+                            "required": [
+                                "_inDesiredState"
+                            ]
+                        }
+                    ]
+                },
+                {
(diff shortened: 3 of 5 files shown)


resources/PSScript/psscript.ps1
```diff
@@ -189,5 +189,22 @@ if ($Operation -eq 'Test') {
         exit 1
     }
 } else {
-    @{ output = $outputObjects } | ConvertTo-Json -Compress -Depth 10
+    $outputData = @{}
+    if ($outputObjects.Count -eq 1) {
+        $outputData.output = $outputObjects | Select-Object -First 1
```
SteveL-MSFT: I think this is slightly better instead of using the pipeline unncessarily

```suggestion
        $outputData.output = $outputObjects[0]
```

SteveL-MSFT (changes requested):

SteveL-MSFT (approved):

PR #1670, opened 2026-08-11
Use JSON Schema defaults in synthetic test get_diff

### LINKED ISSUE #1666: Microsoft.Windows/FirewallRuleList: `unspecifiedRulesAction` is always reported as drift, so `test` and `set` never converge
#### Steps to reproduce

Tested on `dsc 3.3.0-preview.4`, `Microsoft.Windows/FirewallRuleList` v0.2.0, Windows Server 2025, elevated.

Create a rule, then declare it exactly as it exists so there is genuinely no drift:

```powershell
New-NetFirewallRule -Name 'DSC-Repro' -DisplayName 'DSC-Repro' -Direction Inbound `
  -Action Allow -Protocol TCP -LocalPort 32921 -Enabled True -PolicyStore PersistentStore
```

```yaml
# repro.dsc.yaml
resources:
- name: fw
  type: Microsoft.Windows/FirewallRuleList
  properties:
    unspecifiedRulesAction: ignore   # documented as "same as if not specified, does nothing"
    rules:
    - name: DSC-Repro
      direction: Inbound
      action: Allow
      protocol: 6
      localPorts: '32921'
      enabled: true
```

```
dsc config test -f repro.dsc.yaml
```

#### Expected

`inDesiredState: true`. `unspecifiedRulesAction: ignore` is documented as *"same as if not specified, does nothing"*, so it should be indistinguishable from omitting the property.

#### Actual

```json
{ "inDesiredState": false, "differingProperties": ["unspecifiedRulesAction"] }
```

Removing the `unspecifiedRulesAction` line from the same config yields `inDesiredState: true` with `differingProperties: []`. The only variable is the property itself — actual firewall state is identical in both cases.

All three enum values reproduce it:

| config | `inDesiredState` | `differingProperties` |
|---|---|---|
| property omitted | `true` | `[]` |
| `unspecifiedRulesAction: ignore` | `false` | `["unspecifiedRulesAction"]` |
| `unspecifiedRulesAction: disable` | `false` | `["unspecifiedRulesAction"]` |
| `unspecifiedRulesAction: remove` | `false` | `["unspecifiedRulesAction"]` |

#### Cause

`get_rules()` returns `unspecified_rules_action: None` unconditionally, and the field is `skip_serializing_if = "Option::is_none"`, so `dsc resource get` never emits the property:

```json
{"actualState":{"rules":[{"name":"DSC-Repro","protocol":6,"localPorts":"32921", ... }]}}
```

The engine's synthetic test then compares a desired state that has the property against an actual state that lacks it, and records a difference. This matches the intent stated in #1579 that the property be write-only and absent from output state — but combined with synthetic test, "absent from output" produces permanent false drift.

#### Also affects `set` convergence reporting

After a `set` has fully converged, a second identical `set` still reports:

```json
{ "changedProperties": ["unspecifiedRulesAction"] }
```

The underlying rule work is correctly idempotent (already-disabled rules are skipped), but every run reports a change. For fleet reporting this means a converged estate never shows as converged.

#### Suggested fix

Two options, in preference order:

1. **Exclude the property from state comparison.** In #1579, @SteveL-MSFT argued for a canonical `_`-prefixed property (`_disableUnspecified`) partly for schema discoverability. A canonical write-only property would also sidestep this, since the engine already knows not to diff canonical metadata. If a resource-specific name is preferred, the resource or engine needs an equivalent "write-only, do not compare" marker.
2. **Echo the property back from `get`,** defaulting to `ignore` when absent, so desired and actual agree.

Option 1 seems more correct — the property is an instruction, not observable state — but it needs a mechanism that doesn't exist for non-canonical properties today.

#### Impact

Any configuration using `unspecifiedRulesAction` permanently reports out-of-desired-state and reports a change on every run, regardless of actual firewall state. This makes the property unusable for compliance reporting, which is its primary use case.

Refs: #1579, #1599

SteveL-MSFT:
## Motivation

When a DSC resource does not implement the `test` operation, the engine performs a synthetic test by calling `get` and comparing expected vs actual state using `get_diff()`. Previously, if a property existed in the expected (desired) state but was missing from the actual (get) output, it was always reported as differing -- even if the resource simply omits properties whose values match the schema-defined default.

This causes false positives in synthetic test results. The most impactful case is `Microsoft.Windows/FirewallRuleList` where `unspecifiedRulesAction` is a write-only instruction that `get` never returns. Setting it to `"ignore"` (the documented default) permanently reports drift, making the property unusable for compliance reporting.

## Approach

- Introduced `get_diff_with_schema(expected, actual, schema)` which accepts an optional JSON Schema. When a property is in `expected` but absent from `actual`, the function looks up the schema's `default` value for that property. If the expected value matches the default, it is *not* flagged as a diff.
- The existing `get_diff(expected, actual)` remains as a convenience wrapper (passes `None` for the schema), so all non-synthetic-test call sites are unaffected.
- Both synthetic test paths (command resources and adapted resources) now retrieve the resource schema and pass it to the new function.
- Schema is read directly from the `RESOURCE_SCHEMAS` cache (avoiding redundant serialization), with a fallback to `get_schema()` to populate the cache on a miss.

## Test coverage

- **Rust unit tests** (6 new): Cover matching/mismatching schema defaults for booleans and integers, missing defaults, and the no-schema fallback.
- **Pester integration tests** (7 new): End-to-end validation using a new `Test/SchemaDefault` resource.
- **FirewallRuleList Pester tests** (4 new): Verify that `unspecifiedRulesAction: ignore` no longer causes false drift, while non-default values (`disable`, `remove`) are still correctly reported.
- Verified existing `dsc_resource_test.tests.ps1` tests still pass with no regressions.

Fixes #1666

### DIFF
--- a/lib/dsc-lib/src/dscresources/dscresource.rs
+++ b/lib/dsc-lib/src/dscresources/dscresource.rs
@@ -470,7 +470,12 @@ impl Invoke for DscResource {
                             response.actual_state
                         }
                     };
-                    let diff_properties = get_diff( &desired_state, &actual_state);
+                    let schema: Option<Value> = if let Some(s) = &self.schema {
+                        serde_json::to_value(s).ok()
+                    } else {
+                        self.schema().ok().and_then(|s| serde_json::from_str(&s).ok())
+                    };
+                    let diff_properties = get_diff_with_schema( &desired_state, &actual_state, schema.as_ref());
                     desired_state = redact(&desired_state);
                     let test_result = TestResult::Resource(ResourceTestResponse {
                         desired_state,
@@ -647,6 +652,24 @@ pub fn get_adapter_input_kind(adapter: &DscResource) -> Result<AdapterInputKind,
 ///
 /// An array of top level properties that differ, if any
 pub fn get_diff(expected: &Value, actual: &Value) -> Vec<String> {
+    get_diff_with_schema(expected, actual, None)
+}
+
+#[must_use]
+/// Performs a comparison of two JSON Values using an optional JSON Schema.
+/// If a property exists in `expected` but not in `actual`, the schema's `default` value
+/// for that property is used for comparison when available.
+///
+/// # Arguments
+///
+/// * `expected` - The expected value
+/// * `actual` - The actual value
+/// * `schema` - Optional JSON Schema to look up default values for missing properties
+///
+/// # Returns
+///
+/// An array of top level properties that differ, if any
+pub fn get_diff_with_schema(expected: &Value, actual: &Value, schema: Option<&Value>) -> Vec<String> {
     let mut diff_properties: Vec<String> = Vec::new();
     if expected.is_null() {
         return diff_properties;
@@ -702,8 +725,17 @@ pub fn get_diff(expected: &Value, actual: &Value) -> Vec<String> {
                             diff_properties.push(key.to_string());
                         }
                     } else {
-                        info!("{}", t!("dscresources.dscresource.diffKeyMissing", key = key));
-                        diff_properties.push(key.to_string());
+                        // Property not in actual - check schema for a default value
+                        let schema_default = get_schema_default(schema, key);
+                        if let Some(default_value) = schema_default {
+                            if value != &default_value {
+                                info!("{}", t!("dscresources.dscresource.diffKeyMissing", key = key));
+                                diff_properties.push(key.to_string());
+                            }
+                        } else {
+                            info!("{}", t!("dscresources.dscresource.diffKeyMissing", key = key));
+                            diff_properties.push(key.to_string());
+                        }
                     }
                 } else {
                     info!("{}", t!("dscresources.dscresource.diffKeyNotObject", key = key));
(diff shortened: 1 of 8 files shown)


lib/dsc-lib/src/dscresources/dscresource.rs
```diff
@@ -716,6 +748,23 @@ pub fn get_diff(expected: &Value, actual: &Value) -> Vec<String> {
...
+///
+/// # Returns
+///
+/// The default value if found in the schema's properties definition, otherwise None
+fn get_schema_default(schema: Option<&Value>, property_name: &str) -> Option<Value> {
+    let schema = schema?;
+    let properties = schema.get("properties")?.as_object()?;
+    let property_schema = properties.get(property_name)?.as_object()?;
+    property_schema.get("default").cloned()
+}
```
michaeltlombardi, line 765: Nit: Out of scope for this PR, but I think this fits better as an extension method in `dsc-lib-jsonschema` - probably something like the following signature:

```rust
pub get_default_property_value(
  &Self,
  property_name: &str,
) -> Option<&Value> {}
```

lib/dsc-lib/src/dscresources/dscresource.rs
```diff
@@ -702,8 +725,17 @@ pub fn get_diff(expected: &Value, actual: &Value) -> Vec<String> {
                             diff_properties.push(key.to_string());
                         }
                     } else {
-                        info!("{}", t!("dscresources.dscresource.diffKeyMissing", key = key));
-                        diff_properties.push(key.to_string());
+                        // Property not in actual - check schema for a default value
+                        let schema_default = get_schema_default(schema, key);
+                        if let Some(default_value) = schema_default {
```
michaeltlombardi: Nit: I think we can collapse the assignment and the if let statement into a single line:

```suggestion
                        if let Some(default_value) = get_schema_default(schema, key) {
```

michaeltlombardi (approved):
Overall, this PR looks good to me.

Out of scope but related, we should probably leverage this new comparison with schemas in a follow up PR to handle write-only and read-only properties that differ between actual/expected - write-only properties can never be compared and there's an argument for read-only properties only being compared when they're in both states (because the user is specifically saying "I expect `output` to be `foo`").

lib/dsc-lib/src/dscresources/dscresource.rs
SteveL-MSFT: Agree we can improve this outside of this PR

Issue #1391, opened 2026-02-17
MAINT: Refactor `dsc-lib` for consistency, library API usage, and maintainability


michaeltlombardi:
### Summary of the new feature / enhancement

> As a maintainer of the `dsc-lib` crate and integrating developer of the same crate,
> I want to be able to more quickly navigate, contribute, test, document, and reference the types and functions in the library
> So that I can contribute and integrate more quickly and reliably.

The layout and library API for `dsc-lib` has grown organically from the beginning of the project. Along the way, we've learned a lot about writing idiomatic and performant Rust code that differs from the current structure.

Additionally, when we first began work on `dsc-lib`, we worked under the assumption that the only Rust projects that would take a dependency on the library are the `dsc` CLI and a limited set of crates developed in this monorepo. Now that the Bicep local extension depends on `dsc-lib` and we plan to develop a Development Kit in Rust for DSC resources and extensions, we should carefully consider updating the library to follow more consistent conventions for the library APIs and structure.

### Proposed technical implementation details (optional)

There are a few different areas we can pursue as part of this overall effort:

1. Decomposing large rust files into submodules with the re-export pattern.

   We're currently using this pattern to various degrees with the `types` module. We're almost using this pattern for the `functions`, except that the submodules are public, so you reference a function type as `dsc_lib::functions::<snake_case_name>::<PascalCaseName>`. Making this change would enable a simpler access for these types, like `dsc_lib::functions::CreateArray` instead of `dsc_lib::functions::create_array::CreateArray`.

   This would also enable decomposing larger files with numerous types into smaller files that are easier to read, maintain, and reason about.

   > [!NOTE]
   > While out of scope for this issue, it will also make it easier for us to extract subsets of functionality into separate crates to improve compile time performance (for Rust, the _crate_ is a compilation unit). For example, if we decomposed the `dscresources` module into a `dsc-lib-resources` crate and re-exported it from within `dsc-lib`, any compilation for local development that _doesn't_ affect the decomposed crate is able to reuse prior compilation.  I'm not sure the performance benefits are worth this more drastic move, but we might find that to be the case in the future.
2. Reviewing the naming for structs, functions, and other items for consistency and clarity.

   Currently, we have a mix of naming conventions that have organically grown with the project. We should be more consistent in the naming for types, traits, and functions both to ease general cognitive load for maintainers and to provide a consistent model for integrating developers to rely on.
3. Standardize reference documentation for items in the library.

[...]

PR #1672, opened 2026-08-12
(GH-538) Enhance schema export functionality and add error handling for duplicate paths

Gijsreyn:
# PR Summary

The primary purpose of this change is to make cargo xtask schema export reliable and complete enough to publish versioned DSC schemas from source code. It fixes two silent output collisions, exports every type that derives DscRepoSchema, and adds version targeting so the exporter can produce any recognized schema version folder instead of only vNext.

As a maintainer or contributor, you can now invoke:

- `cargo xtask schema export` — exports schemas/vNext as before.
- `cargo xtask schema export --schema-version v3.2 --schema-version v3.2.3` — exports specific version folders.
- `cargo xtask schema export --release 3.3.0` — exports the full set a release must publish: schemas/v3.3.0, schemas/v3.3, and schemas/v3.

## PR Context

As part of #538 and following up on #1406, the exporter needs to produce correct, complete, versioned schemas from source before it can replace the hand-maintained YAML pipeline under schemas/src.


lib/dsc-lib/src/dscresources/adapted_resource_manifest.rs
```diff
@@ -25,7 +25,7 @@ pub enum AdaptedPathOrContent {
 #[serde(deny_unknown_fields, rename_all = "camelCase")]
 #[dsc_repo_schema(
     base_name = "manifest",
-    folder_path = "resource",
+    folder_path = "adaptedresource",
```
michaeltlombardi: Should this more correctly be `resource/adapted`? It's a data object that represents a specific _kind_ of resource manifest, but it's still semantically there - and many of the fields reuse the same definition as the typical resource manifest.


```suggestion
    folder_path = "resource/adapted",
```

lib/dsc-lib/src/dscresources/adapted_resource_manifest.rs
```diff
@@ -36,7 +36,7 @@ pub enum AdaptedPathOrContent {
 pub struct AdaptedDscResourceManifest {
     /// The version of the resource manifest schema.
     #[serde(rename = "$schema")]
-    #[schemars(schema_with = "ResourceManifest::recognized_schema_uris_subschema")]
+    #[schemars(schema_with = "AdaptedDscResourceManifest::recognized_schema_uris_union_subschema")]
```
michaeltlombardi: What is the reason for using this special associated function instead of `recognized_schema_uris_subschema()` on _this_ type (I see that it was previously using the same subschema as the `ResourceManifest`, which was incorrect)?

Gijsreyn: @michaeltlombardi - I looked a bit back in the history, and this function exists to widen the set of `$schema` URIs the adapted-manifest schema accepts (as far as my understanding goes). It takes the type's own subschema and replaces its `enum` with the union of the adapted manifest's URI and the `ResourceManifest` URIs. Using the plain earlier one would restrict the `enum` to only the new `resource/adapted/manifest.json` URIs. And I think that would have invalidated every adapted resource manifest that already exists, because until this branch those documents could only declare a resource manifest schema URI.

michaeltlombardi: Thinking about this for a bit, I'm torn - the prior versions were incorrectly defined and when you defined `$schema` in a given document the validator uses _that_ URI to retrieve and validate the data.

I don't think we should merge the lists, not least because this schema isn't valid for many of the early resource schema definitions.

Gijsreyn: Can't disagree with that. Want me to replace the union with `Self::recognized_schema_uris_subschema()` and delete `recognized_schema_uris_union_subschema`? And then treat the old URI in existing files as a bug to migrate (perhaps worth a `warn!` at load time during a transition period if you want to help people update)?

michaeltlombardi: A few things:

1. I don't see that we actually exported/published the adapted manifest schema into any existing schema folders. We'll need to get the schemas exported, of course, so people can retrieve them without invoking DSC, but we could _correctly_ export them with the bug fixed.
2. In the current implementation (and previous), we're not actually using a schema registry for lookups/validation. It grabs the schema from the type definition in that release and uses that.
3. So this would only apply to adapted resource manifests that are published with the incorrect schema URI and used in future versions of DSC.
4. I'm open to the idea of a JSON schema for the property like:

   ```yaml
   $schema:
     oneOf:
       - <AdaptedDscResourceManifest::recognized_schema_uris_subschema>
       - <ResourceManifest::recognized_schema_uris_subschema>
         deprecated: true
         deprecationMessage: <explanation>
   ```
   
   But there's no _good_ way to raise an info/warning message during schema validation. We would have to special-case this, I think, which is unfortunate.

Gijsreyn: Thanks Mikey, I applied the changes. However, while wiring it up for the first point you mentioned, I found that there's actually a third URI family in play.

1. `.../v3/bundled/resource/manifest.json` — what the buggy-generated schema declared as valid (the deprecated branch we discussed)
2. `.../v3/bundled/resource/adapted/manifest.json` — the new canonical family from your folder_path suggestion
3. `.../v3/bundled/adaptedresource/manifest.json` — the URI from the YAML-authored layout (schemas/src/adaptedresource/), which was never published

The catch was with the last one, and these are the ones shipped currently. It was declared, for example, in `windows_personalization.dsc.adaptedResource.yaml` (also `dsctest` fixtures, and the PowerShell adapter). Now, we can argue if people have copied this, but yeah, most people copy from examples that are shipped. So, should the deprecated `oneOf` branch cover family 3 as well?

michaeltlombardi: I think it should cover that third branch but only the specific URI we shipped those manifests using.

Gijsreyn: Done!

lib/dsc-lib-jsonschema/src/dsc_repo/recognized_schema_version.rs
```diff
@@ -6,4 +6,56 @@
...
 include!(concat!(env!("OUT_DIR"), "/recognized_schema_version.rs"));
+
+/// Defines the error when parsing a string that isn't a recognized schema version folder.
+#[derive(Error, Debug, Clone, PartialEq)]
+#[error(
+    "{t}: {0}. {t2}: {1:?}",
+    t = t!("dsc_repo.recognized_schema_version.unrecognizedVersion"),
+    t2 = t!("dsc_repo.recognized_schema_version.validVersionsAre")
+)]
+pub struct UnrecognizedSchemaVersion(pub String, pub Vec<String>);
```
michaeltlombardi: I missed this before, I think, but we should suffix standalone error types with `Error` - so this should be:


```suggestion
pub struct UnrecognizedSchemaVersionError(pub String, pub Vec<String>);
```

But the update needs to be done in the editor to refactor all references to the type.

lib/dsc-lib-jsonschema/src/dsc_repo/recognized_schema_version.rs
```diff
@@ -6,4 +6,56 @@
...
+        let parsed: RecognizedSchemaVersion = " VNEXT ".parse().unwrap();
+        assert_eq!(parsed, RecognizedSchemaVersion::VNext);
+    }
+
+    #[test]
+    fn from_str_rejects_unrecognized_versions() {
+        assert!("v99.0.0".parse::<RecognizedSchemaVersion>().is_err());
+        assert!("not-a-version".parse::<RecognizedSchemaVersion>().is_err());
+    }
+}
```
michaeltlombardi: Nit: We should prefer to do integration testing in the `lib/dsc-lib-jsonschema/tests/integration` folder - here we're testing a public API.

As I wrote in #1391 about placing public API tests in `tests/integration`:

> 
> The primary motivating factor here is build and test timings. When integration tests are colocated with module code, any changes to either the tests or the implementation requires recompilation because the file changed.
>
> Moving the tests and following the current pattern that mirrors the src directory structure can also help us quickly identify where we are lacking integration tests for the library. Currently, given the existing layout, we need to carefully check every implementation file for test coverage.

This is non-blocking but I wanted to note it for future PRs.

michaeltlombardi (changes requested):
Aside from a non-blocking nit on test placement, I think we should align to the [Rust API Guidelines](https://rust-lang.github.io/api-guidelines/naming.html#names-use-a-consistent-word-order-c-word-order) and ensure our error types have the `Error` suffix for API clarity.

lib/dsc-lib-jsonschema/src/dsc_repo/recognized_schema_version.rs
Gijsreyn: Done!

Gijsreyn:
> Aside from a non-blocking nit on test placement, I think we should align to the [Rust API Guidelines](https://rust-lang.github.io/api-guidelines/naming.html#names-use-a-consistent-word-order-c-word-order) and ensure our error types have the `Error` suffix for API clarity.

Can you take another look if this is alright for now?

michaeltlombardi (approved):

PR #1674, opened 2026-08-13
Ignore write-only properties in schema-aware diffs

### LINKED ISSUE #1668: Enable JSONSchema property decorator to inform synthetic test to skip

Some properties passed to a resource are instructions rather than properties informing state.  In the WindowsFirewallRules resource, the `unspecifiedRule` is an example.  Since it's not canonical, the engine doesn't know anything about it and during synthetic test it will compare the input and output such that when this property is used, it will fail the comparison since the default `ignore` is also unspecified.

Add support for `x-instruction` to JSONSchema which will inform the DSC engine to not compare that particular property.

SteveL-MSFT:
Properties used only as resource instructions may not be returned by a resource, which caused schema-aware comparisons to report false differences.

This change makes `get_diff_with_schema()` skip properties marked `writeOnly: true` in the resource schema. The Windows Firewall resource now marks `unspecifiedRulesAction` as write-only, and unit and resource tests cover differing, omitted, and explicitly non-write-only properties.

Fix https://github.com/PowerShell/DSC/issues/1668

### DIFF
--- a/lib/dsc-lib/src/dscresources/dscresource.rs
+++ b/lib/dsc-lib/src/dscresources/dscresource.rs
@@ -11,7 +11,7 @@ use rust_i18n::t;
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
 use serde_json::{Map, Value};
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 use std::path::PathBuf;
 use tracing::{debug, info, trace, warn};
 
@@ -769,16 +769,40 @@ fn get_schema_default(schema: Option<&Value>, property_name: &str) -> Option<Val
     property_schema.get("default").cloned()
 }
 
-/// Returns whether a property's JSON Schema sets `writeOnly` to `true`.
+/// Returns whether a property's JSON Schema sets `writeOnly` to `true`, directly or
+/// through a local JSON Pointer reference.
 fn is_schema_write_only(schema: Option<&Value>, property_name: &str) -> bool {
-    schema
-        .and_then(|schema| schema.get("properties"))
+    let Some(schema) = schema else {
+        return false;
+    };
+    let Some(mut property_schema) = schema
+        .get("properties")
         .and_then(Value::as_object)
         .and_then(|properties| properties.get(property_name))
-        .and_then(Value::as_object)
-        .and_then(|property_schema| property_schema.get("writeOnly"))
-        .and_then(Value::as_bool)
-        .unwrap_or(false)
+    else {
+        return false;
+    };
+    let mut visited_references = HashSet::new();
+
+    loop {
+        if property_schema.get("writeOnly").and_then(Value::as_bool) == Some(true) {
+            return true;
+        }
+
+        let Some(reference) = property_schema.get("$ref").and_then(Value::as_str) else {
+            return false;
+        };
+        let Some(pointer) = reference.strip_prefix('#') else {
+            return false;
+        };
+        if !visited_references.insert(pointer) {
+            return false;
+        }
+        let Some(resolved_schema) = schema.pointer(pointer) else {
+            return false;
+        };
+        property_schema = resolved_schema;
+    }
 }
 
 /// Validates the properties of a resource against its schema.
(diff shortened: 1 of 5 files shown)


lib/dsc-lib/src/dscresources/dscresource.rs
```diff
@@ -764,6 +769,18 @@ fn get_schema_default(schema: Option<&Value>, property_name: &str) -> Option<Val
...
+fn is_schema_write_only(schema: Option<&Value>, property_name: &str) -> bool {
+    schema
+        .and_then(|schema| schema.get("properties"))
+        .and_then(Value::as_object)
+        .and_then(|properties| properties.get(property_name))
+        .and_then(Value::as_object)
+        .and_then(|property_schema| property_schema.get("writeOnly"))
+        .and_then(Value::as_bool)
+        .unwrap_or(false)
+}
```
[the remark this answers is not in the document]
SteveL-MSFT: Updated `is_schema_write_only()` to follow local JSON Pointer `$ref` values, including chained references with cycle detection. Added coverage matching the `$defs` pattern used by PSScript resources.

lib/dsc-lib/src/dscresources/dscresource.rs
```diff
@@ -764,6 +769,42 @@ fn get_schema_default(schema: Option<&Value>, property_name: &str) -> Option<Val
     property_schema.get("default").cloned()
 }
 
+/// Returns whether a property's JSON Schema sets `writeOnly` to `true`, directly or
+/// through a local JSON Pointer reference.
+fn is_schema_write_only(schema: Option<&Value>, property_name: &str) -> bool {
```
michaeltlombardi, line 774: Initially, I was going to recommend implementing this would be easier in the `dsc-lib-jsonschema` crate, but thinking about this a bit more, I think the ergonomics and simplicity will mostly improve when we implement the in-memory schema registry and retriever, when we can call [`deference()`](https://docs.rs/jsonschema/latest/jsonschema/fn.dereference.html) to get the full property definition for verification.

lib/dsc-lib/src/dscresources/dscresource.rs
```diff
@@ -764,6 +769,42 @@ fn get_schema_default(schema: Option<&Value>, property_name: &str) -> Option<Val
...
+        if property_schema.get("writeOnly").and_then(Value::as_bool) == Some(true) {
+            return true;
+        }
+
+        let Some(reference) = property_schema.get("$ref").and_then(Value::as_str) else {
+            return false;
+        };
+        let Some(pointer) = reference.strip_prefix('#') else {
+            return false;
+        };
```
michaeltlombardi, line 797: This works for pointer references like `#/$defs/foo` but not URI references (site-relative or absolute), like `https://schemas.contoso.com/foo` or `/foo`.

The [formalized bundling format for JSON Schema 2020-12](https://json-schema.org/blog/posts/bundling-json-schema-compound-documents) (known more correctly as a _compound schema document_) is to ensure that you include any external schemas in the `$defs` keyword _with_ their `$id` keyword. 

For example, the following non-bundled schema:

```json
{
  "$id": "https://jsonschema.dev/schemas/examples/non-negative-integer",
  "description": "Must be a non-negative integer",
  "$ref": "#/$defs/nonNegativeInteger"
  "$defs": {
    "nonNegativeInteger": {
      "allOf": [
        { "$ref": "/schemas/mixins/integer" },
        { "$ref": "/schemas/mixins/non-negative" }
      ]
    }
  },
}
```

Bundles to the following compound schema document:

```json
{
  "$id": "https://jsonschema.dev/schemas/examples/non-negative-integer-bundle",
  "description": "Must be a non-negative integer",
  "$ref": "#/$defs/nonNegativeInteger"
  "$defs": {
    "nonNegativeInteger": {
      "allOf": [
        { "$ref": "/schemas/mixins/integer" },
        { "$ref": "/schemas/mixins/non-negative" }
      ]
    },
    "https://jsonschema.dev/schemas/mixins/integer": {
      "$id": "https://jsonschema.dev/schemas/mixins/integer",
      "description": "Must be an integer",
      "type": "integer"
    },
    "https://jsonschema.dev/schemas/mixins/non-negative": {
      "$id": "https://jsonschema.dev/schemas/mixins/non-negative",
      "description": "Not allowed to be negative",
      "minimum": 0
    }
  },
}
```

When a validator resolves a relative URI like `/schemas/mixins/integer`, it does so _relative to the schemas `$id` URI.

The _key_ in the `$defs` for bundled schema resources doesn't matter, using the absolute URI is a convention that ensures unique keys for each resource. What the resolver does is look for a subschema in `$defs` that defines an `$id` that matches the reference.

We have an extension method on the `schemars::Schema` type for checking whether a reference is to a bundled schema resource that handles both pointers and absolute/relative URI references:

https://github.com/PowerShell/DSC/blob/88ca70645fc2f16d020f0e554b5b11bf21047cb9/lib/dsc-lib-jsonschema/src/schema_utility_extensions.rs#L1619-L1666

SteveL-MSFT: One of the limitations we have is we CAN'T do a HTTP call to retrieve URLs due to compliance restrictions

SteveL-MSFT:
@copilot resolve the merge conflicts in this pull request

lib/dsc-lib/src/dscresources/dscresource.rs
michaeltlombardi: After verbal discussion, I think we can defer resolving canonically bundled resources to a future PR. For now, being able to resolve URI identifier fragments like `#/$defs` is fine.

michaeltlombardi (approved):

PR #1675, opened 2026-08-13
Add Windows environment variable resource

### LINKED ISSUE #419: Resource to manage Windows env vars

The PSDSCResource for `environment` has limitations that may be better suited for a rewrite as a v3 Windows only resource.  Specially handle User vs System env vars (only persisted kind, not the current process) and a new model to handle `PATH` type variables treating them as arrays.

_No response_

SteveL-MSFT:
Windows DSC needs a native resource for managing persistent user and machine environment variables without requiring callers to manipulate registry paths and value formatting directly.

This adds `Microsoft.Windows/EnvironmentVariable` and `Microsoft.Windows/EnvironmentVariableList`, a Windows-only Rust resource with an `environmentVariables` list schema. It supports scalar values, path arrays with prepend/append/clobber behavior, case-insensitive path deduplication, scope and existence defaults, and deletion through `_exist: false`. Values are stored in the appropriate CurrentUser or AllUsers registry location, preserving expandable-string behavior when needed.

AllUsers reads remain available without elevation, while AllUsers writes and removals fail before any mutation with a dedicated actionable elevation error. CurrentUser operations therefore do not unnecessarily require an elevated security context.

Pester coverage exercises get and set behavior, list ordering, defaults, all path actions, deduplication, deletion, invalid input, and the non-elevated AllUsers error path.

Fix https://github.com/PowerShell/DSC/issues/419


resources/environment_variable/tests/environment_variable_get.tests.ps1
```diff
@@ -0,0 +1,93 @@
...
+            $testValue,
+            [EnvironmentVariableTarget]::User)
+    }
+
+    AfterAll {
+        [Environment]::SetEnvironmentVariable(
+            $testName,
+            $null,
+            [EnvironmentVariableTarget]::User)
+    }
```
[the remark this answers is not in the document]
SteveL-MSFT: Test creates a guid based variable, so it wouldn't exist prior

resources/environment_variable/tests/environment_variable_set.tests.ps1
```diff
@@ -0,0 +1,212 @@
...
+    }
+
+    AfterEach {
+        foreach ($name in $testNames) {
+            [Environment]::SetEnvironmentVariable(
+                $name,
+                $null,
+                [EnvironmentVariableTarget]::User)
+        }
+    }
```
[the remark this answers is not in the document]
SteveL-MSFT: The env var is guid based so should not exist prior

resources/environment_variable/environment_variable.dsc.resource.json
```diff
@@ -0,0 +1,124 @@
...
+                                "title": "Path action",
+                                "description": "How pathValue entries are combined with the current value.",
+                                "writeOnly": true,
+                                "default": "clobber",
+                                "enum": [
+                                    "prepend",
+                                    "append",
+                                    "clobber"
+                                ]
+                            },
```
michaeltlombardi: I think this is more properly extracted into a [`dependentSchemas`](https://json-schema.org/understanding-json-schema/reference/conditionals#dependentSchemas) keyword since it's only valid when `pathValue` is defined and should never be specified with `value`.

Relatedly, we could use the [`dependentRequired`](https://json-schema.org/understanding-json-schema/reference/conditionals#dependentRequired) keyword to mark an instance as invalid when it defines `pathAction` without `pathValue`.

resources/environment_variable/src/types.rs
```diff
@@ -0,0 +1,169 @@
...
+    Get,
+    Set,
+}
+
+#[derive(Debug, Clone, Copy, Default, PartialEq, Eq, Hash, Serialize, Deserialize)]
+pub enum Scope {
+    AllUsers,
+    #[default]
+    CurrentUser,
+}
```
michaeltlombardi, line 21: Is there a reason _not_ to camel case these enum values to match existing conventions for DSC and resources generally? I know that PascalCase is what the Windows API emits, but this does break from how we name properties and enum values in this and other resources.

michaeltlombardi (commented):
A few thoughts on the design and implementation:

1. I think we should probably add some tests that cover the `test` operation because it's not obvious to me how we report changes and out-of-state details for path values to the user. Consider the following scenarios:

   - > The variable defines `pathValue` with two paths and `pathAction` as `prepend`. One of the values is already defined but the other would be inserted. The final value includes one or more segments not defined in the input.

     We don't seem to emit any messages indicating which path segments are missing or will be inserted. Comparing the actual or final path value to the input requires manually verifying each segment.

   - > The variable defines `pathValue` with two paths and `pathAction` as `clobber`. Both values are already defined but the actual value includes an empty segment.

     Does the resource report that the instance is currently out-of-state since it will need to remove the duplicate when canonicalizing the value? When the user invokes the set operation, how does the user know that the empty segment was removed?

1. Not required for this PR, but we may want to make the canonicalizing behavior more explicit and opt-out with a write-only property like `canonicalize` which defaults to `true`.

   This would make it easier to use the resource for setting up test environments, since specifying path values with duplicates and empty segments is technically valid though discouraged.

1. While it doesn't have to be in this PR, I think we should always try to ensure that whenever we define a `*List` resource we also define a single-instance resource when it's coherent to do so.

   Needing to wrap a single-variable definition in an array and losing the reporting granularity/readability is frustrating as an end user.

   In the current engine, this resource can only ever indicate that the `environmentVariables` property is out-of-state. The user then needs to compare every entry in desired/actual to see which environment variable is actually in an invalid state and _how_ it's invalid.

1. Also in the current implementation, we use the `items` keyword to define each environment variable as an object in an array. There's no way to represent the constraint "every item in this array must be unique by the combination of name and scope property values" - though that is handled in the code in the validation step.

   We could address this by altering the design to using the `propertyNames` and `unevaluatedProperties` keywords:

   <details open><summary>Alternate schema</summary>

   ```yaml
   type: object
   propertyNames:
     description: >-
       Every property name must be a non-empty string, optionally prefixed
       with a scope (allUsers\ or currentUser\). When no prefix is provided,
       the property name is assumed to be scoped to the current user.
     pattern: '^(?:(?<scope>(allUsers|currentUser)\\))?\S+$'
   # Unevaluated properties (in this case, every property) uses the provided
   # schema to validate the property value. We could use additionalProperties, but
   # that precludes extending the schema.
   unevaluatedProperties:
     type: object
     properties:
       value:
         title: Value
         description: >-
           The value for the environment variable.
         type: string
       pathValue:
         title: Path Value
         description: >-
           The semicolon-delimited value for the environment variable
           represented as path entries.
         type: string
         minLength: 1
       _exist:
         $ref: https://raw.githubusercontent.com/PowerShell/DSC/main/schemas/v3/resource/properties/exist.json
     # pathAction is only valid when pathValue is specified
     dependentSchemas:
       pathValue:
         properties:
           pathAction:
             title: Path Action
             description: >-
               How pathValue entries are combined with the current value of the
               environment variable. If not specified, the default is to clobber
               the current value with the new value.
             writeOnly: true
             default: clobber
             type: string
             enum:
               - prepend
               - append
               - clobber
     # Always nest constraints in an allOf to support adding new constraints
     # and giving readers a way to quickly review them.
     allOf:
       - not: { required: [value, pathValue] }
         description: >-
           Either value or pathValue must be specified, but not both.
   # Bundle the canonical property definition in $defs, don't redefine it inline.
   $defs:
     "https://raw.githubusercontent.com/PowerShell/DSC/main/schemas/v3/resource/properties/exist.json":
       $schema: https://json-schema.org/draft/2020-12/schema
       $id: https://raw.githubusercontent.com/PowerShell/DSC/main/schemas/v3/resource/properties/inDesiredState.json
       title: Instance is in the Desired State
       description: Indicates whether the instance is in the desired state. This property is only returned by the `test` method.
       type:
         - boolean
         - "null"
       readOnly: true
   ```

   </details>

   Where the design would then be to parse each property name for the variable scope/name. This would be a significant design of the resource implementation but compare the effective definition and result data:

   <details open><summary>Synthetic test result for current design</summary>

   ```yaml
   desiredState:
     environmentVariables:
       - name: foo
         scope: AllUsers
         value: bar
       - name: MY_PATH
         scope: CurrentUser
         pathValue:
           - D:\infra\dsc\resources
           - D:\infra\dsc\extensions
         pathAction: prepend
   actualState:
     environmentVariables:
       - name: foo
         scope: AllUsers
         value: baz
       - name: MY_PATH
         scope: CurrentUser
         pathValue:
           - D:\infra\utils\bin
   inDesiredState: false
   differingProperties:
     - environmentVariables
   ```

   </details>

   <details open><summary>Synthetic test result for proposed design</summary>

   ```yaml
   desiredState:
     AllUsers\foo:
       value: bar
     CurrentUser\MY_PATH:
       pathValue:
         - D:\infra\dsc\resources
         - D:\infra\dsc\extensions
   actualState:
     AllUsers\foo:
       value: baz
     CurrentUser\MY_PATH:
       pathValue:
         - D:\infra\utils\bin
   inDesiredState: false
   differingProperties:
     - AllUsers\foo
     - CurrentUser\MY_PATH
   ```

   </details>

   I bring this up because the ergonomics (particularly for test/set) of list resources is currently very painful and worse the more entries a user defines for that resource.

   I think migrating towards a `*Set` pattern would make it much easier to reason about these convenience resources and review their results.

resources/environment_variable/src/types.rs
SteveL-MSFT: Should be changed to camelCase

resources/environment_variable/environment_variable.dsc.resource.json
SteveL-MSFT: will update the schema

SteveL-MSFT:
It makes sense for this resource to implement `test` explicitly and it's likely a pattern that's going to be true for many `writeOnly` resources where the property is an instruction and not a state

resources/environment_variable/src/types.rs
SteveL-MSFT: Changed the serialized scope values and schema enum/default to camelCase (`allUsers` and `currentUser`) and updated the Pester expectations.

resources/environment_variable/environment_variable.dsc.resource.json
SteveL-MSFT: Updated the embedded schema to draft 2020-12 and now use `dependentSchemas` to make `value` and `pathValue` mutually exclusive plus `dependentRequired` to require `pathValue` when `pathAction` is present.

Issue #1539, opened 2026-05-21
Design Document: Built-in resources


michaeltlombardi:
We need to define some design principles for built-in resources so we can ensure that the resources that ship with DSC follow best practices and provide first-class ergonomics to users. The built-in resources serve both as examples for resource design and will be among the most-used resources.

During the review for the 3.2 GA, I noticed a few specific design principles we should call out. The following list is non-exhaustive:

1. Only define the `_exist` canonical property at the _top-level_ for a resource instance JSON Schema and only when the resource can both create and delete instances. The resource _must_ implement the canonical property in adherence to the contract. Don't define `_exist` for nested instances of a `*List` resource, since those instances can't participate in the engine semantics for existence.
1. When a resource has the `_exist` canonical property it should implement the `get`, `set`, `export`, and `delete` operations.
1. Always implement what-if support for `set` and `delete` operations.
1. Always return all non-write-only properties for the actual state of a resource if those property values are knowable. If the values are _not_ knowable for a given operation, _omit_ those properties from the return data. In JSON Schema defining a property as `null` and omitting the property are _semantically distinct_ representations of data. Only populate a field with `null` when that actually matches the underlying resource behavior (when the resource has a _nullable_ property, rather than an unset or undiscoverable property value).
1. Always mark properties that can be passed to but not returned by the resource with the `writeOnly` keyword.
1. Always mark properties that can be returned by but not passed to the resource with the `readOnly` keyword.
1. Never add `null` as a valid value for a property in the instance JSON Schema when `null` indicates that you aren't passing a value for the property. Only allow `null` for properties where `null` is a _valid value_ for that property.
1. When creating a `*List` resource to manage multiple instances, also be sure to create a companion resource that manages a single instance. We should direct users towards the `*List` resource options only when the performance impact for single instances is too high. Those resources are much less ergonomic for _result output_ and are somewhat obfuscated in the node graph. Using those resources necessitates tradeoffs in ergonomics for authoring vs reporting and performance vs clarity.

Issue #567, opened 2024-10-05
`File` resource as part of base package


SteveL-MSFT:
### Summary of the new feature / enhancement

Need a cross-platform `file` resource.  Probably base it initially on the built-in `file` resource that is part of DSCv1.

michaeltlombardi:
What properties are we expecting of the initial release of the `File` resource?

For reference and prior art, see:

- The [Puppet `file` resource](https://www.puppet.com/docs/puppet/7/types/file.html)
- The [Chef `file` resource](https://docs.chef.io/resources/file/)
- The [Ansible `file` module](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/file_module.html)
- The [Salt `file.managed` state](https://docs.saltproject.io/en/latest/ref/states/all/salt.states.file.html)

One thing to note is that while you _could_ use a single resource to control file existence, contents, permissions, and link type for Windows and not-Windows, everything but existence and contents gets messy quickly when trying to represent things with a single resource.

Puppet handles this with providers and converting unix-isms to Windows, advising users to add a separate ACL resource to manage permissions more idiomatically for Windows if needed. Chef uses similar behavior, having users define the `file` resource and apply cumulative `rights` to it for Windows. Ansible has separate modules for Windows files and related resources (template files, copying, etc). While the salt state works on Windows, any unix file system specific options are ignored or not supported on Windows, and Salt has a separate `win_file` for managing permissions.

I would recommend either following the Puppet/Chef model (assume Unix, translate unix options to Windows in a standardized, documented way, and advise using a separate ACL resource for more control) or the Ansible model (one resource for unix files and another for Windows, fully separated idiosyncracies).

The benefit of the one file resource + Windows permissions is that you're always using a single resource type for files regardless of OS, though you may conditionally need to add ACLs for Windows (having a group depend on the OSInfo assertion). On the other hand, it's more complex to implement (not necessarily maintain, as the APIs for files aren't exactly high-churn for changes and features).

The benefit of two resources is primarily simplicity of implementation and declaration, though they can make manifests fork a bit when managing Windows/not-Windows simultaneously. On the other hand, for _servers_, I don't think that's as common an issue, as servers tend to be more focused in their configuration needs than client devices.

theJasonHelmick:
@theJasonHelmick -- To add spec.

ThomasNieto:
[...]

PR #1676, opened 2026-08-13
Add Microsoft.Filesystem.File/Content resource

SteveL-MSFT:
Adds a cross-platform DSC resource for managing UTF-8 file content and verifying file integrity while consistently reporting the observed content and hashes.

## Summary

- Implements `Microsoft.Filesystem.File/Content` in Rust with get, set, explicit test, and export operations.
- Returns `content`, SHA-256, and SHA-512 from every operation when the file exists.
- Reads each file into a single UTF-8 snapshot so returned content and hashes describe the same bytes.
- Compares content-only test input by hashing the desired UTF-8 content.
- Supports file removal through `_exist: false`.
- Adds the executable and resource manifest to Linux, macOS, and Windows package payloads.

## Testing

- Rust formatting and pedantic Clippy checks
- Rust package build and tests
- 11 cross-platform Pester tests covering get, set, test, export, CLI input handling, and deletion


resources/filecontent/locales/en-us.toml
```diff
@@ -0,0 +1,25 @@
+_version = 1
+
+[main]
+missingOperation = "Missing operation. Usage: filecontent <get|set|test|export> --input <json>"
+unknownOperation = "Unknown operation: '%{operation}'. Expected: get, set, test, or export"
+missingInput = "Missing --input argument"
+invalidJson = "Invalid JSON input: %{error}"
+serializeError = "Failed to serialize output: %{error}"
```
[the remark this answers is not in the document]
SteveL-MSFT: Added the localized `main.missingInputValue` string and wired it into the missing-value path.

resources/filecontent/src/file.rs
```diff
@@ -0,0 +1,233 @@
...
+    }
+
+    let bytes = fs::read(&input.path).map_err(|error| {
+        t!(
+            "export.readError",
+            path = input.path.as_str(),
+            error = error.to_string()
+        )
+        .to_string()
+    })?;
```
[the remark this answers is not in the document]
SteveL-MSFT: Refactored export to read the file once, derive both content and hashes from that byte snapshot, and return `_exist: false` when the read reports NotFound.

resources/filecontent/src/main.rs
```diff
@@ -0,0 +1,86 @@
...
+fn parse_input_arg(args: &[String]) -> Option<&str> {
+    let mut index = 0;
+    while index < args.len() {
+        if args[index] == "--input" {
+            return args.get(index + 1).map(String::as_str);
+        }
+        index += 1;
+    }
+    None
+}
```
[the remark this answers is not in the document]
SteveL-MSFT: Updated `parse_input_arg` to accept both `--input` and `-i`, and to report a dedicated error when the flag has no value. Added direct CLI Pester coverage for both cases.

kilasuit:
My only comment is could this be extended & support any other encoding but default to UTF8

SteveL-MSFT:
> My only comment is could this be extended & support any other encoding but default to UTF8

I considered exposing an `encoding` property, but the JSON/YAML itself needs to be in utf-8 encoding as that is what DSC supports.  I think the majority of the world has decided on UTF8.  I think if there's a need for other type of encoding, I'm planning on a `FileCopy` resource to follow-up that could be used instead.

Thinking about the encoding further, the JSON/YAML and therefore `content` would need to be UTF-8, but when written to disk, it could be a different encoding.  That can still be added later if needed.

ThomasNieto:
The resource should be called `Microsoft/File` not `FileContent`. Reason being is that it handles file existence.

SteveL-MSFT:
> The resource should be called `Microsoft/File` not `FileContent`. Reason being is that it handles file existence.

I named it `FileContent` because a separate `FileCopy` resource will be coming to differentiate.  `FileCopy` will also support `_exist`.  Unlike the old DSC File resource, I didn't want to encapsulate multiple capabilities into one resource (like ParameterSets).

ThomasNieto:
Why don't we want to have multiple different usages in a single resource? Json schema gives us that capability. I'm not a fan of having multiple resources managing the same underlying item.

SteveL-MSFT:
@ThomasNieto Let's discuss at next WG, I'm of the opinion it's confusing to users similar to having multiple parametersets.  Note that multiple resources can be handled by the same exe.

resources/filecontent/filecontent.dsc.resource.json
```diff
@@ -0,0 +1,111 @@
...
+            "type": "object",
+            "required": [
+                "path"
+            ],
+            "properties": {
+                "content": {
+                    "type": "string",
+                    "title": "Content",
+                    "description": "The UTF-8 content of the file.",
+                    "writeOnly": true
```
michaeltlombardi, line 76: I think we should remove the `writeOnly` keyword since this property _is_ returnable by the resource.

```suggestion
                    "description": "The UTF-8 content of the file."
```

resources/filecontent/filecontent.dsc.resource.json
```diff
@@ -0,0 +1,111 @@
...
+                    "title": "SHA-256",
+                    "description": "The SHA-256 hash of the file content.",
+                    "pattern": "^[A-Fa-f0-9]{64}$"
+                },
+                "sha512": {
+                    "type": "string",
+                    "title": "SHA-512",
+                    "description": "The SHA-512 hash of the file content.",
+                    "pattern": "^[A-Fa-f0-9]{128}$"
+                },
```
michaeltlombardi, line 89: I think both of these properties should be marked as read-only, per our [guidance for resource properties](https://learn.microsoft.com/powershell/dsc/concepts/resources/properties#read-only-resource-properties):

> Resources can define read-only properties to describe information about an instance that the resource can retrieve but not directly set. For example, file APIs don't generally allow a user to set the property describing the last time the file was modified.
>
> Generally, you shouldn't include read-only properties when defining the desired state for an instance. Assertion resources that don't support the Set operation can include read-only properties you can use for validating system state for conditional behavior.

```suggestion
                "sha256": {
                    "type": "string",
                    "title": "SHA-256",
                    "description": "The SHA-256 hash of the file content.",
                    "pattern": "^[A-Fa-f0-9]{64}$",
                    "readOnly": true,
                },
                "sha512": {
                    "type": "string",
                    "title": "SHA-512",
                    "description": "The SHA-512 hash of the file content.",
                    "pattern": "^[A-Fa-f0-9]{128}$",
                    "readOnly": true
                },
```

resources/filecontent/filecontent.dsc.resource.json
```diff
@@ -0,0 +1,111 @@
...
+                    "title": "Exists",
+                    "description": "Indicates whether the file should exist. Set to false to remove the file."
+                },
+                "_inDesiredState": {
+                    "type": "boolean",
+                    "title": "In desired state",
+                    "description": "Indicates whether the file is in the desired state. Only returned by test.",
+                    "readOnly": true
+                }
+            },
```
michaeltlombardi, line 107: Per our documentation on canonical resources, both of these property definitions should reference the canonical property schema and bundle it into `$defs`:


```suggestion
                "_exist": {
                    "$ref": "https://raw.githubusercontent.com/PowerShell/DSC/main/schemas/v3/resource/properties/exist.json",
                },
                "_inDesiredState": {
                    "$ref": "https://raw.githubusercontent.com/PowerShell/DSC/main/schemas/v3/resource/properties/inDesiredState.json"
                }
            },
            "$defs": {
                "https://raw.githubusercontent.com/PowerShell/DSC/main/schemas/v3/resource/properties/exist.json": {
                    "$schema": "https://json-schema.org/draft/2020-12/schema",
                    "$id": "https://raw.githubusercontent.com/PowerShell/DSC/main/schemas/v3/resource/properties/exist.json",
                    "title": "Instance should exist",
                    "description": "Indicates whether the DSC resource instance should exist.",
                    "type": "boolean",
                    "default": true,
                    "enum": [
                        false,
                        true
                    ]
                },
                "https://raw.githubusercontent.com/PowerShell/DSC/main/schemas/v3/resource/properties/inDesiredState.json": {
                    "$schema": "https://json-schema.org/draft/2020-12/schema",
                    "$id": "https://raw.githubusercontent.com/PowerShell/DSC/main/schemas/v3/resource/properties/inDesiredState.json",
                    "title": "Instance is in the Desired State",
                    "description": "Indicates whether the instance is in the desired state. This property is only returned by the `test` method.",
                    "type": [
                        "boolean",
                        "null"
                    ],
                    "readOnly": true
                }
            }
```

> [!NOTE]
> We have to bundle the canonical property schemas for now because we don't yet use a schema registry with an in-memory retriever to be able to lookup the property references. Eventually resource authors will be able to just define the property with the `$ref` keyword and DSC will correctly validate the instance.

resources/filecontent/src/types.rs
```diff
@@ -0,0 +1,43 @@
...
+    pub path: String,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub content: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub sha256: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub sha512: Option<String>,
+    #[serde(rename = "_exist")]
+    pub exist: bool,
+}
```
michaeltlombardi, line 43: As a user, I think I would be surprised to invoke the get operation and _not_ get the file content:

```powershell
dsc resource get -r Microsoft/FileContent -i '{"path":"example.txt"}'
```
```yaml
path: example.txt
_exist: true
sha256: <sha-256>
sha512: <sha-512>
```

In general, without specifying a write-only property to control the behavior, resources should always return the full representation of the resource state for an instance in all operations.

To support minimizing the return data, I think we should use a write-only property like `omitContent` to opt into this behavior.

Otherwise, the only way for a user to effectively query for the current content of a file with the resource is to use the `export` operation.

I raised this point in #1539 but we haven't written that document to guide development for built-in resources yet.

michaeltlombardi (changes requested):

michaeltlombardi:
Agree on discussion in WG, we should review in context of https://github.com/PowerShell/DSC/issues/567#issuecomment-2397364274, where I made some recommendations and linked to prior art on designing a resource for this purpose.

resources/filecontent/src/types.rs
SteveL-MSFT: Let's discuss this in the WG, current design is only return content on `export` particularly since it can be large and the hash is an easier way to detect a diff for the user.

SteveL-MSFT:
WG discussed this and prefers to have Content returned in all cases

resources/filecontent/filecontent.dsc.resource.json
```diff
@@ -0,0 +1,111 @@
+{
+    "$schema": "https://aka.ms/dsc/schemas/v3/bundled/resource/manifest.json",
+    "type": "Microsoft/FileContent",
```
SteveL-MSFT: Change to `Microsoft.Filesystem.File/Content`

resources/filecontent/src/types.rs
SteveL-MSFT: Based on WG discussion, `get` will return the content

resources/filecontent/src/types.rs
```diff
@@ -0,0 +1,45 @@
...
```
michaeltlombardi: Nit: Do we still need separate structs for both? Even though we're using the same schema for all operations? 

It _looks like_ we could use a single struct definition - `FileContent` and `ExportState` look nearly identical and the only difference between their fields and `FileState` is the optional `in_desired_state` field that isn't serialized when it's `None`.

resources/filecontent/filecontent.dsc.resource.json
```diff
@@ -0,0 +1,110 @@
...
+    "export": {
+        "executable": "filecontent",
+        "args": [
+            "export",
+            {
+                "jsonInputArg": "--input",
+                "mandatory": true
+            }
+        ],
+        "supportsFiltering": true
```
michaeltlombardi: It's a little unclear to me how filtering is supported. 

Looking through the implementation, I don't think `path` can be a filtering property (export calls `read_state` which seems to operate on a single file path). Similarly, I see no filtering for the other properties (though arguably only content makes sense in this context).

I think this should probably be `false` and we can circle back in a future PR to add support for exporting where:

1. We support path globs for `path`.
2. We support a regular expression for `content`.
3. We forbid, ignore, or do literal matching for the remaining properties.

resources/filecontent/src/types.rs
SteveL-MSFT: This was a carryover when only export returned content, will use single struct

resources/filecontent/filecontent.dsc.resource.json
SteveL-MSFT: You are right, this should be `false`

resources/filecontent/src/types.rs
SteveL-MSFT: Consolidated `FileContent`, `FileState`, and `ExportState` into the single schema-shaped `FileContent` type. The optional `_inDesiredState` field is omitted outside test output.

resources/filecontent/filecontent.dsc.resource.json
SteveL-MSFT: I tested this change with `supportsFiltering: false`, but DSC then does not forward the input to the export executable. Since this resource requires `path` to identify the file to export, the operation receives an empty input and fails before it can read the file. I retained `supportsFiltering: true` so `path` acts as the required export selector; all export tests pass with that setting.

SteveL-MSFT: Corrected as requested: `supportsFiltering` is now `false`. The export test invokes the executable directly with its explicit path input rather than relying on DSC to forward filter input.

michaeltlombardi (approved):

PR #1680, opened 2026-08-18
Fix nested diff w/ schema and schema cache miss

SteveL-MSFT:
# PR Summary

- Enhance when there is a schema cache miss
- Fix synthetic test diff when value is an object to also consider `writeOnly` properties
- Fix test schema to be consistent with the embedded schema


michaeltlombardi (commented):
This looks pretty coherent to me. My only request would be a test to validate that it behaves as expected, particularly for the nested property with a reference.

michaeltlombardi (approved):

PR #1692, opened 2026-08-25
Fix registry resource what-if for non-existing key

### LINKED ISSUE #1691: `--what-if` fails on a Registry resource with `_exist: false`
On dsc 3.3.0-rc.2, asking for a preview of a configuration that removes a registry value fails with `Error: JSON: EOF while parsing a value at line 1 column 0`. The same document runs fine under `config test` and `config set`, and the same `--what-if` call works if the resource is creating a value rather than removing one.

Here is the document:

```yaml
resources:
- name: remove the value
  type: Microsoft.Windows/Registry
  properties:
    keyPath: HKCU\Software\DscWhatIfProbe
    valueName: V
    _exist: false
```

`dsc config set --what-if -f` on that gives the error above and nothing else. `dsc config validate -f` says it is valid, and `dsc config test -f` runs it and reports state normally, so the document itself is fine.

Changing only `_exist` to `true` and adding a `valueData` makes the same `--what-if` call work as expected, and it returns the preview you would hope for, including `"whatIf":["Key 'DscWhatIfProbe' not found, would create it"]`. So `--what-if` is clearly wired up for this resource in general, and it is specifically the removal case that falls over. I mention that because issue #1529 tracks adding `--what-if` to built-in resources that do not have it yet, and `Microsoft.Windows/Registry` is not on that list, presumably because it already does. This looks like a different problem to me rather than part of that work.

I first hit this on 3.2.3 and confirmed it is unchanged on 3.3.0-rc.2.

The reason it matters more than a normal preview gap: removing a value is the one operation you cannot undo by running the configuration again, so it is the one people most want to rehearse first. At the moment a document that removes values is exactly the document `--what-if` will not describe, and the error gives no clue that removal is the reason, which sent me looking at my document rather than at the operation.

For context on why we generate removals at all, the Windows Group Policy client deletes values rather than blanking them in several ordinary cases, so any tool converting Group Policy into DSC produces `_exist: false` resources routinely rather than exceptionally.

Environment is Windows 11 Pro 26200, dsc 3.3.0-rc.2 and 3.2.3.

SteveL-MSFT:
# PR Summary

The registry library didn't handle the case where it's `what-if` and the key didn't exist and `_exist: false` which simply returned nothing as nothing needed to be done, however, this empty result causes the `set` operation to fail with an error.

The fix here is if this is `what-if` and the key doesn't exist, then add metadata recognizing that the key already doesn't exist so nothing would be done.

This required one other change to allow additional properties to `DeleteResult` which gets thrown away.  The reason for this is to simplify resource development so they can return the same struct output for `delete` as in other cases instead of special casing within the resource.

Due to a known issue with the rust code cov tools, needed to update how the main process exits to return an exit code as the current use of `process::exit()` somehow causes the coverage data to not be collected.

For this config:

```yaml
resources:
- name: remove the value
  type: Microsoft.Windows/Registry
  properties:
    keyPath: HKCU\Software\NotExist
    valueName: V
    _exist: false
```

now results in this output (instead of an error):

```yaml
executionInformation:
  duration: PT5.4564664S
  endDatetime: 2026-08-25T14:33:44.902565600-07:00
  executionType: whatIf
  operation: set
  securityContext: restricted
  startDatetime: 2026-08-25T14:33:39.446099200-07:00
  version: 3.4.0-preview.1
metadata:
  Microsoft.DSC:
    duration: PT5.4564299S
    endDatetime: 2026-08-25T14:33:44.902529100-07:00
    executionType: whatIf
    operation: set
    securityContext: restricted
    startDatetime: 2026-08-25T14:33:39.446099200-07:00
    version: 3.4.0-preview.1
results:
- executionInformation:
    duration: PT0.066319S
    whatIf:
    - Key 'HKCU\Software\NotExist' not found, would do nothing
  metadata:
    Microsoft.DSC:
      duration: PT0.066319S
    whatIf:
    - Key 'HKCU\Software\NotExist' not found, would do nothing
  name: remove the value
  type: Microsoft.Windows/Registry
  result:
    beforeState:
      keyPath: HKCU\Software\NotExist
      _exist: false
    afterState:
      keyPath: HKCU\Software\NotExist
      _exist: false
    changedProperties: []
messages: []
hadErrors: false
```

## PR Context

Fix #1691

### DIFF
--- a/lib/dsc-lib-registry/locales/en-us.toml
+++ b/lib/dsc-lib-registry/locales/en-us.toml
@@ -18,6 +18,7 @@ unsupportedValueDataType = "Unsupported registry value data type"
 whatIfCreateKey = "Key '%{subkey}' not found, would create it"
 whatIfDeleteValue = "Would delete value '%{value_name}'"
 whatIfDeleteSubkey = "Would delete subkey '%{subkey_name}'"
+whatIfDeleteNonexistingKey = "Key '%{subkey}' not found, would do nothing"
 
 [offreg]
 loadFailed = "Failed to load offreg.dll"
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -487,6 +487,7 @@ jobs:
           steps.coverage.outputs.has_rust_changes == 'true'
           && steps.coverage.outputs.coverage_failed != 'true'
           && steps.coverage.outputs.percentage < 70
+          && !contains(github.event.pull_request.labels.*.name, 'Ok-CodeCoverage')
         run: |
           Write-Error "Code coverage is ${{ steps.coverage.outputs.percentage }}%, which is below the 70% minimum threshold."
           exit 1
--- a/lib/dsc-lib-registry/src/lib.rs
+++ b/lib/dsc-lib-registry/src/lib.rs
@@ -330,6 +330,7 @@ impl RegistryHelper {
                     what_if_metadata.push(t!("registry_helper.whatIfDeleteNonexistingKey", subkey = &self.config.key_path).to_string());
                     return Ok(Some(Registry {
                         key_path: self.config.key_path.clone(),
+                        value_name: self.config.value_name.clone(),
                         metadata: Some(Metadata { what_if: Some(what_if_metadata) }),
                         ..Default::default()
                     }));
--- a/lib/dsc-lib-registry/src/lib.rs
+++ b/lib/dsc-lib-registry/src/lib.rs
@@ -316,6 +316,9 @@ impl RegistryHelper {
             return self.remove_offline();
         }
 
+        // Accumulate what-if metadata like set()
+        let mut what_if_metadata: Vec<String> = Vec::new();
+
         // For deleting a value, we need SetValue permission (KEY_SET_VALUE).
         // Try to open with the minimal required permission.
         // If that fails due to permission, try with AllAccess as a fallback.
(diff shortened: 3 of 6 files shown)


SteveL-MSFT:
Code coverage collection failure is due to https://github.com/rust-lang/rust/issues/77553

lib/dsc-lib-registry/locales/en-us.toml
```diff
@@ -18,6 +18,7 @@ unsupportedValueDataType = "Unsupported registry value data type"
 whatIfCreateKey = "Key '%{subkey}' not found, would create it"
 whatIfDeleteValue = "Would delete value '%{value_name}'"
 whatIfDeleteSubkey = "Would delete subkey '%{subkey_name}'"
+whatIfDeleteNonexistingKey = "Key '%{subkey}' not found, would do nothing"
```
tgauth, line 21: Minor suggestion on wording, even though it deviates from the format of the existing messages:
```suggestion
whatIfDeleteNonexistingKey = "Key '%{subkey}' does not exist, no action would be taken"
```

michaeltlombardi (approved):

lib/dsc-lib-registry/locales/en-us.toml
SteveL-MSFT: I like your suggestion, feel free to create a new PR if you want to change it (maybe also review the other messages)

PR #1699, opened 2026-08-31
(GH-538) Set URI and docs keywords for `DscRepoSchema` types

michaeltlombardi:
# PR Summary

This change:

1. Defines new extension methods for operating on `schemars::Schema` instances:

   - `[get|set]_title` - Retrieve and override the `title` keyword as a string slice.
   - `[get|set]_description` - Retrieve and override the `description` keyword as a string slice.
   - `set_markdown_description` - Override the `markdownDescription` keyword as a string slice.
   - `[get|set]_meta_schema` - Retrieve and override the `$schema` keyword as a string slice.
   - `get_meta_schema_as_url` - Retrieve the `$schema` keyword as a `Url`.
   - `has_meta_schema_keyword` - Indicate whether schema defines the `$schema` keyword.

1. Defines new transformer functions for the `DscRepoSchema` trait:

   - `transform_export_schema_uris` - Insert the `$id` and `$schema` keywords with the `default_export_*` function outputs.
   - `transform_schema_docs` - Insert the `title`, `description`, and `markdownDescription` keywords if they have defined localization strings.
   - `transform_schema_docs_strict` - Insert the `title`, `description`, and `markdownDescription` keywords if they have defined localization strings. If any translation strings are missing, panic.

1. Updates the type definitions in `dsc-lib` for every type that derives or implements `DscRepoSchema` to ensure that the `$id` and `$schema` keywords are always populated and the docs keywords are defined _if_ the translation strings are provided. Where the type was already using translated docs this change uses the strict transform to catch regressions in translation string definitions.

## PR Context

Prior to this change the generated schemas were not consistently defining the `$id` and `$schema` keywords, which we need to properly populate for every schema that we publish in its own file. Additionally, the documentation keywords were manually inserted for every schema (that has translated docs already defined).

Effectively, before this PR, every struct needed a definition like the following when deriving `JsonSchema` and `DscRepoSchema`:

```rust
#[derive(Debug, Clone, JsonSchema, DscRepoSchema)]
#[dsc_repo_schema(base_name = "struct", folder_path = "example")]
#[schemars(
    title = schema_i18n!("title"),
    description = schema_i18n("description"),
    extend(
        "$schema" = ExampleStruct::default_export_meta_schema_uri(),
        "$id"     = ExampleStruct::default_export_schema_id_uri(),
        "markdownDescription" = schema_i18n!("markdownDescription"),
    )
)]
pub struct ExampleStruct {
    // Elided for brevity
}
```

With these changes, the definition now looks like:

```rust
#[derive(Debug, Clone, JsonSchema, DscRepoSchema)]
#[dsc_repo_schema(base_name = "struct", folder_path = "example")]
#[schemars(
    transform = ExampleStruct::transform_export_schema_uris,
    transform = ExampleStruct::transform_schema_docs,
)]
pub struct ExampleStruct {
  // Elided for brevity
}
```

This corrects various problems in the schema generation and export pipeline, which relies on the `$id` keyword to correctly manage bundled schema resources, and ensures that we can update the docs keywords for types without needing to modify the type definition directly.


michaeltlombardi:
Investigating test failures led me down a rabbit hole. In short, reference lookups begin failing for bundled schema resources when they define the `$id` keyword. Contributing factors:

1. Schemars emits references to other defined types as `"$ref": "#/$defs/<TypeName>".
2. Schemars inserts definitions for other defined types in `$defs/<TypeName>`.
3. When resolving a reference that begins with `#/` the resolution is against _the current schema resource_. For non-bundled schemas, this works fine with schemars/jsonschema because the only schema resource is the root document. For bundled schemas (which schemars generates for any type with fields/variants that also implement `JsonSchema`), this breaks resolution because references in a bundled schema resource resolve against _that_ resource.

   For example, the references in this snippet all resolve:
   
   ```yaml
   $id: https://contoso.com/schemas/example # root schema URI
   type: object
   properties:
     foo: { $ref: '#/$defs/foo' } # effectively 'https://contoso.com/schemas/example#/$defs/foo'
     bar: { $ref: 'https://contoso.com/schemas/fields/bar' }
   $defs:
     foo:
       type: object
       properties:
         baz: { $ref: '#/$defs/baz' } # effectively 'https://contoso.com/schemas/examples#/$defs/baz'
     bar:
       $id: https://contoso.com/schemas/fields/bar
       type: array
       items: { $ref: '/schemas/fields/baz' } # effectively 'https://contoso.com/schemas/fields/baz'
     baz:
       $id: https://contoso.com/schemas/fields/baz
       type: string
       pattern: '^\w+$'
   ```
   
   However, when we define `$id` for the `foo` definition, the schema is no longer valid:
   
   ```yaml
   $id: https://contoso.com/schemas/example # root schema URI
   type: object
   properties:
     foo: { $ref: '#/$defs/foo' } # effectively 'https://contoso.com/schemas/example#/$defs/foo'
     bar: { $ref: 'https://contoso.com/schemas/fields/bar' }
   $defs:
     foo:
       $id: https://contoso.com/schemas/fields/foo
       type: object
       properties:
         baz: { $ref: '#/$defs/baz' } # effectively 'https://contoso.com/schemas/fields/foo#/$defs/baz
     bar:
       $id: https://contoso.com/schemas/fields/bar
       type: array
       items: { $ref: '/schemas/fields/baz' } # effectively 'https://contoso.com/schemas/fields/baz'
     baz:
       $id: https://contoso.com/schemas/fields/baz
       type: string
       pattern: '^\w+$'
   ```
   
   At this point, validator compilation fails with a message like `Error when resolving schema reference '#/$defs/baz'. Path '$defs.foo.properties.baz'`

To work around these limitations, the latest commit:

1. Sets all types that derive `JsonSchema` but not `DscRepoSchema` to use the `inline` schemars attribute (instead of inserting a reference/definition, schemars pushes the generated schema into wherever it's being used).
2. Defines a new helper macro, `dsc_repo_schema_for!()` to retrieve the generated schema and canonicalize all references and definitions. This fixes references to any schema that defines the `$id` keyword.
3. Updates the `schema_for` integration tests to use the new macro, validate that the schema is valid for its meta schema, and validate that the schema compiles.

SteveL-MSFT (approved):

PR #1700, opened 2026-08-31
Update version and dependencies for 3.3.0 (GA)

michaeltlombardi:
# PR Summary

This change updates the version for the `dsc` CLI to `3.3.0` and updates the dependencies for external crates.

## PR Context

Updating prior to release.

### DIFF
--- a/dsc/Cargo.toml
+++ b/dsc/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "dsc"
-version = "3.3.0-rc.2"
+version = "3.3.0"
 edition = "2024"
 
 [dependencies]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -164,7 +164,7 @@ checksum = "82f6aeea286b8eb4dd3431a1be1b59d290ace00f5bfd8e2a159bc2a05e2c1667"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 3.0.3",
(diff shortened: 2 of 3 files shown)


michaeltlombardi:
@SteveL-MSFT - didn't update the `jsonschema` dependency. It looks like there's a lot of breaking changes that require code updates. Will address tomorrow if we want to use the latest version for 3.3.0

SteveL-MSFT:
@michaeltlombardi not 3.3-rc.3, should just be 3.3.0 for GA

dsc/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "dsc"
-version = "3.3.0-rc.2"
+version = "3.3.0-rc.3"
```
SteveL-MSFT:
```suggestion
version = "3.3.0"
```

You should do this change locally so cargo.lock gets the update as well

SteveL-MSFT (changes requested):

SteveL-MSFT:
> Steve Lee (@SteveL-MSFT) - didn't update the `jsonschema` dependency. It looks like there's a lot of breaking changes that require code updates. Will address tomorrow if we want to use the latest version for 3.3.0

Yes, don't update `jsonschema`, I believe I already made those changes for `main` branch, but not needed for 3.3 release

SteveL-MSFT (approved):

